PR #5, opened 2025-09-19
Makefile: cleanup

mook-as:
This is just an optional clean as I start to dig into the repo; feel free to defer.

- Simplify expression to find Kubernetes module versions
- Simplify expression to find API groups
- Add `ltag` and `openshift-goimports` as tools in `go.mod` so they can be pinned and updated via depdendabot
- Add clean target
- Use a different RDD instance for each BATS test group so `make -j` works


Makefile
```diff
@@ -91,14 +91,12 @@ lint:
...
-	go install github.com/containerd/ltag@latest
 	# exclude bats/lib, but --excludes only takes a dir name, not a path name
-	$$(go env GOPATH)/bin/ltag -v -t .ltag -path . --excludes=lib
+	go tool ltag -v -t .ltag -path . --excludes=lib
 .PHONY: ltag
 
 imports:
-	go install github.com/openshift-eng/openshift-goimports@latest
-	$$(go env GOPATH)/bin/openshift-goimports
+	go run openshift-goimports
```
jandubois, line 99: Should move the `.PHONY: imports` line here from the bottom; not sure how they got separated.

jandubois (approved):
Thanks, LGTM

PR #6, opened 2025-09-20
Create "lima" api group with "LimaVM" controller

jandubois:
This is just copied or generated code; the reconciler is not doing anything and the LimaVM type has no fields defined yet.


pkg/apis/lima/v1alpha1/limavm_types.go
```diff
@@ -0,0 +1,80 @@
...
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
mook-as, line 12: Do we need to keep these instructional comments?

mook-as (approved):
This is mostly generated code anyway

pkg/apis/lima/v1alpha1/limavm_types.go
jandubois: We don't; only the kubebuilder tags are required.

I would leave the comments for now and remove them once we have custom content in the reconciler. (I think I've left it in the other controller files too, so will have to remove there too).

PR #7, opened 2025-09-22
Fix --secure-port setting

jandubois:
The Kubernetes control plane already defines a `--secure-port` flag, so `rdd svc serve` should not define one as well; they are just competing with setting default value, depending on load order, so created a bug that looked like a race condition.

Also the actual (fallback) port needs to be determined when the control plane is started, not when the service is created.

Finally we should always use `127.0.0.1` as the bind address to make sure it is covered by the cert. Otherwise, if the local machine address changes, the new address will cause cert validation issues.

---

Fun fact: The hashed instance index for `RDD_INSTANCE=0` is the same as for `RDD_INSTANCE=test` (`148`, in case you are wondering).


mook-as:
I think it would be nice (in the future, in a separate PR) to make the RDD-internal kubernetes connection use a unix socket (named pipe on Windows) instead of a TCP port; that would resolve issues like this, as well as port collisions, etc. But that's going to be medium term at best (and I assume I'll need to be the one to hack on that, since it'd affect the frontend).  That would only make sense if we only expect to talk to it via `rdd` (e.g. `rdd ctl`) and never `kubectl --config=…` though.

All that, of course, is completely irrelevant for this PR.

jandubois:
> That would only make sense if we only expect to talk to it via `rdd` (e.g. `rdd ctl`) and never `kubectl --config=…` though.

So how would `rdd ctl ...` work then? It is just setting `KUBECONFIG` and then invokes the embedded `kubectl` command. It seems like upstream support for adding Unix domain socket support is going to be rejected: https://github.com/kubernetes/kubernetes/issues/131053

Port collisions are not really an issue, as the server will fall back to a random port. The current mechanism is probably a bit over-engineered. Having a predictable port is just for the convenience of the developer reading logs and recognizing the port number; the software does not care.

jandubois:
I marked this "ready for review", but we should probably merge #8 and #9 first, and then rebase this one on top.

Changes I just pushed:
* also define `--secure-port` on `rdd svc start` (and add a BATS test for it)
* delete `pkg/service/cmd/service_test.go` which is outdated and superseded by `bats/tests/20-service/5-port-fallback.bats`

jandubois:
Superseded by #12

PR #8, opened 2025-09-22
golangci-lint fixes

mook-as:
The first commit is the automated fixes; the second commit is the manual fixes.

This is needed before we can try doing lints in CI.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,8 +109,6 @@ linters:
       - name: unexported-return
       - name: unreachable-code
       - name: unused-parameter
-        arguments:
-          - allowRegex: ^_
       - name: use-any
       - name: var-declaration
       - name: var-naming
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -109,6 +109,8 @@ linters:
       - name: unexported-return
       - name: unreachable-code
       - name: unused-parameter
+        arguments:
+          - allowRegex: ^_
       - name: use-any
       - name: var-declaration
       - name: var-naming
--- a/cmd/app-controller/main.go
+++ b/cmd/app-controller/main.go
@@ -7,15 +7,11 @@ package main
 import (
 	"os"
 
-	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 	// Import app controller packages to trigger init() functions.
 	_ "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/app/demo"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/external"
 )
 
 func main() {
-	err := external.RunControllers("app", base.GetAllControllers)
-	if err != nil {
-		os.Exit(1)
-	}
+	os.Exit(external.RunControllers("app"))
 }
--- a/cmd/app-controller/main.go
+++ b/cmd/app-controller/main.go
@@ -5,6 +5,8 @@
 package main
 
 import (
+	"os"
+
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 	// Import app controller packages to trigger init() functions.
 	_ "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/app/demo"
@@ -12,7 +14,8 @@ import (
 )
 
 func main() {
-	external.RunControllers("app", func() []base.Controller {
-		return base.GetAllControllers()
-	})
+	err := external.RunControllers("app", base.GetAllControllers)
+	if err != nil {
+		os.Exit(1)
+	}
 }
--- a/cmd/rdd/main.go
+++ b/cmd/rdd/main.go
@@ -71,7 +71,7 @@ func newVersionCommand() *cobra.Command {
 		Use:   "version",
 		Short: "Print RDD version information",
 		Long:  help.Doc(`Print RDD version information and exit.`),
-		Run: func(cmd *cobra.Command, _args []string) {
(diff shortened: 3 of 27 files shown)


.golangci.yml
```diff
@@ -109,6 +109,8 @@ linters:
       - name: unexported-return
       - name: unreachable-code
       - name: unused-parameter
+        arguments:
+          - allowRegex: ^_
```
jandubois: I would prefer to just use `_` for unused parameters. If you use e.g. `_args`, then you won't get an error if the variable is unexpectedly being used in the function.

cmd/app-controller/main.go
```diff
@@ -5,14 +5,17 @@
...
 )
 
 func main() {
-	external.RunControllers("app", func() []base.Controller {
-		return base.GetAllControllers()
-	})
+	err := external.RunControllers("app", base.GetAllControllers)
+	if err != nil {
+		os.Exit(1)
+	}
```
jandubois: Hah, good catch on the useless wrapper.

I'm tempted to leave the `os.Exit(1)` calls where they are and rename the function to `main`, which is really what it is (with non-standard options). The linter will also not complain when the function is called `main` iirc:

```suggestion
	external.main("app", base.GetAllControllers)
```

cmd/rdd/main.go
```diff
@@ -71,7 +71,7 @@ func newVersionCommand() *cobra.Command {
 		Use:   "version",
 		Short: "Print RDD version information",
 		Long:  help.Doc(`Print RDD version information and exit.`),
-		Run: func(cmd *cobra.Command, args []string) {
+		Run: func(cmd *cobra.Command, _args []string) {
```
jandubois: As I mentioned above, just use `_`. And if all parameters are unused, you can drop the `_` as well and leave them unnamed.

```suggestion
		Run: func(cmd *cobra.Command, _ []string) {
```

pkg/controllers/rdd/notary/controller.go
```diff
@@ -289,7 +289,7 @@ func (v *NotaryValidator) validateNotary(ctx context.Context, notary *v1alpha1.N
 			// For dry run requests, we still perform validation but can skip side effects
 			// In this case, we don't have side effects, so we proceed with normal validation
 			// but log that this is a dry run
-			fmt.Printf("[DryRun] Webhook validating Notary %s/%s\n", req.Namespace, req.Name)
+			fmt.Fprintf(os.Stdout, "[DryRun] Webhook validating Notary %s/%s\n", req.Namespace, req.Name)
```
jandubois: This looks like left-over debugging code, and I don't see where stdout would actually go; this should be a logging call. But maybe leave it for now.

pkg/external/main.go
```diff
@@ -24,7 +23,7 @@ import (
 
 // RunControllers is the main function for external API group controllers.
 // It handles command line parsing, kubeconfig retrieval, discovery checking, and shared manager setup.
-func RunControllers(apiGroupName string, controllersFunc func() []base.Controller) {
+func RunControllers(apiGroupName string, controllersFunc func() []base.Controller) error {
```
jandubois:
```suggestion
func main(apiGroupName string, controllersFunc func() []base.Controller) {
```

pkg/service/cmd/service_test.go
jandubois, line 1: I need to double-check, but this is ancient, and is probably redundant with the latest test changes. For now we can just accept the changes to make the linter happy.

jandubois (commented):
Looks fine to me, but I would like to use `_` for unused parameters, and do the renaming to `main` thing to keep the various external controllers as minimal as possible.

pkg/controllers/rdd/notary/controller.go
jandubois: I still think this can be removed completely, but if you make other changes, maybe change it to

```go
klog.V(1).Infof("[DryRun] Webhook validating Notary %s/%s", req.Namespace, req.Name)
```

cmd/app-controller/main.go
mook-as: That would mean `main` isn't exported, so we can't call it?

jandubois: I just realized that this should be just

```go
external.main("app")
```

Because the list of controllers is automatically generated now. This is just leftover stuff from incomplete refactoring earlier.

cmd/rdd/main.go
mook-as: Will do (and dropped the config change, so everything must be named `_`). I didn't expect having multiple arguments named `_` to work, but it does, so yay?

pkg/service/cmd/service_test.go
mook-as: I'd expect something like `pkg/service/controllers/ports_test.go#TestGetAvailablePort`, in that case. But I don't feel like I have a good handle on the code yet, so I prefer leaving this in for now and dropping it in a follow-up PR if we confirm it's being tested correctly.

cmd/app-controller/main.go
mook-as: I realized that we can have it return the exit code and have the caller call `os.Exit()` inline instead, which looks nicer (now that we no longer pass in `base.GetAllControllers).

jandubois (approved):
Thanks, LGTM

cmd/rdd/main.go
jandubois: > I didn't expect having multiple arguments named `_` to work, but it does, so yay?

Yeah. And if all arguments are unused, you can leave them off all together:

```
func (string) { ... }
```

PR #9, opened 2025-09-22
CI: Add golangci-lint

mook-as:
This adds golangci-lint to a GitHub action.

- We run it for darwin, linux, and windows, but do so using `GOOS` because Linux runners are faster.
- I elected to go for the action instead of running `make lint` for caching &etc.; however, `make lint` should be kept working (and ideally run as a pre-commit hook on developer machines).
- This should be tested after #8 merges; hence draft for now.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,24 @@
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  golangci-lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [darwin, linux, windows]
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  golangci-lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [darwin, linux, windows]
+    steps:
+    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5
```
jandubois: I think the versions must have a leading space and include the full semver. Otherwise dependabot will not update it when it updates the action commit hash.

```suggestion
    - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```

jandubois (approved):
Thanks, LGTM

PR #11, opened 2025-09-23
Fix lint issues

mook-as:
Turns out the SO_LINGER stuff wasn't actually needed, after all.

Fixes CI issues.

### DIFF
--- a/cmd/lima-controller/main.go
+++ b/cmd/lima-controller/main.go
@@ -5,14 +5,11 @@
 package main
 
 import (
-	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 	// Import rdd controller packages to trigger init() functions.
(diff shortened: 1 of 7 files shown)


cmd/lima-controller/main.go
```diff
@@ -5,14 +5,11 @@
...
 	// Import rdd controller packages to trigger init() functions.
 	_ "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/lima/limavm"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/external"
 )
 
 func main() {
-	external.RunControllers("lima", func() []base.Controller {
-		return base.GetAllControllers()
-	})
+	external.RunControllers("lima")
```
jandubois, line 14: Should this call `os.Exit()` now?

```suggestion
	os.Exit(external.RunControllers("lima"))
```

jandubois (approved):

PR #12, opened 2025-09-23
Fix --secure-port setting

jandubois:
The Kubernetes control plane already defines a --secure-port flag, so `rdd svc serve` should not define one as well; they are just competing with setting default value, depending on load order, so created a bug that looked like a race condition.

Also the actual (fallback) port needs to be determined when the control plane is started, not when the service is created.

Finally we should always use 127.0.0.1 as the bind address to make sure it is covered by the cert. Otherwise, if the local machine address changes, the new address will cause cert validation issues.


bats/tests/20-service/5-port-fallback.bats
```diff
@@ -0,0 +1,177 @@
...
+# Track netcat PIDs for cleanup
+declare -a NETCAT_PIDS=()
+
+local_teardown() {
+    kill "${NETCAT_PIDS[@]}" 2>/dev/null || :
+}
+
+# Extract port from kubeconfig server URL
+get_kubeconfig_port() {
+    rdd svc config | grep "server:" | sed 's/.*127\.0\.0\.1:\([0-9]*\).*/\1/'
```
mook-as: `rdd ctl config view --output='jsonpath={.clusters[].cluster.server}'` would be nice, but I assume this is done this way because `rdd ctl` would not be available at some points?

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -0,0 +1,177 @@
...
+}
+
+# Check if a port is available on localhost
+is_port_available() {
+    local port="$1"
+    ! netstat -an | grep -q "^tcp4.*127\.0\.0\.1\.$port "
```
mook-as: On modern linux, we may need to use `ss` instead of `netstat`.

Also, using `-f inet` instead of grepping for `tcp4` may be more portable.

cmd/rdd/service.go
```diff
@@ -91,6 +94,14 @@ func serviceCreateAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	args = append(args, "--controllers", controllers)
+	if cmd.Flags().Changed("secure-port") {
+		securePort, err := cmd.Flags().GetInt("secure-port")
+		if err != nil {
+			return err
+		}
+		args = append(args, "--secure-port", strconv.Itoa(securePort))
```
mook-as, line 102: Alternatively, `fmt.Sprintf` works too (since we already import `fmt`). But it's probably built with `strconv` underneath.

pkg/service/cmd/options/options.go
```diff
@@ -166,6 +167,21 @@ func (o *Options) Complete() (*CompletedOptions, error) {
 	o.ControlPlane.Admission.GenericAdmission.DisablePlugins = sets.List[string](rddadmission.DefaultOffAdmissionPlugins())
 	o.ControlPlane.Admission.GenericAdmission.RecommendedPluginOrder = rddadmission.AllOrderedPlugins
 
+	// Force external address to localhost to ensure certificate includes 127.0.0.1 in SAN list
+	o.ControlPlane.SecureServing.ExternalAddress = net.ParseIP("127.0.0.1")
```
mook-as: Any reason not to use [`net.IPv4`](https://pkg.go.dev/net@go1.25.1#IPv4) directly?

```suggestion
	o.ControlPlane.SecureServing.ExternalAddress = net.IPv4(127, 0, 0, 1)
```
Same with the next line.

Makefile
```diff
@@ -110,6 +110,11 @@ check-bats:
 	fi
 .PHONY: check-bats
 
+# Run BATS tests for BATS helpers
+bats-helpers: check-bats
```
mook-as, line 114: Should this go into `bats-all` too?

mook-as (changes requested):

Makefile
jandubois: Sure, why not; they are very fast.

I first thought you would just run them when you change the helpers, but since they only take a few seconds, we can just as well run them always.

bats/tests/20-service/5-port-fallback.bats
jandubois: No, `rdd ctl` should always be available. Just like `rdd svc config`, it will start the control plane by itself, if it isn't running already. We would still have to extract the port name though:

```shell
rdd ctl config view --output='jsonpath={.clusters[].cluster.server}' | sed 's/.*127\.0\.0\.1:\([0-9]*\).*/\1/'
```

I guess we could simplify the `sed` expression, but this way it is robust against the URL including a path.

bats/tests/20-service/5-port-fallback.bats
jandubois: I don't know; do they actually remove `netstat` now? The problem is that `ss` is not available for macOS; it is Linux-only afaik.

How about:

```shell
! lsof -i -nP 2>/dev/null | grep -q -E " 127\.0\.0\.1:$port \(LISTEN\)$"
```

cmd/rdd/service.go
jandubois: I don't see the point in using `fmt.Sprintf("%d", securePort)`. As you wrote, it probably will call `strconv` internally anyways. I guess we could use `"--secure-port=%d"` as the format, but 🤷. I think this is fine.

pkg/service/cmd/options/options.go
jandubois: No. We just use `net.ParseIP(constant)` all over the place in Lima, so I'm not used to the other form (which I now see we use in a very few places there too). I guess Lima also needs an update.

jandubois (commented):
Thanks! Will make some changes...

bats/tests/20-service/5-port-fallback.bats
mook-as: Yeah I'm not too concerned about the `sed`; it's mostly the `grep` that I was worried about (and your changes fix it).

bats/tests/20-service/5-port-fallback.bats
mook-as: Yes, the Leap machine I'm staring at (actually a WSL instance, but close enough) has `ss` but no `netstat`.

```
> lsof -i -nP 2>/dev/null
> ss -an -f inet | grep tcp
tcp   LISTEN 0      1000   10.255.255.254:53        0.0.0.0:*
tcp   LISTEN 0      128           0.0.0.0:22        0.0.0.0:*
```
`lsof` actually doesn't produce any output here…

Note that (for this subset at least) `netstat -an -f inet` and `ss -an -f inet` have the same command lines, so it would be something like `command -v ss && echo ss || echo netstat` at least.

mook-as (dismissed):
Everything looks fine, except for the `netstat`/`ss` thing.

Approving anyway so we can argue about it in a different PR, since I assume you're only testing on macOS for now.

jandubois:
> Approving anyway so we can argue about it in a different PR, since I assume you're only testing on macOS for now.

Yeah, but it is supposed to run in CI, so let's get it right while we are at it.

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -0,0 +1,184 @@
...
+    # Calculate expected port dynamically and occupy it
+    expected_port=$(get_expected_port)
+    nc -l 127.0.0.1 "$expected_port" &
+    NETCAT_PIDS+=($!)
+
+    # Give netcat time to bind
+    sleep 1
+
+    # Verify the port is actually occupied by checking netstat
+    run -0 netstat -an
```
mook-as: This will be an issue.

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -0,0 +1,184 @@
...
+    nc -l 127.0.0.1 6443 &
+    NETCAT_PIDS+=($!)
+    nc -l 127.0.0.1 "$expected_port" &
+    NETCAT_PIDS+=($!)
+
+    # Give netcat time to bind
+    sleep 2
+
+    # Verify both ports are occupied by checking netstat
+    run -0 netstat -an
```
mook-as: This too.

I think line 103 is the last occurrence.

mook-as (changes requested):

bats/tests/20-service/5-port-fallback.bats
jandubois: Oops, they should all be replaced by `is_port_available`. Done now.

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -0,0 +1,184 @@
...
+    run -0 rdd ctl config view --output='jsonpath={.clusters[0].cluster.server}'
+    sed 's/.*127\.0\.0\.1:\([0-9]*\).*/\1/' <<<"$output"
+}
+
+# Check if a port is available on localhost
+is_port_available() {
+    local port="$1"
+    # some Linux distros no longer include netstat, but ss is not available on macOS
+    if command -v ss >/dev/null; then
+        run -0 ss -an -f inet
```
mook-as, line 23: I forget, this is run from a test, does this need the `|| return` now? Or did we get the patched BATS with `errexit`?

mook-as (dismissed):

bats/tests/20-service/5-port-fallback.bats
jandubois:
```
❯ (cd bats/lib/bats-core; git log -3 --oneline)
397c4a5 (HEAD -> errexit, origin/errexit) Make sure tmpdir exists in case run() is called outside a @test
507da84 Add --errexit flag to enable errexit in run commands
855844b (origin/master, origin/HEAD) Merge pull request #1103 from lucasbernardestoradex/master
```

(same in `rancher-sandbox/rancher-desktop` and `lima-vm/lima`).

But you are right, we are missing the part in `load.bash` that force-enables it.

I'll add it in another commit on this PR.

jandubois: Still passes all tests after enabling `--errexit`.

mook-as (approved):

PR #13, opened 2025-09-23
Drop openshift-goimports, since golangci-lint supports gci

mook-as:
The formatters are checked with `golangci-lint run`, so no need to add to CI; run `make format` locally to fix things.


jandubois (approved):
Thanks, LGTM

PR #16, opened 2025-09-23
CI: Skip on unprotected upstream branches

mook-as:
This disables the CI run on push to upstream repository branch that is _not_ protected, i.e. topic branches that are upstream.  This is meant to prevent running duplicate checks (because these branches are expected to be made into pull requests instead).


jandubois (approved):
Thanks, LGTM

PR #18, opened 2025-09-24
CI: Add dependabot

### LINKED ISSUE #14: Setup dependabot to update Go and GHA dependencies

mook-as:
The Kubernetes modules are kind of weird due to how they manage `/staging` (see info [here](https://github.com/kubernetes/kubernetes/blob/c0b9ab3351622bec1d921690cbe9e2f4e86cddcd/staging/README.md)), so there's something extra to do it separately and then make fix-up commits like we did with the Rancher Desktop dependabot action.

The script can be tested locally with `go run scripts/go-module-k8s-bump/main.go` (try modifying `go.mod` to use `k8s.io/kubernetes v1.34.1` before running it, for example; no need to use `go get` or anything like that).

We may need to ask for EIO help to set up vault secrets; we may need to confirm that after the first run fails.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,49 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: daily
+  cooldown:
+    default-days: 7
+- package-ecosystem: gomod # spellchecker:ignore
+  directory: /
+  schedule:
+    interval: daily
+  cooldown:
+    default-days: 7
+  ignore:
+  - { dependency-name: k8s.io/api } # spellchecker:ignore
+  - { dependency-name: k8s.io/apiextensions-apiserver } # spellchecker:ignore
+  - { dependency-name: k8s.io/apimachinery } # spellchecker:ignore
+  - { dependency-name: k8s.io/apiserver } # spellchecker:ignore
+  - { dependency-name: k8s.io/cli-runtime } # spellchecker:ignore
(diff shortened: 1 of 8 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,49 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: /
+  schedule:
+    interval: daily
+  cooldown:
+    default-days: 7
+- package-ecosystem: gomod # spellchecker:ignore
```
jandubois, line 9: You have added `gomod` now, so there is no reason to ignore this line anymore.

But given all the ignore lines below, doesn't it make more sense to add this file to `.github/actions/spelling/excludes.txt`? We don't really care about spelling issues in this config file.

scripts/go-module-k8s-bump/main.go
```diff
@@ -0,0 +1,201 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Check the dependabot configuration to make sure it is ignoring all of the
+// Kubernetes modules correctly; print out a list of missing modules and return
+// an error if any are missing.  We do not modify the file in-place because
+// doing so would end up dropping any comments in the file.
```
jandubois, line 152: You can modify YAML files with `yq`, which will preserve comments.

jandubois (approved):
Thanks, LGTM

.github/dependabot.yml
mook-as: I'll do this (and the other thing) in a follow-up.

PR #19, opened 2025-09-24
Add BATS linting

### LINKED ISSUE #17: Setup shellcheck and shfmt for the bats tests and helpers

mook-as:
Fixes #17

This moves the BATS-related make targets to the `BATS` directory, so that the top level makefile is easier to parse.  All the targets can still be run at the top level (it just runs `make -C`).

For the fixes commit, `git show --ignore-space-at-eol` is recommended.


jandubois (approved):
Thanks, LGTM

Issue #10, opened 2025-09-22
Add GitHub action checks to this repo


jandubois:
Should do:
- [x] linting: #9 #19
- [x] openshift-imports (superceeded by gci via #9)
- [x] ltags: #26 
- [ ] run bats: #20

Maybe:
- [x] spellcheck: #25

mook-as:
I don't think there is anything left; we can make more changes in further PRs.

PR #20, opened 2025-09-25
CI: Add BATS

mook-as:
This adds BATS to CI, so that we run it on every push (but not when there's a PR in the same repo).

Depends on #19, so this is draft for now.
Part of #10

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -1,0 +1,139 @@
+name: BATS
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  get-pr-number:
+    name: Lookup Pull Request
+    outputs:
+      number: ${{ steps.lookup.outputs.number }}
+    runs-on: ubuntu-latest
+    steps:
+    - id: lookup
+      if: >-
+        github.event_name == 'push' &&
+        github.repository_owner == 'rancher-sandbox' &&
+        github.ref_type == 'branch' &&
+        !github.ref_protected
+      shell: bash
+      run: >-
+        printf "number=%s\n"
+        $(gh pr list
+        --repo ${{ github.repository }}
+        --search ${{ github.sha }}
+        --json number
+        --jq '.[0].number // 0')
+        | tee --append "$GITHUB_OUTPUT"
+      env:
+        GH_TOKEN: ${{ github.token }}
+  bats:
+    name: BATS
+    needs: get-pr-number
+    if: ${{ ! needs.get-pr-number.outputs.number }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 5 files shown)


jandubois:
> we run it on every push (but not when there's a PR in the same repo).

Can you explain the reasons behind this? And add corresponding comments to the job?

There should be no pushes unless they are part of a PR or merge, so when is this supposed to run?

mook-as:
Currently this is a private repo, so pushing to forks would run into quota issues, hence I'm just pushing to the main repo (in a branch named after my user name), where I _do_ want CI.

This would be unnecessary once this repo is public.

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -19,16 +19,23 @@ get_kubeconfig_port() {
...
     else
         run -0 netstat -an -f inet
+        awk "/^tcp.*127\.0\.0\.1\.${port} / { print \$NF; exit }" <<<"$output"
     fi
-    ! grep -q "^tcp.*127\.0\.0\.1\.$port " <<<"$output"
+}
+
+# Check if a port is available on localhost
+is_port_available() {
+    local port="$1" status
```
jandubois: Let's keep one variable per line, especially if we also include value assignments.
And there is no word splitting on assignment, so you don't need to quote expressions, only literal strings with whitespace.

```suggestion
    local port=$1
    local status
```

bats/tests/20-service/5-port-fallback.bats
```diff
@@ -64,6 +71,14 @@ is_port_available() {
 
     # Calculate expected port dynamically and occupy it
     expected_port=$(get_expected_port)
+
+    while [[ "$(get_port_status "$expected_port")" == "TIME_WAIT" ]]; do
+        sleep 1
+    done
```
jandubois: This should have a timeout and not hang indefinitely. Maybe use `try`?

jandubois (changes requested):
Only some minor issues, but the indefinite loop needs a timeout.

.github/workflows/bats.yaml
```diff
@@ -104,9 +104,9 @@ jobs:
         Get-ItemPropertyValue -Path bin -Name FullName | Out-File -Encoding UTF8 -Append "$env:GITHUB_PATH"
 
         # Set up git configuration
-        git config --global --replace-all core.autocrlf false
-        git config --global --replace-all core.eol lf
-        git config --global --replace-all core.longpath true
+        git config --global --replace-all core.autocrlf false # spellchecker:ignore
+        git config --global --replace-all core.eol lf # spellchecker:ignore
+        git config --global --replace-all core.longpath true #spellchecker:ignore
```
jandubois, line 109: I think we can also exclude workflow files from spell checking, but let's leave that for another day.

jandubois (approved):
Thanks, LGTM

PR #21, opened 2025-09-26
BATS: Fix Windows

mook-as:
This (mostly) fixes BATS for Windows, except for one issue I haven't tracked down yet.

`31-rdd-controllers/notary-events` is failing because the `NoChange` event sometimes gets coalesced with `ValueRecorded` (so we end up with `ValueRecorded` as the reason, but _Spec update received but value unchanged, no ConfigMap update needed_ as the message).  That test is temporarily disabled on Windows; I'll continue to look into it, though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,10 @@ EXE := $(if $(shell sh -c 'command -v winver.exe'),.exe,)
 KUBE_VERSION := $(shell go$(EXE) list -m -f '{{.Version}}' 'k8s.io/kubernetes')
 KUBE_MAJOR_VERSION := $(shell echo $(KUBE_VERSION) | sed 's/v\([0-9]*\).*/\1/')
 KUBE_MINOR_VERSION := $(shell echo $(KUBE_VERSION) | sed "s/v[0-9]*\.\([0-9]*\).*/\1/")
-GIT_COMMIT := $(shell git rev-parse --short HEAD || echo 'local')
-GIT_DIRTY := $(shell git diff --quiet && echo 'clean' || echo 'dirty')
-GIT_VERSION := $(KUBE_VERSION)+rdd-$(shell git describe --tags --match='v*' --abbrev=14 "$(GIT_COMMIT)^{commit}" 2>/dev/null || echo v0.0.0-$(GIT_COMMIT))
-RDD_VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+GIT_COMMIT := $(shell git$(EXE) rev-parse --short HEAD || echo 'local')
+GIT_DIRTY := $(shell git$(EXE) diff --quiet && echo 'clean' || echo 'dirty')
+GIT_VERSION := $(KUBE_VERSION)+rdd-$(shell git$(EXE) describe --tags --match='v*' --abbrev=14 "$(GIT_COMMIT)^{commit}" 2>/dev/null || echo v0.0.0-$(GIT_COMMIT))
+RDD_VERSION := $(shell git$(EXE) describe --match 'v[0-9]*' --dirty='.m' --always --tags)
 RDD_VERSION_TRIMMED := $(RDD_VERSION:v%=%)
 BUILD_DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 PACKAGE := github.com/rancher-sandbox/rancher-desktop-daemon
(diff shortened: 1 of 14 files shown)


bats/helpers/commands.bash
```diff
@@ -22,5 +22,29 @@ curl() {
     command "curl$EXE" "$@"
 }
 rdd() {
-    "$PATH_REPO_ROOT/bin/rdd" "$@"
+    local arg args=("$@")
+    local env envs=()
```
jandubois: It kind of makes sense in this particular situation, but generally we should follow the "one variable per line" rule:

```suggestion
    local arg
    local args=("$@")
    local env
    local envs=()
```

Makefile
```diff
@@ -48,25 +48,25 @@ build: build-rdd build-all-controllers
...
+API_GROUPS := $(notdir $(shell find pkg/controllers -mindepth 1 -maxdepth 1 -type d -not -name base))
 
 # Generate build targets for API group controllers
 define API_GROUP_CONTROLLER_TARGETS
 bin/$(1)-controller$$(EXE): $$(GOLANG_SOURCES)
-	CGO_ENABLED=0 go build -tags="$(TAGS)" -buildvcs=false -gcflags="all=$${GCFLAGS}" -ldflags="$(LDFLAGS)" -o $$@ ./cmd/$(1)-controller
+	CGO_ENABLED=0 go$$(EXE) build -tags="$(TAGS)" -buildvcs=false -gcflags="all=$${GCFLAGS}" -ldflags="$(LDFLAGS)" -o $$@ ./cmd/$(1)-controller
 	ls -lh $$@
-build-$(1)-controller: bin/$(1)-controller$(EXE)
+build-$(1)-controller: bin/$(1)-controller$$(EXE)
```
jandubois, line 67: Don't you need the same change on lines 69 and 72 as well then?

bats/tests/10-cli/3-devmode.bats
```diff
@@ -2,26 +2,26 @@ load '../../helpers/load'
 
 local_setup_file() {
     # Use the local rdd binary from the project (use absolute path to avoid relative path issues)
-    ln -sf "${PATH_REPO_ROOT}/bin/rdd" /tmp/rdd
+    cp "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${BATS_FILE_TMPDIR}/rdd${EXE}"
```
jandubois: Can we put `"${BATS_FILE_TMPDIR}/rdd${EXE}"` into an environment variable and use it in all the tests?

pkg/controllers/base/controller.go
```diff
@@ -113,15 +114,19 @@ func GetKubeConfigFromRDD(ctx context.Context) (*rest.Config, error) {
 // getRDDKubeConfig executes `rdd svc config` and returns the kubeconfig YAML.
 // It first tries to find rdd on PATH, then looks in the same directory as the current executable.
 func getRDDKubeConfig(ctx context.Context) ([]byte, error) {
+	exeName := "rdd"
+	if runtime.GOOS == "windows" {
+		exeName = "rdd.exe"
+	}
```
jandubois, line 120: I assume you want to avoid matching e.g. `rdd.bat` earlier in the `PATH`? Because `exec.LookPath("rdd")` should otherwise work just fine on Windows, trying all the extensions for `PATH_EXT`.

pkg/service/cmd/service.go
```diff
@@ -380,6 +385,8 @@ func StopWithWait(wait bool) error {
 					_ = os.Remove(instance.KubeConfig()) // Ignore error as file might not exist
 					return nil
 				}
+			default:
+				time.Sleep(100 * time.Millisecond)
```
jandubois: What is the point of this default clause? Why not let the `select` block on the channels? The `ticker` already provides the 100ms sleep.

pkg/service/cmd/service_windows.go
```diff
@@ -6,19 +6,31 @@ package service
...
 		return fmt.Errorf("failed to open process %d: %w", pid, err)
 	}
 	defer func() {
 		_ = windows.CloseHandle(hProcess)
 	}()
-	return windows.TerminateProcess(hProcess, 1)
+	if err := windows.TerminateProcess(hProcess, 1); err != nil {
+		return fmt.Errorf("failed to terminate process %d: %w", pid, err)
+	}
+	_, err = windows.WaitForSingleObject(hProcess, uint32(10*time.Second/time.Millisecond))
```
jandubois, line 28: Is 10ms enough? Why not make it 200ms? Is it really important to get the error so quickly, at the risk of sometimes getting an error on a loaded system?

jandubois (changes requested):
I have some comments...

bats/helpers/commands.bash
mook-as: Will do.

bats/tests/10-cli/3-devmode.bats
mook-as: Ah, all tests _in this file_; sure. I was initially interpreting it as across files… Makes sense.

pkg/controllers/base/controller.go
mook-as: Oh, I didn't realize `LookPath` did that; that means we don't need this then. Except that in the `err != nil` case, we still need to use `rdd.exe` on line 129, so that doesn't really help…

pkg/service/cmd/service.go
mook-as: … Actually, I don't know why I added this. Good catch.

pkg/service/cmd/service_windows.go
mook-as: This is ten _seconds_, not milliseconds.  If killing a process (~= `SIGKILL`) takes that long, something is really wrong anyway.  `time.Milliseconds` shows up because this function takes a number of milliseconds as the timeout.

Makefile
mook-as: Ah, right. (It's actually not necessary, since it just delays the expansion of the variable, but it made more sense in my brain.)

But it was _definitely_ needed on line 73 (`$<`), because that needs to be expanded when the final rule is evaluated…

mook-as (commented):
0b4085cc03340b65be677f18f0154b0d25203582 pushed

bats/tests/10-cli/3-devmode.bats
```diff
@@ -2,26 +2,27 @@ load '../../helpers/load'
 
 local_setup_file() {
     # Use the local rdd binary from the project (use absolute path to avoid relative path issues)
-    ln -sf "${PATH_REPO_ROOT}/bin/rdd" /tmp/rdd
+    export RDD_FILE="${BATS_FILE_TMPDIR}/rdd${EXE}"
```
jandubois, line 5: I would have called it `RDD_EXE`, but it doesn't matter...

pkg/service/cmd/service_windows.go
jandubois: Oops, my mistake. 10s is obviously plenty long enough.

jandubois (dismissed):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

PR #22, opened 2025-09-29
Implement unified logging system with global --log-level flag

jandubois:
Eliminates need to run `rdd svc serve` manually with logging options; user doesn't need to know about klog settings.

### DIFF
--- a/cmd/rdd/main.go
+++ b/cmd/rdd/main.go
@@ -66,6 +66,28 @@ func parseInstanceFlag() {
 	}
 }
 
+// setLogLevel sets the log level from command flags.
+func setLogLevel(cmd *cobra.Command, _ []string) error {
+	logLevel, err := cmd.Root().Flags().GetString("log-level")
+	if err != nil {
+		return err
+	}
+	if logLevel == "" {
+		// Default log level: warn for regular mode, debug for developer mode
+		if developer.Mode() {
+			logLevel = "debug"
+		} else {
+			logLevel = "warn"
+		}
+	}
+	level, err := logrus.ParseLevel(logLevel)
+	if err != nil {
+		return err
+	}
+	logrus.SetLevel(level)
+	return nil
+}
+
 func newVersionCommand() *cobra.Command {
 	return &cobra.Command{
 		Use:   "version",
@@ -86,12 +108,22 @@ func main() {
 		Long: help.Doc(`
 			RDD manages the Rancher Desktop 2 background services.
 		`),
-		SilenceUsage:  true,
-		SilenceErrors: true,
+		SilenceUsage:      true,
+		SilenceErrors:     true,
+		PersistentPreRunE: setLogLevel,
 	}
 
 	// Add version flag with verflag-compatible behavior
 	version.AddVersionFlag(cmd.Flags())
+
+	// Add global log-level flag
+	var levels []string
+	for _, level := range logrus.AllLevels {
+		if level != logrus.PanicLevel {
+			levels = append(levels, level.String())
+		}
+	}
+	cmd.PersistentFlags().String("log-level", "", "Log level: "+strings.Join(levels, ", "))
 
 	ctlCmd := newCtlCommand()
 	ctlCmd.Hidden = !developer.Mode()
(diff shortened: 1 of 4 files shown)


cmd/rdd/main.go
```diff
@@ -88,11 +127,22 @@ func main() {
...
+			setLogrusLevel(logLevel)
+			return nil
+		},
 	}
 
 	// Add version flag with verflag-compatible behavior
 	version.AddVersionFlag(cmd.Flags())
 
+	// Add global log-level flag
+	cmd.PersistentFlags().String("log-level", "", "Log level: error, warn, info, debug, trace")
```
mook-as: Instead of a string, I think it would be clearer to implement [`pflag.Value`](https://pkg.go.dev/github.com/spf13/pflag#Value) for [`logrus.Level`](https://pkg.go.dev/github.com/sirupsen/logrus#Level) (that is, have a wrapper type that implements it)?

cmd/rdd/main.go
```diff
@@ -66,6 +66,45 @@ func parseInstanceFlag() {
...
+	} else {
+		// Default log level: warn for regular mode, debug for developer mode
+		if developer.Mode() {
+			logLevel = "debug"
+		} else {
+			logLevel = "warn"
+		}
+	}
+
+	// Validate the log level
```
mook-as: At least use [`logrus.ParseLevel`](https://pkg.go.dev/github.com/sirupsen/logrus#ParseLevel) for this?

mook-as (commented):
FWIW, I had c3efe5b485d07b75100671a942a312b1e1e45cb5 but that doesn't seem any better than what you're doing (I have no idea what I'm doing), so taking this is probably better.

Could you write a quick overview (in the PR is fine, doesn't need to be in the code) on how we're using _both_ `logrus` and `klog`? Standardizing on one seems to be saner here (and we don't seem to use `logrus` beyond the two `cmd/…` files).

jandubois:
> Could you write a quick overview (in the PR is fine, doesn't need to be in the code) on how we're using _both_ `logrus` and `klog`? Standardizing on one seems to be saner here (and we don't seem to use `logrus` beyond the two `cmd/…` files).

We are using `klog` because it is what Kubernetes is using, so any logging done by `rdd svc serve` needs to use `klog` to write to `rdd.stderr.log`.

All other logging typically goes to the console and uses `logrus`. I can't remember why I thought `klog` would not be appropriate for it, except I do not want to have to use the Kubernetes options parsing code for our regular commands.

I might have picked `logrus` because it is being used by Lima for all logging, so at some level (at least in the `limavm` controllers) we are going to use it anyways.

`logrus` is also used by `k3s`, which is conceptually similar to `rdd`, but I'm not sure it is relevant.

cmd/rdd/main.go
jandubois: I thought I replied here before, but can't see my message anymore.

Why would a custom pflag.Value be clearer? Wouldn't it just be a lot of extra methods to do the same thing the string validation already does?

mook-as: Mainly I want to push off the list of valid levels to be somebody else's problem (in that having a list is another place to introduce a bug).

But I guess we're not going to ever _touch_ the list, so once it's there things are fine.

cmd/rdd/service.go
```diff
@@ -16,6 +16,18 @@ import (
 	service "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/service/cmd"
 )
 
+// logrusLevelToKlog converts current logrus level to klog numeric level.
+func logrusLevelToKlog() int {
```
mook-as: All callers pass this to `Itoa`, so that seems kind of silly…

mook-as (dismissed):

cmd/rdd/service.go
jandubois: I've now generated the list of levels from `logrus.AllLevels`, and include `fatal` in the list (hide errors), but omit `panic`, as that seems silly (you can still specify it).

It now lists `warning` instead of `warn`, but I guess this is fine.

As for the klog level, it is conceptually an int, which is why I used that type, and only converted it for marshalling to JSON. But I don't really care, so I changed it to a string now.

mook-as (dismissed):

mook-as (approved):

PR #23, opened 2025-10-02
Refactor the notary-events BATS tests

jandubois:
This is based on a older refactoring that got lost in the rebase shuffle.

The previous events tests are not actually testing the right things, as you can't test that no events are generated by setting the value repeatedly. The calls get eliminated by the apiserver before they even reach the admission controller. The way to test this is by making some other change to the resource, like adding a label or annotation. Those can then generate duplicate events, that in turn can be deduplicated by Kubernetes again.

But if we never produce duplicate events, then we cannot prove that they would be de-duplicated.

### DIFF
--- a/bats/tests/31-rdd-controllers/notary-events.bats
+++ b/bats/tests/31-rdd-controllers/notary-events.bats
@@ -61,43 +61,36 @@ wait_for_events() {
     try --max 10 --delay 1 -- assert_events_exist "${resource_name}" "${reason}"
 }
 
-assert_events_after_timestamp() {
-    local resource_name=$1
-    local after_timestamp=$2
-    local reason=$3
-
-    run -0 rdd ctl get events  --field-selector involvedObject.name="${resource_name}" -o json
-    run -0 jq ".items | map(select(.reason == \"$reason\" and .lastTimestamp > \"$after_timestamp\")) | length" <<<"$output"
-    refute_output 0
-}
-
-count_events_with_reason() {
-    local resource_name=$1
-    local reason=$2
-
-    run rdd ctl get events  --field-selector involvedObject.name="${resource_name}" -o json
-    if [ "$status" -ne 0 ]; then
-        echo "0"
-        return
-    fi
-
-    jq ".items | map(select(.reason == \"$reason\")) | length" <<<"$output"
-}
-
 get_events_after_timestamp() {
     local resource_name=$1
-    local after_timestamp=$2
+    local timestamp=$2
     local reason=$3
-
-    try --max 10 --delay 1 -- assert_events_after_timestamp "${resource_name}" "${after_timestamp}" "${reason}"
 
     run rdd ctl get events --field-selector involvedObject.name="${resource_name}" -o json
     if [ "$status" -ne 0 ]; then
-        echo ""
-        return 1
+        echo "[]"
+        return 0
     fi
 
-    jq -r ".items | map(select(.lastTimestamp > \"$after_timestamp\" and .reason == \"$reason\"))" <<<"$output"
+    jq -r ".items | map(select(.lastTimestamp > \"$timestamp\" and .reason == \"$reason\"))" <<<"$output"
+}
+
+assert_events_after_timestamp() {
+    local resource_name=$1
+    local timestamp=$2
+    local reason=$3
+
+    run -0 get_events_after_timestamp "$resource_name" "$timestamp" "$reason"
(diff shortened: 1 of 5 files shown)


jandubois:
Instead of looking at the diff, it may be easier to review by just reading the new code on its own and checking if it makes sense and seems complete.

bats/tests/31-rdd-controllers/notary-events.bats
```diff
@@ -63,41 +63,34 @@ wait_for_events() {
 
 assert_events_after_timestamp() {
```
mook-as: Shouldn't this be re-written in terms of `get_events_after_timestamp`?

bats/tests/31-rdd-controllers/notary-events.bats
```diff
@@ -143,88 +138,44 @@ get_latest_event_timestamp() {
...
-    wait_for_events "anno" "ValueRecorded"
+    # Count events after the timestamp - should be deduplicated by Kubernetes
+    # It is not clear if Kubernetes will always catch all duplicates, but it should get at least 1
+    run -0 get_events_after_timestamp "dedup" "$timestamp" "SpecUpdate"
+    assert_output --partial "constant-value"
 
-    # Get the timestamp of the most recent event before annotation update
-    run -0 get_latest_event_timestamp "anno"
-    local timestamp=$output
+    run -0 jq 'length < 4' <<<"$output"
```
mook-as: Would doing the assert in BATS be clearer?
```bats
run -0 jq length <<<"$output"
assert [ $output -lt 4 ]
```
This depends on which would print out the length (so we can eyeball it), of course.
Maybe `assert_not_equal "$output" 4`?

---
I'm kind of skeptical of the _should_ catch at least 1 duplicate here, especially given that the tests were magically passing before (where as deduplication _should_ have caused it to fail, if I understand correctly).  But I don't really have any better ideas either, so I guess we stick with this for now.

mook-as (changes requested):
Great, that was the exact test I had issues on Windows (and I deferred looking into it because it didn't make sense to me), now I'm glad I did that… :D

Still not sure why it passes on mac and linux though.

bats/tests/31-rdd-controllers/notary-events.bats
jandubois: > Would doing the assert in BATS be clearer?

Yes, I guess it would. This is actually the change that added `assert_output_lt` in `numeric.bash`. And the fact that I couldn't find it made me search for it in the reflog. But now I found there were only 4 uses of it, and I eliminated 2 of them, so it felt like overkill. But I still have the files, so will add them back.

The functions use the helpers from `bats-assert` so do produce nice debugging output:

```bash
@test '4 < 2' {
    output=4
    assert_output_lt 2
}
```

produces

```
 ✗ 4 < 2 [28]
   (from function `assert_output_lt' in file bats/helpers/numeric.bash, line 71,
    in test file bats/tests/31-rdd-controllers/notary-external.bats, line 23)
     `assert_output_lt 2' failed

   -- output is not less than expected maximum --
   expected : < 2
   actual   : 4
   --
```

bats/tests/31-rdd-controllers/notary-events.bats
jandubois: Yes, I missed that.

PTAL

bats/tests/31-rdd-controllers/notary-events.bats
```diff
@@ -61,43 +61,35 @@ wait_for_events() {
...
-get_events_after_timestamp() {
+assert_events_after_timestamp() {
     local resource_name=$1
-    local after_timestamp=$2
+    local timestamp=$2
     local reason=$3
 
-    try --max 10 --delay 1 -- assert_events_after_timestamp "${resource_name}" "${after_timestamp}" "${reason}"
+    get_events_after_timestamp "$resource_name" "$timestamp" "$reason"
+    refute_output 0
```
mook-as, line 85: `get_events_after_timestamp` returns a JSON array, right? How would it ever be `0`?
Also, it's not using `run`, so `$output` wouldn't be set?

mook-as (changes requested):

bats/tests/31-rdd-controllers/notary-events.bats
jandubois: Indeed, 2 bugs 😞 

Should be fixed now

mook-as (approved):

PR #25, opened 2025-10-02
CI: Add spelling

mook-as:
This adds a `make spelling` target that runs the same `check-spelling` script that Rancher Desktop 1.x uses.

Note that it doesn't really understand golang imports (where it's typical to mush a bunch of words together as the import name), so there's some code to auto-generate a word list based on that.  That's messier than I'd like, but that would remain until we find a spell checker that has more language-specific support.

I added the `en-CA` dictionary to work around an issue with the spell checker on macOS, where it would randomly ignore some files (‽) and therefore decide that the word _tarball_ is unused, and then flag _tarballs_ (plural) as an unknown word.  I realize that dictionary _probably_ doesn't contain the relevant word, but it jiggles the complex system enough to shut it up.

### DIFF
--- a/scripts/check-spelling.sh
+++ b/scripts/check-spelling.sh
@@ -1,0 +1,112 @@
+#!/usr/bin/env bash
+
+# This script runs check-spelling on the parent repository.
+
+set -o errexit -o nounset
+
+check_prerequisites() {
+    case $(uname -s) in # BSD uname doesn't support long option `--kernel-name`
+        Darwin) check_prerequisites_darwin;;
+        Linux) check_prerequisites_linux;;
+        CYGWIN*|MINGW*|MSYS*) check_prerequisites_windows;;
+        *) printf "Prerequisites not checked on %s\n" "$(uname -s)" >&2 ;;
+    esac
+}
+
+check_prerequisites_darwin() {
+    if command -v cpanm &>/dev/null; then
+        return
+    fi
+    echo "Please install cpanminus first:" >&2
+    if command -v brew &>/dev/null; then
+        echo "brew install cpanminus" >&2
+    fi
+    exit 1
+}
+
+check_prerequisites_linux() {
+    if ! command -v cpanm &>/dev/null; then
+        echo "Please install cpanminus first:" >&2
+        if command -v zypper &>/dev/null; then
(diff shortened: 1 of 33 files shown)


scripts/spell-check-generate-golang-expect.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil
+	}
+
+	f, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("failed to open %s: %w", path, err)
+	}
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	matcher := regexp.MustCompile(`^\s*(?:import\s+)?(?:(\S+)\s+)?"([^"]+)"$`)
```
jandubois, line 51: I guess this is good enough, given that we are also in control of writing the import statements. I just generally don't like to parse source code with regular expressions instead of using the AST (or tools provided by the language).

I could produce the list of imports with

```shell
go list -f '{{.Imports}}' ./... | tr -d '[]' | tr ' ' '\n' | sort -u
```

But I guess that doesn't fit well inside the Go program.

scripts/check-spelling.sh
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env bash
+
+# This script runs check-spelling on the parent repository.
```
jandubois, line 3: Did you mean "directory"? Not sure what the parent repository is.

jandubois (approved):
Thanks, LGTM

scripts/spell-check-generate-golang-expect.go
mook-as: More importantly, it doesn't show the import package name alias things:
https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/159d3943b71ceae4371eb2f7db5f72cc4eb6e531/pkg/controllers/rdd/configmapreplicaset/controller.go#L10
That should add `ctrl` somewhere because that's a word we use.

scripts/check-spelling.sh
mook-as: That would be `rancher-desktop-daemon`, i.e. something that's often known as `$topsrcdir`.

PR #26, opened 2025-10-02
Add license tag checking

mook-as:
Turns out I missed it when creating two files, so having this would be a good thing.


jandubois:
> Turns out I missed it when creating two files, so having this would be a good thing.

We also need to fix `ltag` to support additional file types (like `.bats`).

jandubois (dismissed):
Thanks, LGTM

jandubois (approved):

PR #28, opened 2025-10-03
Fix PATH setting in BATS

jandubois:
The value must be exported so it stays in effect even if rdd is invoked via env, e.g. `env RDD_DEVELOPER_MODE=1 rdd ...`.

### DIFF
--- a/bats/helpers/load.bash
+++ b/bats/helpers/load.bash
@@ -55,9 +55,8 @@ source "$PATH_BATS_HELPERS/paths.bash"
 # and PATH_* variables from paths.bash
 source "$PATH_BATS_HELPERS/commands.bash"
 
-# Add BATS helper executables to $PATH.  On Windows, we use the Linux version
-# from WSL.
-export PATH="$PATH_BATS_ROOT/bin/${OS/windows/linux}:$PATH"
+# Add repo-root/bin directory to the PATH. This is where the Makefile puts all compiled programs.
+export PATH="$PATH_BATS_ROOT/../bin:$PATH"
 
 # If called from foo() this function will call local_foo() if it exist.
 call_local_function() {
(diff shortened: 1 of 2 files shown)


bats/helpers/load.bash
```diff
@@ -57,7 +57,7 @@ source "$PATH_BATS_HELPERS/commands.bash"
 
 # Add BATS helper executables to $PATH.  On Windows, we use the Linux version
 # from WSL.
-export PATH="$PATH_BATS_ROOT/bin/${OS/windows/linux}:$PATH"
+export PATH="$PATH_BATS_ROOT/bin:$PATH"
```
mook-as: If I'm reading this correctly:
https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/cefdf7551ffb880dac1212e53b0136f3ba3f8528/bats/helpers/load.bash#L15-L17

Variable | Value
--- | ---
`$PATH_BATS_HELPERS` | `…/rancher-desktop-daemon/bats/helpers`
`$PATH_BATS_ROOT` | `…/rancher-desktop-daemon/bats`

So we're adding `…/rancher-desktop-daemon/bats/bin` (which doesn't exist) to the `$PATH`? It only worked before because this line wasn't actually _necessary_, I think…

mook-as (approved):
LGTM

Side comment: in WSL, we don't need to expose the `$PATH` change to the Windows side, so we don't need to modify `$WSLENV`. (With experience from before, not touching that would be good because it would clobber the Windows-side paths.)

bats/helpers/load.bash
mook-as: I thought that comment disappeared; this is no longer relevant by the time I submitted the comment.

PR #29, opened 2025-10-07
Add generate-crds and generate-deepcopy targets to the Makefile

jandubois:
Also regenerate CRD with latest controller-gen. Updates ConfigMapReplicaSet CRD from GoDoc strings.

I also had to upgrade Kubernetes to 1.34. I couldn't manage to add the latest controller-gen as tool with 1.33 because of package conflicts.


scripts/generate-crds.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+set -o errexit -o nounset
+
+API_GROUPS=$(find pkg/controllers -type d -mindepth 1 -maxdepth 1 -not -name base -exec basename {} \;)
```
mook-as: Hmm, I guess macOS `find` doesn't support `-printf %f`. Ah well.

mook-as: Hmm, I guess macOS `find` doesn't support `-printf %f`. Ah well.

scripts/generate-crds.sh
```diff
@@ -0,0 +1,29 @@
...
+
+TMPDIR=$(mktemp -d)
+trap 'rm -rf $TMPDIR' EXIT
```
mook-as: I'd prefer `trap "rm -rf '$TMPDIR'"` (i.e. expand early, single quote), but I think shellcheck hates it so this should be fine; might still want to quote though.

scripts/generate-crds.sh
```diff
@@ -0,0 +1,29 @@
...
+for apigroup in $API_GROUPS; do
+	# Generate all CRDs for this API group to temp directory
+	go tool controller-gen crd "paths=./pkg/apis/${apigroup}/..." "output:crd:dir=${TMPDIR}"
+
+	# Distribute CRDs to their respective controller directories
+	for controller_dir in "pkg/controllers/${apigroup}"/*/; do
+		if [ -f "$controller_dir/crd.yaml" ]; then
+			controller=$(basename "$controller_dir")
+			# Try plural form first (e.g., configmapreplicasets), then singular (e.g., notary)
+			crd_file=$(ls "$TMPDIR"/*_"${controller}"s.yaml 2>/dev/null || ls "$TMPDIR"/*_"${controller}".yaml 2>/dev/null || echo "")
```
mook-as: Or `find "$TMPDIR" '(' -name "*_${controller}s.yaml" -o -name "*_${controller}.yaml" ')' -print -a -quit`? I'd probably do that in a `while read` loop though.

Makefile
```diff
@@ -87,6 +87,18 @@ test-all-controllers: $(addprefix test-,$(addsuffix -controllers,$(API_GROUPS)))
 run-all-controllers: $(addprefix run-,$(addsuffix -controller,$(API_GROUPS)))
 .PHONY: run-all-controllers
 
+# Code generation targets
+generate-deepcopy:
+	scripts/generate-deepcopy.sh
```
mook-as, line 92:
```suggestion
generate-deepcopy: scripts/generate-deepcopy.sh
	$<
```

This definitely depends on the script :)

mook-as (dismissed):
Some optional comments.

Makefile
jandubois: I don't think having source dependencies on `.PHONY` targets makes sense.

It can make sense if the dependency needs to be built too, and you rely on the side-effects of the build. But I don't see how a dependency on a source file with no rules can make a difference.

scripts/generate-crds.sh
jandubois: Yes, `shellcheck` is the reason for the single-quotes here. I'll add a `readonly TMPDIR`.

scripts/generate-crds.sh
jandubois: I'll use this instead:

```
API_GROUPS=$(cd  pkg/controllers; ls -d */ | grep -v base | tr -d /)
```

pkg/controllers/app/demo/crd.yaml
mook-as, line 1: Actually, if these are generated, why do we check them in? Wouldn't just generating them as a pre-requisite of building whatever else been sufficient?

scripts/generate-crds.sh
```diff
@@ -0,0 +1,36 @@
...
+
+for apigroup in $API_GROUPS; do
+	# Generate all CRDs for this API group to temp directory
+	go tool controller-gen crd "paths=./pkg/apis/${apigroup}/..." "output:crd:dir=${TMPDIR}"
+
+	# Distribute CRDs to their respective controller directories
+	for controller_dir in "pkg/controllers/${apigroup}"/*/; do
+		if [ -f "$controller_dir/crd.yaml" ]; then
+			controller=$(basename "$controller_dir")
+			crd_file=$(grep --ignore-case --files-with-matches "^ *kind: ${controller}$" "$TMPDIR"/*.yaml || :)
```
mook-as, line 27: I guess we don't want to demand `yq` exist?

The other option is to convert this to golang because the shell script is getting unwieldy…

mook-as (approved):
I'm kind of suspicious of the `grep`, but if you think that's enough we can wait until it breaks.

pkg/controllers/app/demo/crd.yaml
jandubois: Maybe, but mixing `make` and `go build` is already kind of weird because `go build` wants to track its own dependencies. Just like you typically check in all the files generated by protobuf.

I also think it is good to have the CRDs as source in the repo, so you can see diffs in `git log`.

scripts/generate-crds.sh
jandubois: We will eventually have `rdd yq` because I seem to end up embedding it everywhere. There is already `limactl yq` and `lima-guestagent yq`. 😄 

For now the `grep` should be good enough.

PR #30, opened 2025-10-09
Add admission controller to LimaVM

jandubois:
Started out as a copy of the Notary admission controller, but then shared functionality has been moved into a base implementation.

Also adds boolean `running` property to LimaVM.

Similar cleanup to the Notary admission controller BATS test will come in a separate PR.


bats/helpers/utils.bash
```diff
@@ -117,12 +117,16 @@ join_map() {
...
 }
 
-# Run jq on the current $output
+# Extract raw value from a JSON object in a string (last argument).
 # Note that when capturing $output, you may need to use `run --separate-stderr`
 # to avoid also capturing stderr and ending up with invalid JSON.
-jq_output() {
-    local json=$output
-    run jq --raw-output "$@" <<<"${json}"
+jq_raw() {
```
mook-as, line 123: Hmm, if `$json` is the first arg, we can just `shift`; but I'm not sure that's a better API for the caller. I guess leave it.

bats/tests/33-lima-controllers/limavm-admission.bats
```diff
@@ -0,0 +1,158 @@
+load '../../helpers/load'
+
+# LimaVM admission controller tests - tests webhook validation for cross-namespace
+# uniqueness enforcement. LimaVM names must be unique across all namespaces
+# because they correspond to actual VM instances on the host system.
```
mook-as, line 5: That sounds like LimaVM objects shouldn't be namespaced… But the docs explain why, so that's okay.

bats/tests/33-lima-controllers/limavm-admission.bats
```diff
@@ -0,0 +1,158 @@
...
+
+# Usage: apply_limavm <name> <namespace> [labels] [extra_args]
+# Example: apply_limavm "my-vm" "test-ns1" "updated=true" "--dry-run=server"
+apply_limavm() {
+    local name=$1
+    local namespace=$2
+    local labels=${3:-}
+
+    run -0 create_limavm_yaml "${name}" "${namespace}" "${labels}"
+    rdd ctl apply -f - <<<"$output"
```
mook-as, line 49: Did you mean to add `extra_args` here?

bats/tests/33-lima-controllers/limavm-admission.bats
```diff
@@ -0,0 +1,158 @@
...
+}
+
+# create_limavm_yaml <name> <namespace> [labels]
+# Example: create_limavm_yaml "my-vm" "test-ns1" "updated=true"
+create_limavm_yaml() {
+    local name=$1
+    local namespace=$2
+    local labels=${3:-}
+
+    cat <<EOF
```
mook-as, line 21: I think this would be way more readable as an `yq` expression, but I guess we can't assume that's available?

bats/tests/33-lima-controllers/limavm-admission.bats
```diff
@@ -0,0 +1,158 @@
...
+}
+
+@test "create LimaVM in first namespace succeeds" {
+    # Create first LimaVM in test-ns1
+    run -0 apply_limavm "my-vm" "test-ns1"
+    assert_output --partial "my-vm created"
+
+    # Verify the LimaVM was created
+    run -0 rdd ctl get limavm my-vm -n test-ns1
+    assert_output --partial "my-vm"
```
mook-as: Would using `rdd ctl get … --output name` and then `assert_line limavm.lima.rancherdesktop.io/my-vm` (or whatever the correct value is) be better?

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,195 @@
...
+	// APIGroup is the API group for the webhook rules (e.g., "rdd.rancherdesktop.io")
+	APIGroup string
+
+	// APIVersion is the API version for the webhook rules (e.g., "v1alpha1")
+	APIVersion string
+
+	// Resource is the resource type for the webhook rules (e.g., "notaries")
+	// TODO: Consider deriving this automatically from the CRD or RESTMapper to maintain
+	// a single source of truth. Currently must match the plural name in the CRD.
+	Resource string
```
mook-as, line 60: So I looked it up.
This is just the type name ("Notary") run through https://pkg.go.dev/github.com/gobuffalo/flect#section-readme and lower cased ("notaries") when the CRD is generated. This detail is probably not useful…

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// WebhookManager handles the creation and management of ValidatingWebhookConfigurations.
+type WebhookManager struct {
+	config WebhookConfig
+	mgr    ctrl.Manager
+}
+
+// NewWebhookManager creates a new WebhookManager with the given configuration.
+func NewWebhookManager(config WebhookConfig, mgr ctrl.Manager) *WebhookManager {
```
mook-as: Does this need to exist? The only caller is in this function, and it just returns a new instance using the given parameters. That can just be inlined.

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,195 @@
...
+		config: config,
+		mgr:    mgr,
+	}
+}
+
+// SetupWebhookForResource is a helper function that sets up a webhook for a resource type.
+// It registers the webhook with controller-runtime and creates a WebhookManager for later setup.
+// The validator parameter must implement the admission.CustomValidator interface.
+// Returns the created WebhookManager for storage in the controller.
+func SetupWebhookForResource[T client.Object, V admission.CustomValidator](mgr ctrl.Manager, obj T, validator V, config WebhookConfig) (*WebhookManager, error) {
```
mook-as: Does this need to be generic at all?

```suggestion
func SetupWebhookForResource(mgr ctrl.Manager, obj client.Object, validator admission.CustomValidator, config WebhookConfig) (*WebhookManager, error) {
```

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,195 @@
...
+// This blocks until the webhook is successfully registered or max attempts are exceeded.
+// This should be called after the webhook server is registered with the manager.
+func (wm *WebhookManager) Setup() error {
+	klog.Info("Starting webhook configuration creation...")
+
+	const maxAttempts = 20
+	const retryDelay = 3 * time.Second
+
+	for attempt := 1; ; attempt++ {
+		klog.Infof("Webhook configuration creation attempt %d/%d", attempt, maxAttempts)
```
mook-as, line 110: I believe `mgr` has a `GetLogger()`; is that useful?

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,195 @@
...
+	// The certificates were already created by SharedControllerManager.
+	caBundlePath := filepath.Join(instance.TLSDir(), DefaultWebhookCACertFileName)
+	klog.Infof("Reading CA bundle from: %s", caBundlePath)
+	caBundleBytes, err := os.ReadFile(caBundlePath)
+	if err != nil {
+		klog.Errorf("Failed to read CA bundle file: %v", err)
+		return fmt.Errorf("failed to read CA bundle file %s: %w", caBundlePath, err)
+	}
+
+	webhookURL := fmt.Sprintf("https://127.0.0.1:%d%s", wm.config.WebhookPort, wm.config.WebhookPath)
```
mook-as, line 140: I guess we don't need `net.JoinHostPath` because we hard code the host?

pkg/controllers/lima/limavm/controller.go
```diff
@@ -24,14 +30,28 @@ const ControllerName = "limavm"
 // APIGroup is the API group this controller belongs to.
 const APIGroup = "lima"
 
+// Webhook configuration constants.
+const (
+	// WebhookName is the name used for the webhook configuration.
+	WebhookName = "limavm.lima.rancherdesktop.io"
+	// ValidatorConfigName is the name of the ValidatingWebhookConfiguration.
+	ValidatorConfigName = "limavm-validator"
```
mook-as: Do these need to be exported? I don't think anybody else uses them.

pkg/controllers/lima/limavm/controller.go
```diff
@@ -48,22 +68,108 @@ func (c *Controller) GetAPIGroup() string {
...
+		FinalizerHelper: base.NewFinalizerHelper(mgr.GetClient(), mgr.GetScheme(), controllers.FinalizerName),
+	}).SetupWithManager(mgr)
+}
+
+// setupWebhookWithRuntimeConfig sets up webhook with shared certificate configuration.
+func (c *Controller) setupWebhookWithRuntimeConfig(mgr ctrl.Manager) error {
+	webhookConfig := base.WebhookConfig{
+		Name:        ValidatorConfigName,
+		WebhookName: WebhookName,
+		WebhookPath: "/validate-lima-rancherdesktop-io-v1alpha1-limavm",
```
mook-as: This is, effectively, arbitrary, right? Any reason `base.WebhookConfig` can't just generate something automatically?

pkg/controllers/lima/limavm/validation.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/lima/v1alpha1"
+)
+
+// ValidateLimaVMUniqueName validates that the LimaVM name is unique across all namespaces.
+// This is critical because LimaVM names correspond to actual VM instances on the host system,
+// which must be unique.
+func ValidateLimaVMUniqueName(ctx context.Context, c client.Client, limavm *v1alpha1.LimaVM) error {
```
mook-as: Any particular reason these functions are exported? I guess it might make sense if you wanted to unit test them, but I don't see a test file.

pkg/controllers/lima/limavm/validation.go
```diff
@@ -0,0 +1,60 @@
...
+	// List all LimaVMs across all namespaces
+	limavmList := &v1alpha1.LimaVMList{}
+	if err := c.List(ctx, limavmList, &client.ListOptions{}); err != nil {
+		return fmt.Errorf("failed to list LimaVMs for uniqueness check: %w", err)
+	}
+
+	// Check for name conflicts in other namespaces
+	for _, existingVM := range limavmList.Items {
+		// Skip if it's an update to the existing resource (same namespace and name)
+		if existingVM.Namespace == limavm.Namespace && existingVM.Name == limavm.Name {
```
mook-as: You don't actually need to check the name here (because if the name is different, it doesn't violate the rule).

pkg/controllers/lima/limavm/validation.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+	}
+
+	return nil
+}
+
+// ValidateLimaVM validates a complete LimaVM object and returns warnings.
+func ValidateLimaVM(ctx context.Context, c client.Client, limavm *v1alpha1.LimaVM) ([]string, error) {
+	if limavm == nil {
+		return nil, errors.New("limavm object cannot be nil")
```
mook-as, line 47: This is a duplicate check, but it's probably harmless.

pkg/service/controllers/manager.go
```diff
@@ -195,8 +200,37 @@ func (scm *SharedControllerManager) Start(ctx context.Context) error {
...
+	webhookWaitGroup.Wait()
+	close(webhookErrors)
+
+	var errors []error
+	for err := range webhookErrors {
+		errors = append(errors, err)
+	}
+
+	if len(errors) > 0 {
+		return fmt.Errorf("webhook setup errors: %v", errors)
```
mook-as: Consider using [`errors.Join`](https://pkg.go.dev/errors@go1.25.2#Join):

```suggestion
		return fmt.Errorf("webhook setup encountered errors: %w", errors.Join(errs))
```
(That means the variable needs to be renamed…)

mook-as (changes requested):
Generally seems fine, just some minor things.

bats/helpers/utils.bash
jandubois: Yeah, I was wondering about that too, but you normally write `jq -r '.result[].field' data.json`, so I wanted to keep the same order. Doesn't really matter, and if I hadn't figured out how to drop the last element, then I would have changed the order.

bats/tests/33-lima-controllers/limavm-admission.bats
jandubois: I had `extra_args` at first, but found it awkward, so I removed it. Looks like I didn't update the comment.

This changes again in #32 when this all goes away and the test uses `rdd limavm create ...` instead.

bats/tests/33-lima-controllers/limavm-admission.bats
jandubois: I'm fine with requiring `yq`, but I'm not sure it would be more readable. Maybe show me how you would write it.

This specific case goes away with the switch to `rdd limavm create`, but there is almost identical code in the `notary-admission.bats` (which changes in #31, so maybe put your suggestion there).

pkg/controllers/base/webhook.go
jandubois: Yes, but I don't want to derive the name independently, I just want to look it up, so I **know** I have the right string.

We can parse `base.GetCRDData()` to grab it, or use `//go:generate` at build time.

jandubois: And we have to look it anyways because I had to override the plural of `demo` to be `demos` and not `demoes`, which is what `controller-gen` wants to use.

`Demo` will be renamed to `App` soon, but the principle still applies that we should not calculate the name on our own.

pkg/controllers/base/webhook.go
jandubois: It is just a standard pattern I see everywhere, but I'm fine inlining it.

jandubois: Actually, I changed my mind. I find the trailing `, nil` kind of ugly/confusing. It would now look like this:

```go
	return &WebhookManager{
		config: config,
		mgr:    mgr,
	}, nil
```

pkg/controllers/lima/limavm/controller.go
jandubois: They are arbitrary, but since we use the webhook builder from the controller runtime, we need to match the names it generates. I'll add a function to generate them instead.

bats/tests/33-lima-controllers/limavm-admission.bats
mook-as: Something like…
```bash
# create_limavm_yaml <name> <namespace> [label] [label...]
# Example: create_limavm_yaml "my-vm" "test-ns1" "updated=true"
create_limavm_yaml() {
    local name=$1
    local namespace=$2
    shift 2
    local expressions=(
        '.apiVersion="lima.rancherdesktop.io/v1alpha1"'
        '.kind="LimaVM"'
        ".metadata.name=\"$name\""
        ".metadata.namespace=\"$namespace\""
        '.spec={}'
    )
    local label
    for label in "$@"; do
        expressions+=(".metadata.labels.${label}")
    done
    local expression
    expression=.$(printf "|%s" "${expressions[@]}")

    yq --null-input --expression "${expression}"
}
```

The main improvement is just that there's no whitespace-sensitive string munging invlved.

Note that this would require changing how labels are set:
```bash
run -0 apply_limavm "my-vm" "test-ns1" 'updated="true"'
```

Alternatively, just make extra arguments yq expressions instead, so it would be:
```bash
run -0 apply_limavm "my-vm" "test-ns1" 'metadata.labels={"updated":"true"}'
```

pkg/controllers/base/webhook.go
jandubois: Possibly, but I suspect it will end up being the same as just using `klog.Info` directly. I'll look into it later.

pkg/controllers/base/webhook.go
mook-as: If you want to keep it, at least make it unexported; that's just confusing. I still prefer it not existing, but I'm not going to fight over that.

pkg/controllers/base/webhook.go
jandubois: I thought so, I had some compilation error before, but I guess I was confused.

pkg/service/controllers/manager.go
jandubois: Will do, but the errors are just strings coming through a channel, so it won't make much of a difference. And it is only for debugging. Missing webhook registration will be a fatal error aborting the control plane.

bats/tests/33-lima-controllers/limavm-admission.bats
mook-as, line 1: Do you need to edit the Makefile to add a `bats-lima` target?

jandubois: Yes, I forgot and added it in #32.

pkg/controllers/base/webhook.go
```diff
@@ -0,0 +1,209 @@
...
+
+// GenerateValidatingWebhookPath generates the webhook path that controller-runtime uses
+// when registering a validating webhook for a given API group, version, and kind.
+// This follows controller-runtime's convention: /validate-{group}-{version}-{kind}
+// For core resources (empty group), the path becomes: /validate--{version}-{kind}.
+// Dots in the API group are replaced with dashes.
+//
+// Example: GenerateValidatingWebhookPath("lima.rancherdesktop.io", "v1alpha1", "limavm")
+// Returns: "/validate-lima-rancherdesktop-io-v1alpha1-limavm".
+func GenerateValidatingWebhookPath(apiGroup, apiVersion, kind string) string {
```
mook-as, line 35: The only thing that's not already known in a `WebhookConfig` is the `kind`? Any reason to not just use `WebhookConfig#Name` (or `WebhookName`) and let the caller ignore it?

(I couldn't find anywhere that actually builds this string from a different code path; as far as I can see it's just used to both register and respond to requests?)

It's fine to think about it in the future, though.

bats/Makefile
```diff
@@ -46,8 +46,13 @@ bats-app: check-bats build-rdd
 	RDD_INSTANCE=bats-app-controller $(BATS_CORE) tests/32-app-controllers/
 .PHONY: bats-app
 
+# Run lima API group controller tests
+bats-lima: check-bats
```
mook-as: Doesn't this need to depend on `build-lima-controller`?

mook-as (changes requested):

mook-as (approved):

PR #31, opened 2025-10-09
Simplify notary-admission.bats

jandubois:
There was quite a bit of redundancy in the old tests, making them tedious to maintain long term.

This PR include #30, so needs to be rebased once that one has been merged.


mook-as (approved):
We're removing a bunch of test cases, but none of them look very important.

PR #32, opened 2025-10-09
Add lima create/start/stop/delete subcommands

jandubois:
All take the VM name as an argument. "create" also takes a namespace.

Built on top of #30 and #31, so must be rebased once those are merged.

### DIFF
--- a/cmd/rdd/limavm.go
+++ b/cmd/rdd/limavm.go
@@ -1,0 +1,196 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	limav1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/lima/v1alpha1"
+	service "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/service/cmd"
+)
+
+func newLimaCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:     "limavm",
+		Short:   "Manage LimaVM resources",
+		Long:    "Create, start, stop, and delete LimaVM virtual machines",
+		Aliases: []string{"lima"},
+	}
+	command.AddCommand(
+		newLimaCreateCommand(),
+		newLimaStartCommand(),
+		newLimaStopCommand(),
+		newLimaDeleteCommand(),
+	)
+	return command
+}
+
+func newLimaCreateCommand() *cobra.Command {
+	var namespace string
+	command := &cobra.Command{
+		Use:   "create NAME",
+		Short: "Create a new LimaVM resource",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return limaCreateAction(cmd.Context(), args[0], namespace)
+		},
+	}
+	command.Flags().StringVarP(&namespace, "namespace", "n", "default", "Namespace for the LimaVM resource")
+	return command
+}
+
+func newLimaStartCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "start NAME",
+		Short: "Start a LimaVM by setting running=true",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return limaSetRunningAction(cmd.Context(), args[0], true)
+		},
+	}
+	return command
+}
+
+func newLimaStopCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "stop NAME",
+		Short: "Stop a LimaVM by setting running=false",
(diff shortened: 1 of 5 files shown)


jandubois:
```console
❯ rdd ctl get limavms
No resources found in default namespace.

❯ rdd lima create vm1
INFO[0000] LimaVM "vm1" created in namespace "default"

❯ rdd lima create vm2
INFO[0000] LimaVM "vm2" created in namespace "default"

❯ rdd lima start vm2
INFO[0000] LimaVM "vm2" started in namespace "default"

❯ rdd ctl get limavms
NAME   RUNNING
vm1    false
vm2    true
```

I'm considering renaming the command from `lima` to `limavm`.

EDIT: I've renamed it, but kept `lima` as an alias.

cmd/rdd/limavm.go
```diff
@@ -0,0 +1,196 @@
...
+// findLimaVM searches for a LimaVM with the given name across all namespaces.
+func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha1.LimaVM, error) {
+	vmList := &limav1alpha1.LimaVMList{}
+	if err := c.List(ctx, vmList); err != nil {
+		return nil, fmt.Errorf("failed to list LimaVMs: %w", err)
+	}
+
+	var found *limav1alpha1.LimaVM
+	for i := range vmList.Items {
+		if vmList.Items[i].Name == name {
```
jandubois: This should never happen if the admission controller works properly, so maybe we can remove it.

cmd/rdd/limavm.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	command.Flags().StringVarP(&namespace, "namespace", "n", "default", "Namespace for the LimaVM resource")
+	return command
+}
+
+func newLimaStartCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "start NAME",
+		Short: "Start a LimaVM by setting running=true",
+		Args:  cobra.ExactArgs(1),
```
mook-as, line 56: There no way to set `--namespace` on the `start`, `stop`, and `delete` commands?

Would it make sense to put that on `limavm` itself, instead of on the sub-commands?

cmd/rdd/limavm.go
```diff
@@ -0,0 +1,196 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return limaDeleteAction(cmd.Context(), args[0])
+		},
+	}
+	return command
+}
+
+// getKubeClient returns a controller-runtime client configured for the RDD control plane.
+func getKubeClient(ctx context.Context) (client.Client, error) {
```
mook-as, line 89: This seems like it should live somewhere shared, because we'll need it for lots of other things too? 

I'm okay with leaving that until we actually have a use for it though.

cmd/rdd/limavm.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create client: %w", err)
+	}
+	return c, nil
+}
+
+// findLimaVM searches for a LimaVM with the given name across all namespaces.
+func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha1.LimaVM, error) {
+	vmList := &limav1alpha1.LimaVMList{}
+	if err := c.List(ctx, vmList); err != nil {
```
mook-as: Should be able to use a field selector, right? I'm guessing it will involve [`client.MatchingFields`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.22.2/pkg/client#MatchingFields).

Getting _everything_ and then filtering on the client side seems silly.

cmd/rdd/limavm.go
```diff
@@ -0,0 +1,196 @@
...
+		return nil, fmt.Errorf("failed to list LimaVMs: %w", err)
+	}
+
+	var found *limav1alpha1.LimaVM
+	for i := range vmList.Items {
+		if vmList.Items[i].Name == name {
+			if found != nil {
+				return nil, fmt.Errorf("multiple LimaVMs found with name %q in different namespaces", name)
```
mook-as: It would be good to identify the namespaces involved.

mook-as (changes requested):

cmd/rdd/limavm.go
jandubois: There is no need because the VM name is unique. Only the `create` command needs to know in which namespace to create the instance.

cmd/rdd/limavm.go
jandubois: You cannot filter across namespaces with the Kubernetes API, at least I don't know how.

I think the MatchingFields() thing requires a cache, or cache index, so you would still have to fetch the whole list, just at a different place. But I don't really know how that works either.

cmd/rdd/limavm.go
mook-as: Well, currently you're fetching all VMs (regardless of name), so you do need to check it, right? You can have a `vm-1` and a `vm-2` (per the BATS test).

mook-as (dismissed):
We can do the field selector thing later, I guess

mook-as (approved):
LGTM.

I guess we don't need to bother creating a second VM in the BATS test to make sure the filtering is working correctly (that is, the server didn't return VMs with the wrong name)?

jandubois:
> I guess we don't need to bother creating a second VM in the BATS test to make sure the filtering is working correctly (that is, the server didn't return VMs with the wrong name)?

I don't think so. Maybe I add it the next time I change the tests,

I also tested it manually:

```console
❯ rdd ctl get limavms
NAME   RUNNING
vm1    false
vm2    true

❯ rdd ctl get limavms --field-selector metadata.name=vm2
NAME   RUNNING
vm2    true
```

PR #44, opened 2025-10-20
Update lima API and command specs

jandubois:
I already have a local branch implementing the `templateRef` changes and corresponding `rdd limavm` updates.

The annotations and the `params` stuff does not exist at all yet.

### DIFF
--- a/docs/design/api_lima.md
+++ b/docs/design/api_lima.md
@@ -35,19 +35,19 @@ status:
 
 - **metadata.annotations**: Can be used to request "actions" from the reconciler, like resetting (deleting and recreating with the same settings), or restarting the instance. The `status.conditions` can be used to store the state machine information.
 
-- **running**: Set to `true` when the instance should be running, set to `false` when it should be stopped.
+- **spec.running**: Set to `true` when the instance should be running, set to `false` when it should be stopped.
 
-- **templateRef**: Specifies the `lima.yaml` template for the machine. The template must pass Lima validation or the `LimaVM` creation will fail. 
+- **spec.templateRef**: Specifies the `lima.yaml` template for the machine. The template must pass Lima validation or the `LimaVM` creation will fail. 
 
-    Initially the only way to specify a template is via a ConfigMap that will store the "fully embedded" template under the `template` key. It cannot reference external base templates or scripts. This will change eventually when `templateRef.name` can also be set to a URL. The default for `templateRef.namespace` is the same as `metadata.namespace`.
+    Initially the only way to specify a template is via a ConfigMap that will store the "fully embedded" template under the `template` key. It cannot reference external base templates or scripts. This will change eventually when `spec.templateRef.name` can also be set to a URL. The default for `spec.templateRef.namespace` is the same as `metadata.namespace`.
 
-    The `LimaVM` controller will create a new ConfigMap (in `metadata.namespace`) with the `metadata.name` and a `"-template"` suffix to store a copy of the validated template. This name is stored under `status.templateConfigMap`. The original `templateRef` source is never accessed again after this, and can be modified or deleted without affecting the `LimaVM` resource. The `spec.templateRef` is immutable after creation and only serves as documentation.
+    The `LimaVM` controller will create a new ConfigMap (in `metadata.namespace`) with the `metadata.name` and a `"-template"` suffix to store a copy of the validated template. This name is stored under `status.templateConfigMap`. The original `spec.templateRef` source is never accessed again after this, and can be modified or deleted without affecting the `LimaVM` resource. The `spec.templateRef` is immutable after creation and only serves as documentation.
 
-    The `templateConfigMap` can be modified, but must pass Lima validation for the update to succeed. If `spec.running` is `true` and the template has changed, then the instance will be restarted. The `templateConfigMap` cannot be deleted, except by deleting the `LimaVM` resource itself, which will clean up owned resources automatically.
+    The `status.templateConfigMap` can be modified, but must pass Lima validation for the update to succeed. If `spec.running` is `true` and the template has changed, then the instance will be restarted. The `status.templateConfigMap` cannot be deleted, except by deleting the `LimaVM` resource itself, which will clean up owned resources automatically.
 
-- **params**: Override `params` settings in the template. These values will be merged with the template before validation, and when creating/updating the `lima.yaml` file of the actual instance on disk.
+- **spec.params**: Override `spec.params` settings in the template. These values will be merged with the template before validation, and when creating/updating the `lima.yaml` file of the actual instance on disk.
 
-    If the template provisioning scripts are properly parameterized, then the instance settings can be modified by just updating `spec.params`, which is simpler than modifying the `template` inside the ConfigMap.
+    If the template provisioning scripts are properly parameterized, then the instance settings can be modified by just updating `spec.params`, which is simpler than modifying the `template` inside the ConfigMap. If `spec.running` is `true` then changing `spec.params` will restart the instance.
 
 ## LimaDisk
 
--- a/docs/design/api_lima.md
+++ b/docs/design/api_lima.md
@@ -18,14 +18,36 @@ kind: LimaVM
 metadata:
   name: alpine
   namespace: default
+  annotations:
+    lima.rancherdesktop.io/resetRequested: "2025-10-12T08:00:00Z"
+    lima.rancherdesktop.io/restartRequested: "2025-10-12T10:30:00Z"
 spec:
-  template: alpine
-  status: running
+  running: true
+  templateRef:
+    name: alpine
+    namespace: lima-templates
+  params:
+    DOCKER_ROOTFUL: "true"
 status:
-  status: stopped
+  templateConfigMap: alpine-template
+  conditions: []
 ```
 
-The `template` specifies a ConfigMap that contains the `lima.yaml` template for the machine. It must be "fully embedded"; it cannot reference external base templates or scripts.
+- **metadata.annotations**: Can be used to request "actions" from the reconciler, like resetting (deleting and recreating with the same settings), or restarting the instance. The `status.conditions` can be used to store the state machine information.
+
+- **running**: Set to `true` when the instance should be running, set to `false` when it should be stopped.
+
+- **templateRef**: Specifies the `lima.yaml` template for the machine. The template must pass Lima validation or the `LimaVM` creation will fail. 
+
+    Initially the only way to specify a template is via a ConfigMap that will store the "fully embedded" template under the `template` key. It cannot reference external base templates or scripts. This will change eventually when `templateRef.name` can also be set to a URL. The default for `templateRef.namespace` is the same as `metadata.namespace`.
+
+    The `LimaVM` controller will create a new ConfigMap (in `metadata.namespace`) with the `metadata.name` and a `"-template"` suffix to store a copy of the validated template. This name is stored under `status.templateConfigMap`. The original `templateRef` source is never accessed again after this, and can be modified or deleted without affecting the `LimaVM` resource. The `spec.templateRef` is immutable after creation and only serves as documentation.
+
+    The `templateConfigMap` can be modified, but must pass Lima validation for the update to succeed. If `spec.running` is `true` and the template has changed, then the instance will be restarted. The `templateConfigMap` cannot be deleted, except by deleting the `LimaVM` resource itself, which will clean up owned resources automatically.
+
+- **params**: Override `params` settings in the template. These values will be merged with the template before validation, and when creating/updating the `lima.yaml` file of the actual instance on disk.
+
+    If the template provisioning scripts are properly parameterized, then the instance settings can be modified by just updating `spec.params`, which is simpler than modifying the `template` inside the ConfigMap.
 
 ## LimaDisk
 
--- a/docs/design/cmd_lima.md
+++ b/docs/design/cmd_lima.md
@@ -1,13 +1,45 @@
-# Lima Commands
+# LimaVM Commands
 
-## `rdd lima`
+## `rdd limavm`
 
 This command is similar to `limactl`, but uses `rdd` to create/start/stop/delete VM instances. It is a convenience command to work with other VMs and not needed to operate the Rancher Desktop app.
 
-```bash
-rdd lima create --name my-vm --cpus 4 lima.yaml
-rdd lima start my-vm
-rdd lima stop my-vm
-rdd lima delete my-vm
-rdd lima shell my-vm
-```
+### `rdd limavm create NAME TEMPLATE --namespace NAMESPACE`
+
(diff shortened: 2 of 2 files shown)


docs/design/api_lima.md
```diff
@@ -18,14 +18,36 @@ kind: LimaVM
...
 status:
-  status: stopped
+  templateConfigMap: alpine-template
+  conditions: []
 ```
 
-The `template` specifies a ConfigMap that contains the `lima.yaml` template for the machine. It must be "fully embedded"; it cannot reference external base templates or scripts.
+- **metadata.annotations**: Can be used to request "actions" from the reconciler, like resetting (deleting and recreating with the same settings), or restarting the instance. The `status.conditions` can be used to store the state machine information.
+
+- **running**: Set to `true` when the instance should be running, set to `false` when it should be stopped.
```
mook-as:
```suggestion
- **spec.running**: Set to `true` when the instance should be running, set to `false` when it should be stopped.
```
Same with `spec.templateRef`.

docs/design/api_lima.md
```diff
@@ -18,14 +18,36 @@ kind: LimaVM
...
+
+- **templateRef**: Specifies the `lima.yaml` template for the machine. The template must pass Lima validation or the `LimaVM` creation will fail. 
+
+    Initially the only way to specify a template is via a ConfigMap that will store the "fully embedded" template under the `template` key. It cannot reference external base templates or scripts. This will change eventually when `templateRef.name` can also be set to a URL. The default for `templateRef.namespace` is the same as `metadata.namespace`.
+
+    The `LimaVM` controller will create a new ConfigMap (in `metadata.namespace`) with the `metadata.name` and a `"-template"` suffix to store a copy of the validated template. This name is stored under `status.templateConfigMap`. The original `templateRef` source is never accessed again after this, and can be modified or deleted without affecting the `LimaVM` resource. The `spec.templateRef` is immutable after creation and only serves as documentation.
```
mook-as: Presumably it's not actually _immutable_ (that is, the user can still edit it, it would just be ignored)? Or is there a webhook to reject edits here?

docs/design/api_lima.md
```diff
@@ -18,14 +18,36 @@ kind: LimaVM
...
+
+    The `templateConfigMap` can be modified, but must pass Lima validation for the update to succeed. If `spec.running` is `true` and the template has changed, then the instance will be restarted. The `templateConfigMap` cannot be deleted, except by deleting the `LimaVM` resource itself, which will clean up owned resources automatically.
```
mook-as: So there's a webhook guarding that? (Would it make sense to make it a different type than `configmap` in that case, at least to make it easier to understand?)

docs/design/api_lima.md
```diff
@@ -18,14 +18,36 @@ kind: LimaVM
...
+
+- **params**: Override `params` settings in the template. These values will be merged with the template before validation, and when creating/updating the `lima.yaml` file of the actual instance on disk.
```
mook-as: Presumably, updating `spec.params` will restart the VM too?

docs/design/cmd_lima.md
```diff
@@ -1,13 +1,45 @@
-# Lima Commands
+# LimaVM Commands
 
-## `rdd lima`
+## `rdd limavm`
 
 This command is similar to `limactl`, but uses `rdd` to create/start/stop/delete VM instances. It is a convenience command to work with other VMs and not needed to operate the Rancher Desktop app.
```
mook-as: Maybe document that this command is normally hidden?

docs/design/cmd_lima.md
```diff
@@ -1,13 +1,45 @@
...
-rdd lima delete my-vm
-rdd lima shell my-vm
-```
+### `rdd limavm create NAME TEMPLATE --namespace NAMESPACE`
+
+Create a new `LimaVM` instance `NAME` in `NAMESPACE` (or `default`) using `TEMPLATE`.
+
+`TEMPLATE` should be the name of a ConfigMap inside the `NAMESPACE` where the resource should be created. It will be used as the `spec.templateRef.name` in the `LimaVM` resource. Referencing a ConfigMap in a different namespace is currently not supported and requires the use of `rdd ctl apply` instead.
+
+`TEMPLATE` will be treated as a local filename if it contains a `/`.  In that case the `create` command will create a ConfigMap in `NAMESPACE` called `NAME`. It will store the "fully embedded" template from the file inside that ConfigMap and use it as the `spec.templateRef`. If a ConfigMap of this name already exists, then the `create` command will fail. If `LimaVM` creation succeeds then ownership of this ConfigMap is transferred to the `LimaVM` resource, and it will be deleted when the `LimaVM` instance is deleted. But when the command fails, then the ConfigMap is deleted immediately.
```
mook-as: I assume _fully embedded_ is a term from Lima? I'm not familiar with it.

mook-as: If we interpret anything with `/` as a local path, than `github:foo/bar` or `http://example.com/file.yaml` wouldn't work?

mook-as: Does the config map get deleted immediately (when something fails with the VM creation) even if the config map already existed before the command was run? That seems like it may be a bit annoying.  But that's what this text currently says.

docs/design/cmd_lima.md
```diff
@@ -1,13 +1,45 @@
...
+
+### `rdd limavm start NAME`
+
+Set `spec.running` of the specified instance to `true`. There is no `--namespace` option because `LimaVM` names are globally unique (within the controlplane).
```
mook-as:
```suggestion
Set `spec.running` of the specified instance to `true`. There is no `--namespace` option because `LimaVM` names are globally unique (within the control plane).
```
Unless _controlplane_ as a single word (as opposed to two) has some specific usage?

docs/design/api_lima.md
```diff
@@ -18,14 +18,36 @@ kind: LimaVM
...
+  params:
+    DOCKER_ROOTFUL: "true"
 status:
-  status: stopped
+  templateConfigMap: alpine-template
+  conditions: []
 ```
 
-The `template` specifies a ConfigMap that contains the `lima.yaml` template for the machine. It must be "fully embedded"; it cannot reference external base templates or scripts.
+- **metadata.annotations**: Can be used to request "actions" from the reconciler, like resetting (deleting and recreating with the same settings), or restarting the instance. The `status.conditions` can be used to store the state machine information.
```
mook-as, line 36: Presumably the annotations don't get removed after the action is completed? I can see either being useful…

docs/design/cmd_lima.md
```diff
@@ -1,13 +1,45 @@
...
+
+If `spec.running` is true, then the instance will be restarted if the `params` have changed.
+
+### `rdd limavm reset NAME`
+
+Set `lima.rancherdesktop.io/resetRequested` annotation to the current timestamp. This tells the reconciler to delete the existing instance (stopping it, if necessary), and then recreate it using the same `status.templateConfigMap` and `spec.params` template.
+
+### `rdd limavm restart NAME`
+
+Set `lima.rancherdesktop.io/restartRequested` annotation to the current timestamp. This tells the reconciler to stop the instance (if it is running), and then to start it again.
```
mook-as: Does this set `spec.running` to `true`?

mook-as (changes requested):
Reviews of markdown would be a lot easier if the lines are wrapped; that way the comment can go with the correct context.  It all looks the same rendered anyway.

I think it would be easier to read to always refer to the keys by their full names, e.g. `spec.params`; otherwise things can get confusing.

I'm not sure I'm sold on using annotations to control reset / restart yet; I may need to think on that some more, since I don't see any obviously better solutions either.

docs/design/api_lima.md
jandubois: Yes, the webhook will reject it. Because it was trivial to add to the type spec; otherwise it doesn't really matter because it will never be used again anyways.

```
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="templateRef is immutable"
```

docs/design/api_lima.md
jandubois: Yes, we need an admission controller anyways to make sure the template validates before we accept the change.

I'm not sure why we should re-invent something like configmap. Everything is controlled via labels, or owner references. I'm trying to have the base controller do all the cleanup of owned resources automatically (because we have no garbage collection).

docs/design/cmd_lima.md
jandubois: It's not even that; it is just what I use to reference the output of merging base templates and external scripts into a final template without external references.

Eventually the `LimaVM` admission controller will have to learn how to do this, but for now we will require the input template to be "fully embedded`.

Alpine template:

```console
❯ limactl tmpl copy template:alpine -
minimumLimaVersion: 1.1.0
base:
  - template://_images/alpine
  - template://_default/mounts
# The built-in containerd installer does not support Alpine currently.
# Use a provisioning script to install containerd, buildkit, and nerdctl.
containerd:
  system: false
  user: false
```

Fully embedded (base templates are resolved):

```console
❯ limactl tmpl copy --embed-all template:alpine -
minimumLimaVersion: 1.1.0
# The built-in containerd installer does not support Alpine currently.
# Use a provisioning script to install containerd, buildkit, and nerdctl.
containerd:
  system: false
  user: false
images:
  - location: "https://dl-cdn.alpinelinux.org/alpine/v3.22/releases/cloud/nocloud_alpine-3.22.2-x86_64-uefi-cloudinit-r0.qcow2"
    arch: "x86_64"
    digest: "sha512:6dcb2ec1ac3ff160f1a2609fbd7cb8417df6c1c937c259238090d5bc494778924c0eb360c926d90bfbe7d0176692ae00b049f4a06d5070618a3b087a47c96532"
  - location: "https://dl-cdn.alpinelinux.org/alpine/v3.22/releases/cloud/nocloud_alpine-3.22.2-aarch64-uefi-cloudinit-r0.qcow2"
    arch: "aarch64"
    digest: "sha512:481ad2db4fe024dd5d2dddd2abf2f722b738173c9ff6b76425ee877706bd49e26b2877ccae7b389aebae1d5e8df0384cc3c8355fc903378997340ce6c1b85faf"
mounts:
  - location: "~"
# # /tmp/lima is no longer mounted by default since Lima v2.0.
# - location: "{{.GlobalTempDir}}/lima"
#  mountPoint: /tmp/lima
#  writable: true
```

jandubois: It will only be deleted if it has been created just now from a file. The whole paragraph talks about `TEMPLATE` including a `/` and all following sentences in this paragraph are about that situation. How can I make it clearer?

docs/design/api_lima.md
jandubois: They could be, once the action is complete. But they don't have to be because they are set to a timestamp, to prevent them from re-triggering.

I was planning to just leave them around, so you can see some breadcrumbs of the requested action, but it likely will be visible from `status.conditions` as well. Can't tell you the details until I've written the code. 😄

docs/design/cmd_lima.md
jandubois: I think so. If `spec.running` is `false`, then `restart` is the same as `start`.

docs/design/cmd_lima.md
jandubois: Ok, will change it to "local path or URL". I guess we need to say `/` or `:` because not every URL will contain a `/`.

jandubois:
> Reviews of markdown would be a lot easier if the lines are wrapped; that way the comment can go with the correct context. It all looks the same rendered anyway.

All the files in `docs/design/` follow the markdown rule of one line per paragraph; I can't remember why. 🤷 
 
> I think it would be easier to read to always refer to the keys by their full names, e.g. `spec.params`; otherwise things can get confusing.

I think I've changed it everywhere (in the 2 files I edited).
 
> I'm not sure I'm sold on using annotations to control reset / restart yet; I may need to think on that some more, since I don't see any obviously better solutions either.

I first investigated creating custom API endpoints for it, but that seemed very complicated, and is not normally done. It also doesn't help much because you still need a state machine most of the time because you cannot block until the action is done. E.g. you need to wait for the instance to shut down before you can delete it. So using annotations or labels as "triggers" seems a lot easier.

I have not even tried to write code for it though, so I may still change my mind. Let me know if you have a better idea!

docs/design/cmd_lima.md
mook-as: It would be good to define that in this document, then.  I'd probably call it _fully expanded_ or _fully interpolated_, but I don't mind whichever term as long as it's defined.

mook-as: I think just _… then the created ConfigMap is deleted immediately_ is enough; it just wasn't clear if the whole paragraph was about the auto-created config maps, or if it's just expanding on the details of the config map (which may included ones the user created).

jandubois: I just changed it to

> But when the command fails, then any ConfigMap created from a file or URL is deleted immediately.

Does that make it clear?

mook-as (approved):
On `rdd limavm reset`, if `spec.running` is true, the VM would probably need to be restarted.

But I'm okay with that just being implied by the doc for now.

docs/design/cmd_lima.md
mook-as: Thinking about this more (but this doesn't need to go in), and looking up the [Kubernetes object name requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), I think this boils down to:
- If the `TEMPLATE` string is a valid Kubernetes object name, then it will be treated as the name of a config map;
- Otherwise, it will be treated as either a file name or URL, depending on ??? (Presumably, if matches `^[a-z0-9-]` or something along those lines).

jandubois:
> On `rdd limavm reset`, if `spec.running` is true, the VM would probably need to be restarted.

At least to me that is obvious from the way how reconciliation works. If you don't want to start it, you can set `spec.running` to `false`. Can be in the same `patch` request if you do it via `rdd ctl`, or you have to stop it before the `reset` if you are using `rdd limavm`.

I only want `rdd limavm` capable enough to simplify BATS tests and help with manual testing/debugging. It doesn't need to support every possible scenario. There is always `rdd ctl apply`. 😄

docs/design/cmd_lima.md
mook-as: Saw this while looking at other things about config maps:
From the [configmap docs](https://kubernetes.io/docs/concepts/configuration/configmap/#configmap-object):

> The name of a ConfigMap must be a valid [DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

So effectively, if `TEMPLATE` does not match that, we should assume it's some external thing (somehow), and will be imported and deleted later.

jandubois: Way ahead of you already:

```go
	// Check if template is a valid ConfigMap name (DNS-1123 subdomain)
	validationErrs := validation.IsDNS1123Subdomain(template)
	if len(validationErrs) > 0 {
		// Template is not a valid ConfigMap name, treat it as a file path or URL
		templateContent, err := os.ReadFile(template)
```

PR #48, opened 2025-10-21
CI: Lint: Use platform runners

mook-as:
We tried to use Linux runners to run lint (overriding GOOS) to save costs, but that turned out to mess with the cache and took even longer to run. Use the default strategy of using the matching host OS in the hopes of getting faster runs.


jandubois (approved):
Thanks, LGTM

PR #51, opened 2025-10-24
Implement templateRef property for LimaVM resource

jandubois:
* Uses mutating admission controller to copy the template into an owned ConfigMap.
* Validates the template on CREATE and UPDATE, protects against DELETE.

General improvements to the base controller handling:
* adds support for mutating admission controllers (Defaulters)
* support for multiple webhook managers per controller
  (e.g. `LimaVM` has an additional admission controller for `ConfigMap` with label)
* webhook name derived from GVK
* plural resources name determined from REST mapper
* refactored finalizers
  * get rid of `FinalizerHelper`; just use functions
  * unified finalizer name for "deletion protection"
  * resources can own multiple other resources
  * ResourceNamespace interface for cluster-scoped resources like App, Demo
* bump BATS submodules to get newer helper functions like `assert_stderr_*`
* new `logrus` BATS helpers `assert_error` etc
* new BATS `try —wait-fail` option
* updated BATS submodules


bats/helpers/logs.bash
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-FileCopyrightText: The Rancher Desktop Authors
+# SPDX-FileCopyrightText: The Lima Authors
+
+# Format the string the way strconv.Quote() would do.
+# If the input ends with an ellipsis then no closing quote will be added (and the … will be removed).
```
mook-as, line 7: Not sure I like having non-ASCII characters, but I guess that's okay…

bats/helpers/utils.bash
```diff
@@ -291,7 +300,13 @@ try() {
     local count=0
     while true; do
         run --separate-stderr "$@"
```
mook-as: Should this be using `run_e`?

bats/lib/bats-assert
mook-as, line 1: It would be good if the commit message said _why_ we're bumping BATS. Otherwise it needs to be a separate PR.

bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -0,0 +1,138 @@
...
+    local template=$3
+
+    local msg
+    printf -v msg 'LimaVM "%s" created in namespace "%s" with template ConfigMap "%s"' "$name" "$namespace" "$template"
+    assert_info "$msg"
+}
+
+@test "lima create/start/stop/delete with ConfigMap template" {
+    # Create a template ConfigMap first
+    rdd ctl create configmap "test-template" -n "$LIMA_TEST_NS" --from-literal=template='{"vmType":"vz"}'
```
mook-as:
```suggestion
    rdd ctl create configmap "test-template" --namespace "$LIMA_TEST_NS" --from-literal=template='{"vmType":"vz"}'
```

I know you said before that you preferred _some_ arguments to be short, but I think `-n` is too easy to miss…

bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -0,0 +1,138 @@
...
+
+# assert_running verifies the "test-vm" running state, must be "true" or "false"
+assert_running() {
+    local state=$1
+    run -0 rdd ctl get limavm "test-vm" -n "$LIMA_TEST_NS" -o jsonpath='{.spec.running}'
+    assert_output "$state"
+}
+
+# assert_created verifies the $name VM has been created in $namespace from $template
+# The `rdd limavm create` command must have been run with `run --separate-stderr`.
```
mook-as: The comment should say something about this testing `$output`

bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -0,0 +1,138 @@
+load '../../helpers/load'
+
+# LimaVM CLI tests - tests the rdd limavm subcommands for managing LimaVM resources.
+
+LIMA_TEST_NS="lima-test-ns"
+
+local_setup_file() {
+    setup_rdd_control_plane "lima"
+    rdd ctl create namespace "$LIMA_TEST_NS"
```
mook-as, line 9: What do you feel about just using `local_setup` to create the namespace (and `local_teardown` to delete all of it)?

bats/tests/33-lima-controllers/limavm-cli.bats
mook-as: `vmType: vz` will start failing in the future once we start validating right? Because that doesn't exist on Linux & Windows? Any reason this isn't just an empty object, `{}`?

bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -0,0 +1,138 @@
...
+    # Verify the ConfigMap was cleaned up (should not exist in $LIMA_TEST_NS)
+    run -1 rdd ctl get configmap "duplicate-vm" -n "$LIMA_TEST_NS"
+    assert_output --partial "not found"
+
+    # Clean up the first LimaVM
+    rdd limavm delete "duplicate-vm"
+}
+
+@test "lima create fails with non-existent file" {
+    run -1 rdd limavm create "test-vm" "/nonexistent/template.yaml" -n "$LIMA_TEST_NS"
```
mook-as: Probably still better to use `${BATS_TEST_TMPDIR}` to ensure the path gets converted on Windows (to a file that doesn't exist).

bats/tests/33-lima-controllers/limavm-lifecycle.bats
```diff
@@ -0,0 +1,188 @@
...
+    run -1 create_limavm "test-vm-missing" "nonexistent-template"
+    assert_output --partial "not found"
+}
+
+@test "create LimaVM with ConfigMap missing template key fails" {
+    rdd ctl create configmap "invalid-template" -n "$NAMESPACE" --from-literal="foo=bar"
+
+    run -1 create_limavm "test-vm-invalid" "invalid-template"
+    # Mutating webhook tries to create template ConfigMap, ConfigMap webhook validates and rejects
+    assert_output --partial '"template" data cannot be empty'
```
mook-as, line 172: Should also check that the configmap isn't modified?

cmd/rdd/limavm.go
```diff
@@ -38,11 +45,16 @@ func newLimaCommand() *cobra.Command {
 func newLimaCreateCommand() *cobra.Command {
```
mook-as:
```suggestion
func newLimaVMCreateCommand() *cobra.Command {
```

cmd/rdd/limavm.go
```diff
@@ -38,11 +45,16 @@ func newLimaCommand() *cobra.Command {
...
 	var namespace string
 	command := &cobra.Command{
-		Use:   "create NAME",
+		Use:   "create NAME TEMPLATE",
 		Short: "Create a new LimaVM resource",
-		Args:  cobra.ExactArgs(1),
+		Long: `Create a new LimaVM resource with the specified template.
+
+TEMPLATE can be:
+- A ConfigMap name (if it's a valid Kubernetes DNS-1123 subdomain) in the namespace specified by -n
```
mook-as:
```suggestion
- A ConfigMap name (if it's a valid Kubernetes DNS-1123 subdomain) in the namespace specified by --namespace
```
In the help text, it's probably best to be explicit

cmd/rdd/limavm.go
```diff
@@ -38,11 +45,16 @@ func newLimaCommand() *cobra.Command {
 func newLimaCreateCommand() *cobra.Command {
 	var namespace string
 	command := &cobra.Command{
-		Use:   "create NAME",
+		Use:   "create NAME TEMPLATE",
 		Short: "Create a new LimaVM resource",
-		Args:  cobra.ExactArgs(1),
+		Long: `Create a new LimaVM resource with the specified template.
+
+TEMPLATE can be:
```
mook-as:
```suggestion
TEMPLATE can be one of:
```
We _probably_ don't want both :)

cmd/rdd/limavm.go
```diff
@@ -38,11 +45,16 @@ func newLimaCommand() *cobra.Command {
...
+- A ConfigMap name (if it's a valid Kubernetes DNS-1123 subdomain) in the namespace specified by -n
+- A file path or URL (if it's not a valid ConfigMap name) - creates a ConfigMap with the VM name`,
+		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return limaCreateAction(cmd.Context(), args[0], namespace)
+			return limaCreateAction(cmd.Context(), args[0], args[1], namespace)
 		},
 	}
 	command.Flags().StringVarP(&namespace, "namespace", "n", "default", "Namespace for the LimaVM resource")
```
mook-as: `k8s.io/apimachinery/pkg/apis/meta/v1#NamespaceDefault` is probably the correct constant here, but I guess a hard-coded string is probably fine. We're already importing `metav1` though.

cmd/rdd/limavm.go
```diff
@@ -117,27 +132,106 @@ func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha
...
+func limaCreateAction(ctx context.Context, name, template, namespace string) error {
 	c, err := getKubeClient(ctx)
 	if err != nil {
 		return err
 	}
 
+	var templateConfigMapName string
+	var createdConfigMap *corev1.ConfigMap // Track if we created a ConfigMap
+
+	// Check if template is a valid ConfigMap name (DNS-1123 subdomain)
```
mook-as: Might be good to link to the docs that say that's the correct validation for configmap names.

cmd/rdd/limavm.go
```diff
@@ -117,27 +132,106 @@ func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha
...
+	validationErrs := validation.IsDNS1123Subdomain(template)
+	if len(validationErrs) == 0 {
```
mook-as: It might be easier to read if this part is split into a helper function, so it can have an early return if the name is valid and we won't need to read it from remote places.

cmd/rdd/limavm.go
```diff
@@ -117,27 +132,106 @@ func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha
...
 		Spec: limav1alpha1.LimaVMSpec{
+			TemplateRef: limav1alpha1.TemplateReference{
+				Name: templateConfigMapName,
+			},
 			Running: &running,
 		},
 	}
 
 	if err := c.Create(ctx, limaVM); err != nil {
+		// If we created a ConfigMap and LimaVM creation failed, clean it up
```
mook-as: Maybe this should be done in a `defer` (and we just clear `createdConfigMap` before returning successfully), so if we fail to set ownership info the config map is deleted.

cmd/rdd/limavm.go
```diff
@@ -117,27 +132,106 @@ func findLimaVM(ctx context.Context, c client.Client, name string) (*limav1alpha
...
+		configMapToUpdate := &corev1.ConfigMap{}
+		if err := c.Get(ctx, types.NamespacedName{Name: createdConfigMap.Name, Namespace: createdConfigMap.Namespace}, configMapToUpdate); err != nil {
+			logrus.Warnf("Failed to fetch ConfigMap for owner reference update: %v", err)
+			return nil // Don't fail the whole operation
+		}
+
+		// Set owner reference using controller-runtime helper
+		if err := ctrl.SetControllerReference(limaVM, configMapToUpdate, c.Scheme()); err != nil {
+			logrus.Warnf("Failed to set owner reference on ConfigMap: %v", err)
+			return nil // Don't fail the whole operation
```
mook-as: As noted above, if this fails, we should probably delete the config map.

Why shouldn't this fail the whole operation? The resources are in an inconsistent state now.

pkg/apis/app/v1alpha1/demo_types.go
```diff
@@ -70,3 +76,13 @@ type DemoList struct {
...
 }
+
+// GetResourceNamespace implements the base.ResourceNamespace interface.
+// It returns the namespace where this cluster-scoped Demo resource creates
+// and manages its owned namespaced resources.
+func (d *Demo) GetResourceNamespace() string {
+	if d.Spec.Namespace != "" {
+		return d.Spec.Namespace
+	}
+	return "default"
```
mook-as:
```suggestion
	return metav1.NamespaceDefault
```

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
-func (h *FinalizerHelper) CleanupOwnedResources(ctx context.Context, owner client.Object, opts CleanupOptions) error {
-	if !h.Enabled {
-		return nil
-	}
-
+// DeleteOwnedResources finds and deletes all resources owned by the given object.
+// This is RDD's replacement for Kubernetes garbage collection.
+//
+// The function always removes any finalizers from owned resources before deletion
+// to ensure complete cleanup when the owner is being deleted.
```
mook-as: That seems… odd? The description doesn't explain what happens to grandchild resources owned by the thing you're removing finalizers from.

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
+//
+// The function attempts to delete all owned resources, collecting errors along the way.
+// If any deletions fail, it returns a combined error, but continues trying to delete
+// remaining resources to make maximum progress per reconciliation.
+//
+// Namespace detection:
```
mook-as: The wording sounds like it's about deleting whole namespaces, not just the things in them that are owned by this resource?

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
-	Namespace string
-	// LabelSelector to find resources to clean up
+	// ResourceLists is a slice of empty list objects used to query different resource types.
+	// Each list object (e.g., &corev1.ConfigMapList{}) serves as a template that gets
+	// populated by Client.List(). Example: []client.ObjectList{&corev1.ConfigMapList{}, &corev1.SecretList{}}
+	ResourceLists []client.ObjectList
+	// LabelSelector to find resources to clean up (optional).
+	// If provided, uses efficient label-based filtering.
+	// If empty, lists all resources in namespace and filters by owner UID.
 	LabelSelector map[string]string
```
mook-as: Does it make sense to just use `client.MatchingLabels` here instead of `map[string]string`?

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
-			"resourceType", fmt.Sprintf("%T", obj),
-			"resourceNamespace", obj.GetNamespace(),
-			"resourceName", obj.GetName())
-
-		if err := h.Client.Delete(ctx, obj); err != nil && client.IgnoreNotFound(err) != nil {
-			return fmt.Errorf("failed to delete resource during cleanup: %w", err)
+	for _, resourceList := range opts.ResourceLists {
+		// List() mutates resourceList by populating its Items field
+		if err := c.List(ctx, resourceList, listOpts...); err != nil {
+			errs = append(errs, fmt.Errorf("failed to list resources for cleanup: %w", err))
```
mook-as: Shouldn't the error message describe the GVK of the resourceList type?

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
 		}
-	}
 
-	return nil
-}
+		// Use Kubernetes standard library to iterate over list items
+		_ = meta.EachListItem(resourceList, func(runtimeObj runtime.Object) error {
+			obj, ok := runtimeObj.(client.Object)
+			if !ok {
+				errs = append(errs, fmt.Errorf("item is not a client.Object: %T", runtimeObj))
```
mook-as: Would `runtimeObj.GetObjectKind().GroupVersionKind()` be more readable?

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
+// If any deletions fail, it returns a combined error, but continues trying to delete
+// remaining resources to make maximum progress per reconciliation.
+//
+// Namespace detection:
+// - For namespaced owners: uses owner.GetNamespace().
+// - For cluster-scoped owners: uses ResourceNamespace interface if implemented.
+// - Returns error if namespace cannot be determined.
+//
+// Discovery methods:
+// - If opts.LabelSelector is provided, uses efficient label-based filtering.
```
mook-as: But this actually still does the filtering by owner UID.  Documenting that is useful, because that means the caller can use a best-effort label selector, instead of thinking they have to be exact.

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
-	}
+			logger.Info("Deleting owned resource",
+				"resourceType", fmt.Sprintf("%T", obj),
+				"resourceNamespace", obj.GetNamespace(),
+				"resourceName", obj.GetName())
 
-	v = v.Elem()
-	if v.Kind() != reflect.Struct {
-		return nil, errors.New("list must be a struct")
+			if err := c.Delete(ctx, obj); err != nil && client.IgnoreNotFound(err) != nil {
```
mook-as, line 144: I don't think you need to `nil`-test `err` before calling `IgnoreNotFound`?

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
+			return nil // Always continue to next item
+		})
 	}
 
-	// Look for Items field
-	itemsField := v.FieldByName("Items")
-	if !itemsField.IsValid() {
-		return nil, errors.New("list does not have Items field")
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to delete some owned resources (%d errors): %v", len(errs), errs)
```
mook-as: Wouldn't using `errors.Join()` be better?

pkg/controllers/base/webhook.go
```diff
@@ -24,77 +24,150 @@ import (
...
+	}
+	return req.DryRun != nil && *req.DryRun
+}
+
+// GenerateWebhookPath generates the webhook path that controller-runtime uses
+// when registering a webhook for a given GroupVersionKind and webhook type.
+// This follows controller-runtime's conventions:
+// - Validating: /validate-{group}-{version}-{kind}
+// - Mutating: /mutate-{group}-{version}-{kind}
+// For core resources (empty group), the path becomes: /validate--{version}-{kind} or /mutate--{version}-{kind}.
```
mook-as: The code doesn't seem to handle empty group.

This should have a unit test, because it's easy.

pkg/controllers/base/finalizer.go
```diff
@@ -7,195 +7,168 @@ package base
...
-	return func(resource client.Object, owner client.Object) bool {
-		for _, ownerRef := range resource.GetOwnerReferences() {
-			if ownerRef.UID == owner.GetUID() &&
-				ownerRef.Kind == expectedKind &&
-				ownerRef.APIVersion == expectedAPIVersion {
-				return true
+			// Remove all finalizers before deletion to ensure cleanup completes
+			if len(obj.GetFinalizers()) > 0 {
+				logger.Info("Removing finalizers from owned resource",
+					"resourceType", fmt.Sprintf("%T", obj),
```
mook-as: Here, too, `obj.GetObjectKind().GroupVersionKind()` (or at least `gvk.GroupVersion()+gvk.Kind`)

pkg/controllers/base/webhook.go
```diff
@@ -24,77 +24,150 @@ import (
...
-	// Register webhook validation with controller-runtime
-	builder := ctrl.NewWebhookManagedBy(mgr).For(obj).WithValidator(validator)
+// SetupWebhookForResource sets up webhook(s) for a resource type.
+// It registers the webhook(s) with controller-runtime and creates WebhookManager(s) for later setup.
+//
+// The config must specify either a validator, a defaulter, or both. The function will create the
+// appropriate webhook manager(s) based on which are provided.
+func SetupWebhookForResource(mgr ctrl.Manager, obj client.Object, config WebhookConfig) ([]*WebhookManager, error) {
+	if config.Validator == nil && config.Defaulter == nil {
+		return nil, errors.New("config must specify either Validator or Defaulter (or both)")
```
mook-as: Optional suggestion:

```suggestion
		return nil, errors.New("config must specify at least one of either Validator or Defaulter")
```

pkg/controllers/base/webhook.go
```diff
@@ -24,77 +24,150 @@ import (
...
+	}
+
+	// Extract GVK from the resource object
+	gvks, _, err := mgr.GetScheme().ObjectKinds(obj)
```
mook-as, line 138: Is there any reason to not use `gvk := obj.GetObjectKind().GroupVersionKind()` instead?

pkg/controllers/base/webhook.go
```diff
@@ -137,73 +235,113 @@ func (wm *WebhookManager) createWebhookConfiguration() error {
...
+	case MutatingWebhook:
+		return wm.createMutatingWebhook(ctx, c, webhookURL, clientConfig, rules, failurePolicy, sideEffects)
+	case ValidatingWebhook:
+		return wm.createValidatingWebhook(ctx, c, webhookURL, clientConfig, rules, failurePolicy, sideEffects)
+	default:
+		return fmt.Errorf("invalid webhook type: %s", wm.webhookType)
+	}
+}
+
+func (wm *WebhookManager) createValidatingWebhook(ctx context.Context, c client.Client, webhookURL string, clientConfig admissionregistrationv1.WebhookClientConfig, rules []admissionregistrationv1.RuleWithOperations, failurePolicy admissionregistrationv1.FailurePolicyType, sideEffects admissionregistrationv1.SideEffectClass) error {
```
mook-as, line 287: Hmm, not worth it yet, but I wish we could reduce the duplication here.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -7,52 +7,138 @@ package controllers
...
+// Note: templateRef is never accessed after initial creation (handled by mutating webhook).
+// It exists only as documentation of the template's origin.
+func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// Fetch the LimaVM instance
+	var limaVM v1alpha1.LimaVM
+	if err := r.Get(ctx, req.NamespacedName, &limaVM); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("LimaVM resource not found. Ignoring since object must be deleted")
```
mook-as, line 61: Do we need to try to delete the created configmap here? Not for the normal flow, but in case there was a race or something.

I guess we can't because we have no idea what the name / namespace would be?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -7,52 +7,138 @@ package controllers
...
+		}
+		if err := r.Update(ctx, templateConfigMap); err != nil {
+			logger.Error(err, "Failed to update template ConfigMap")
+			return ctrl.Result{}, err
+		}
+		logger.Info("Set owner reference on template ConfigMap", "ConfigMap.Name", configMapName)
+	}
+
+	// Update status.templateConfigMap to mark initialization as complete
+	limaVM.Status.TemplateConfigMap = limaVM.GetTemplateConfigMapName()
```
mook-as: Shouldn't this just be `configMapName` from above? If it returns a different value, it should still be the config map we just updated.

pkg/controllers/lima/limavm/controller.go
```diff
@@ -139,41 +168,159 @@ func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
...
+
+		// Don't create ConfigMap in dry-run mode
+		return nil
+	}
+
+	// Fetch template data from templateRef
+	templateData, err := d.fetchTemplateRefData(ctx, limavm)
+	if err != nil {
+		return fmt.Errorf("failed to fetch template data: %w", err)
+	}
```
mook-as: This part is basically the same, dry run or not; does it make sense to just move this before the dry run check?

pkg/controllers/lima/limavm/controller.go
```diff
@@ -139,41 +168,159 @@ func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
...
+	configMap := &corev1.ConfigMap{}
+	if err := d.Client.Get(ctx, configMapKey, configMap); err != nil {
+		return "", fmt.Errorf("failed to get templateRef ConfigMap %q in namespace %q: %w", configMapKey.Name, configMapKey.Namespace, err)
+	}
+	return configMap.Data[templateKey], nil
+}
+
+// ConfigMapValidator validates ConfigMap resources that are template ConfigMaps for LimaVM resources
+// (identified by label selector controllers.TemplateConfigMapLabel set to "true").
```
mook-as: Maybe leave a reference to `setupWebhookWithRuntimeConfig` here?

pkg/controllers/lima/limavm/controller.go
```diff
@@ -139,41 +168,159 @@ func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	// Fetch template data from templateRef
+	templateData, err := d.fetchTemplateRefData(ctx, limavm)
+	if err != nil {
+		return fmt.Errorf("failed to fetch template data: %w", err)
+	}
+
+	// Create template ConfigMap with label
+	templateConfigMap := &corev1.ConfigMap{
```
mook-as, line 227: I wonder if we can just call `ConfigMapValidator#ValidateCreate` here instead of the special code path for dry run.

pkg/controllers/lima/limavm/crd.yaml
mook-as, line 1: We should really have an auto-generated marker for these files… Different PR, though.

pkg/service/cmd/service.go
```diff
@@ -19,6 +19,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/sirupsen/logrus"
```
mook-as, line 22: We already use `klog` in this file; would it make more sense to stick with that? Otherwise we should at least have this in a separate commit and describe why we switch to logrus in the commit message.

pkg/service/controllers/manager.go
```diff
@@ -201,18 +201,23 @@ func (scm *SharedControllerManager) Start(ctx context.Context) error {
...
+		// Start webhook setup immediately if this controller has webhooks
 		if webhookController, ok := registration.(base.WebhookController); ok {
-			if manager := webhookController.GetWebhookManager(); manager != nil {
+			managers := webhookController.GetWebhookManagers()
+			for i, manager := range managers {
+				if manager == nil {
+					continue
+				}
 				name := registration.GetName()
+				webhookIndex := i
```
mook-as: As of golang 1.22, the loop variables are independent and we no longer need to do this.

pkg/service/readiness/ready.go
```diff
@@ -159,49 +149,36 @@ func waitForCRDsEstablished(ctx context.Context, config *rest.Config, controller
...
-	if logging {
-		logger.Info("Checking CRD establishment", "expectedCRDs", expectedCRDNames)
-	}
+	logger.Info("Checking CRD establishment", "expectedCRDs", expectedCRDNames)
+
+	crdClient := apiextensionsClient.ApiextensionsV1().CustomResourceDefinitions()
 
 	expectedCRDs := sets.New(expectedCRDNames...)
 	establishedCRDs := sets.New[string]()
-	var mu sync.Mutex
```
mook-as, line 174: We no longer need the mutex?  I guess nothing actually run in parallel so it wasn't actually _locking_ anything…

pkg/service/readiness/ready.go
```diff
@@ -309,76 +276,86 @@ func isCRDEstablished(crd *apiextensionsv1.CustomResourceDefinition) bool {
...
-			if logging {
-				logger.Info("Controller has webhook configuration", "controller", controllerName, "webhook", webhookName)
+		webhookController, ok := controller.(base.WebhookController)
+		if !ok {
+			continue
+		}
+		controllerName := controller.GetName()
+		webhookManagers := webhookController.GetWebhookManagers()
+		for _, mgr := range webhookManagers {
+			if mgr == nil {
```
mook-as, line 306: Not sure when this would be true, but I guess it's good to be defensive?

pkg/service/readiness/ready.go
```diff
@@ -309,76 +276,86 @@ func isCRDEstablished(crd *apiextensionsv1.CustomResourceDefinition) bool {
...
-			logger.Info("No webhook configurations expected, skipping webhook readiness check")
-		}
+	if len(expectedWebhooks) == 0 {
+		logger.Info("No webhook configurations expected, skipping webhook readiness check")
 		return nil
 	}
 
-	if logging {
-		logger.Info("Waiting for webhook configurations to be created", "webhooks", expectedWebhookNames)
+	var webhookNames []string
```
mook-as: Should this just be the `Set` directly?

pkg/service/readiness/ready.go
```diff
@@ -387,46 +364,39 @@ func waitForWebhookConfigurations(ctx context.Context, config *rest.Config, cont
...
 		case <-ticker.C:
 			// Check for any missing webhook configurations
-			for _, webhookName := range expectedWebhookNames {
-				mu.Lock()
-				alreadyFound := foundWebhooks.Has(webhookName)
-				mu.Unlock()
+			for _, webhook := range expectedWebhooks {
+				alreadyFound := foundWebhooks.Has(webhook.name)
 
 				if alreadyFound {
```
mook-as:
```suggestion
				if foundWebhooks.Has(webhook.name) {
```
We don't use that anywhere else, and inlining it is short and clear.

pkg/service/readiness/ready.go
```diff
@@ -387,46 +364,39 @@ func waitForWebhookConfigurations(ctx context.Context, config *rest.Config, cont
...
-					mu.Unlock()
+					foundWebhooks.Insert(webhook.name)
+					allFound := foundWebhooks.Equal(expectedWebhookSet)
 
-					if logging {
-						logger.Info("Webhook configuration became available", "webhook", webhookName)
-					}
+					logger.Info("Webhook configuration became available", "webhook", webhook.name, "type", webhook.webhookType)
 
 					if allFound {
```
mook-as:
```suggestion
					logger.Info("Webhook configuration became available", "webhook", webhook.name, "type", webhook.webhookType)

					if foundWebhooks.IsSuperset(expectedWebhookSet) {
```

Same here: we don't need the intermediate.  Also, it should be clearer that what we actually care about is that all of the expected webhooks are found.

mook-as (changes requested):
This PR is a little big, it might be worthwhile splitting it up into multiple PRs if possible.  As-is, getting all of the comments resolved may take a while.

I still don't understand why we're removing finalizers in the manual-GC, since that would stop the GC chain.  I think that might just mean I don't have the algorithm correctly in my head though.

Given that we can do server-side dry runs, I guess `app` doesn't need the `dryRun` spec field, because that would just be done in the admission controller?

bats/helpers/logs.bash
jandubois: What is the point of using UTF-8 encoding when all we use is 7-bit ASCII? 😄

bats/helpers/utils.bash
jandubois: Yeah, why not. I primarily created `run_e` to avoid the visual clutter in actual tests, but best to use it consistently.

bats/lib/bats-assert
jandubois: I'll add it to the commit message. I thought we were at the latest versions of the helper libraries, but we were just at the same level as the rancherdesktop repo.

I needed the `assert_stderr_*` helpers for the `logs.bash` helpers.

bats/tests/33-lima-controllers/limavm-cli.bats
jandubois: I'll try to remember this for my next PR that implements the namespace cleanup.

Doing it right now would just hang indefinitely because the namespace deleter is part of kubelet, so we cannot run it.

pkg/controllers/lima/limavm/crd.yaml
mook-as: Created #52 for this one.

bats/tests/33-lima-controllers/limavm-lifecycle.bats
jandubois: I don't know why; we never modify the source template.

cmd/rdd/limavm.go
jandubois: Thanks!

cmd/rdd/limavm.go
jandubois: +1 (I've also removed `or URL` from the next line because that is not implemented yet)

cmd/rdd/limavm.go
jandubois: +1

cmd/rdd/limavm.go
jandubois: Thanks, I like `metav1.NamespaceDefault` even though I think it will never be different from `"default"`.

cmd/rdd/limavm.go
jandubois: +1

cmd/rdd/limavm.go
jandubois: Not sure if I understand this correctly. I've moved the ConfigMap creation to a separate function, which simplifies this quite a bit. I hope that's what you are asking for.

cmd/rdd/limavm.go
jandubois: I thought failing to take ownership should not fail the command, but it doesn't really matter; I don't think it will ever happen.

pkg/apis/app/v1alpha1/demo_types.go
jandubois: +1

pkg/controllers/base/finalizer.go
jandubois: +1

pkg/controllers/base/finalizer.go
jandubois: An object will not be deleted if it has a finalizer; it will just get a `DeletionTimestamp`. It is then up to the reconciler of the object to do cleanup, and remove the finalizer from itself. That allows Kubernetes to remove the resource from the data store.

When the LimaVM takes ownership of a ConfigMap, it sets the owner reference, but also our shared cleanup finalizer. The ConfigMap reconciler doesn't know anything about the finalizer, so will not remove it, make it essentially unremovable.

When the LimaVM gets deleted, it will remove the finalizer from all owned resources, so they are no longer protected from deletion.

Note that the LimaVM may take ownership of the `spec.templateRef` ConfigMap, but doesn't set the finalizer on it. Which means it will not prevent it from being deleted, but if it is still around when the LimaVM itself is deleted, it will delete that ConfigMap as well.

However the `status.templateConfigMap` is both owned and has the finalizer applied. This is the source for the `lima.yaml` of the instance. We don't allow it to be deleted, and we will not allow it to be set to a template that fails validation.

So: owner references are just for tracking owner ship, finalizers are used to protect against deletion.

We have to remove the finalizers to allow the resources to be deleted normally. If they in turn own other resources themselves, then they are responsible for it. They may have their own finalizers applied.

And after writing this wall of text, I think I just understand your question: we should only remove our own finalizer, not all of them, so this is a bug. Will change it. I'll leave the text above for your amusement.

pkg/controllers/base/finalizer.go
jandubois: I've reworded it; let me know if it needs further changes.

pkg/controllers/base/finalizer.go
jandubois: Yes, it costs nothing to filter by UID because we already have the data, so I thought I'd leave it as a safety feature.

Reworded this too; lmk if it is ok now.

pkg/controllers/base/finalizer.go
jandubois: +1

pkg/controllers/base/finalizer.go
jandubois: Yes, it would.

pkg/controllers/base/finalizer.go
jandubois: I don't know, the documentation is confusing, and the implementation not immediately obvious:

```
// IgnoreNotFound returns nil on NotFound errors.
// All other values that are not NotFound errors or nil are returned unmodified.
```

So this sound like `nil` is being modified. Seems prudent to keep the check.

pkg/controllers/base/finalizer.go
jandubois: +1

pkg/controllers/base/webhook.go
jandubois: How does it not handle `group` being the empty string?

I'm kind of tempted to remove the comment; I only have it because I was initially confused (before we even started copying the implementation from controller-runtime. Now it is not really a special-case anymore; it falls naturally out of the implementation.

pkg/controllers/base/webhook.go
jandubois: Because the ObjectKind has not yet been filled in if `obj` has just been constructed in memory:

```go
managers, err := base.SetupWebhookForResource(mgr, &v1alpha1.LimaVM{}, mutatingConfig)
```

The existing code looks it up in the scheme, which works even if the object does not actually exist yet. I've added a comment.

pkg/controllers/base/webhook.go
jandubois: Yes, I was trying, but it isn't really possible because these are separate types. Using reflection and a switch statement does not really make it better.

pkg/controllers/lima/limavm/controller.go
jandubois: yes, it does

pkg/controllers/lima/limavm/controller.go
jandubois: I don't understand what you mean. The `validateTemplateData()` function is already the refactored version that is shared between dry-run and actual creation.

The main difference is that in dry-run is is being called from the LimaVM Defaulter, but in the regular case it is called from the ConfigMap validator. So we would have to cook up a `ConfigMapValidator` to call the function indirectly. Seems more complicated than what we have.

I probably don't understand what you are asking for.

pkg/controllers/lima/limavm/controller.go
jandubois: What kind of reference? Should I explain that this is a validating admission webhook that is being set up with `setupWebhookWithRuntimeConfig`? Isn't that kind of obvious because it is the only usage of it?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: I don't know; this is just defensive programming in case the reconciler ever gets called on a resource that is simultaneously deleted. If this happens at all, then it should only happen after we have removed the finalizer on the ConfigMap already, so it can be deleted normally. I think this is fine as-is.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yes, you are correct

pkg/controllers/lima/limavm/crd.yaml
jandubois: Good idea

pkg/service/cmd/service.go
jandubois: The idea is that only `rdd service serve` (the background daemon) will use `klog` because Kubernetes already uses `klog`. The CLI code uses `logrus`.

I've split it into a separate commit.

pkg/service/controllers/manager.go
jandubois: I know, but deep down I still worry... 😄

pkg/service/readiness/ready.go
jandubois: Yes, I realized there is no concurrency.

pkg/service/readiness/ready.go
jandubois: It was true before, but I went through several rounds of refactoring, and now it should never be true anymore. I decided to keep it because it doesn't hurt.

pkg/service/readiness/ready.go
jandubois: Yes

pkg/service/readiness/ready.go
jandubois: +1; didn't look closely when I removed the mutex that it could be simplified further.

pkg/service/readiness/ready.go
jandubois: +1

jandubois (commented):
I believe I've addressed all comments, except a few where I asked for clarification.

I've split out the logrus changes into a separate commit, and I've rebased the PR on latest `main`.

bats/tests/33-lima-controllers/limavm-lifecycle.bats
mook-as: Well, to ensure that we never modify the source template, of course.

pkg/controllers/base/finalizer.go
mook-as: Ah, so removing the finalizer on a thing allows it to be deleted, which then causes any other things it in turn owns to be deleted (or at least marked with a deletion timestamp until its finalizers are also gone). In that case, with the fix to just our finalizers I think that could make sense.

pkg/controllers/base/webhook.go
mook-as: Oh, I misread, and missed that it actually says `--` in that case.  So it's actually fine (but yes, the comment is extraneous and in fact lead to my confusion…)

pkg/controllers/lima/limavm/controller.go
mook-as: I was thinking that making a temporary `ConfigMapValidator` would more of the same code is being reused (in particular, we can use check the config map we're actually about to create instead of just part of it).  Currently if we screw up the config map definition (i.e. lines 225 through 237 in this version) we wouldn't notice.

pkg/controllers/lima/limavm/controller.go
mook-as: Yeah, I was initially confused because the immediately following code _didn't_ reference `TemplateConfigMapLabel`.  I think instead of mentioning that (because it's really an implementation detail), just say that `setupWebhookWithRuntimeConfig` applies this validator only to the relevant config maps is fine.

pkg/controllers/base/finalizer.go
```diff
@@ -9,193 +9,158 @@ import (
...
-	if err := h.Client.List(ctx, opts.ResourceType, listOpts...); err != nil {
-		return fmt.Errorf("failed to list resources for cleanup: %w", err)
-	}
+	var namespace string
+	if ns := owner.GetNamespace(); ns != "" {
+		namespace = ns
+	} else if rn, ok := owner.(ResourceNamespace); ok {
+		namespace = rn.GetResourceNamespace()
+	} else {
+		return fmt.Errorf("cannot determine namespace for cleanup: owner %q is cluster-scoped but does not implement ResourceNamespace interface", owner.GetName())
```
mook-as: This doesn't actually say what kind the owner is, does it? Maybe we can ignore that for now…

pkg/controllers/base/finalizer.go
```diff
@@ -9,193 +9,158 @@ import (
...
-		controllerutil.AddFinalizer(obj, h.FinalizerName)
-		if err := h.Client.Update(ctx, obj); err != nil {
+func EnsureFinalizer(ctx context.Context, c client.Client, obj client.Object) (bool, error) {
+	if !controllerutil.ContainsFinalizer(obj, FinalizerName) {
+		controllerutil.AddFinalizer(obj, FinalizerName)
+		if err := c.Update(ctx, obj); err != nil {
 			return false, fmt.Errorf("failed to add finalizer: %w", err)
 		}
 		return true, nil
 	}
```
mook-as:
```suggestion
	if controllerutil.AddFinalizer(obj, FinalizerName) {
		if err := c.Update(ctx, obj); err != nil {
			return false, fmt.Errorf("failed to add finalizer: %w", err)
		}
		return true, nil
	}
```
`AddFinalizer` checks if it already exists, and returns `false` in that case.

pkg/controllers/base/finalizer.go
```diff
@@ -9,193 +9,158 @@ import (
...
-		return nil
-	}
-
-	controllerutil.RemoveFinalizer(obj, h.FinalizerName)
-	if err := h.Client.Update(ctx, obj); err != nil {
+func RemoveFinalizer(ctx context.Context, c client.Client, obj client.Object) error {
+	controllerutil.RemoveFinalizer(obj, FinalizerName)
+	if err := c.Update(ctx, obj); err != nil {
 		return fmt.Errorf("failed to remove finalizer: %w", err)
 	}
```
mook-as:
```suggestion
	if controllerutil.RemoveFinalizer(obj, FinalizerName) {
		if err := c.Update(ctx, obj); err != nil {
			return fmt.Errorf("failed to remove finalizer: %w", err)
		}
	}
```
No need to update the object if we didn't change the set of finalizers.

pkg/controllers/lima/limavm/controller.go
```diff
@@ -139,41 +165,152 @@ func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
...
 	limavm, ok := obj.(*v1alpha1.LimaVM)
 	if !ok {
-		return nil, fmt.Errorf("expected a LimaVM object but got %T", obj)
+		return fmt.Errorf("expected a LimaVM object but got %T", obj)
+	}
+
+	// Add finalizer if not present
+	if !base.HasFinalizer(limavm) {
+		limavm.Finalizers = append(limavm.Finalizers, base.FinalizerName)
+	}
```
mook-as: Why not use `controllerutil.AddFinalizer(limavm, base.FinalizerName)` here?

pkg/controllers/base/finalizer.go
```diff
@@ -9,193 +9,158 @@ import (
...
+// The function attempts to delete all owned resources, collecting errors along the way.
+// If any deletions fail, it returns a combined error, but continues trying to delete
+// remaining resources to make maximum progress per reconciliation.
+//
+// DeleteOwnedResources only looks for owned resources in the owner.GetNamespace() namespace.
+// For cluster-scoped objects it uses the ResourceNamespace interface to determine the namespace.
+//
+// Use opts.LabelSelector to efficiently filter the set of deletion candidates, if possible.
+// Resources not actually owned by the owner will not be touched.
+func DeleteOwnedResources(ctx context.Context, c client.Client, owner client.Object, opts CleanupOptions) error {
```
mook-as, line 87: I think I still don't understand why we're removing child finalizers here, instead of just asking for a deleting (causing a `deletionTimestamp` to be set, and then handling the finalizers in the webhooks).  After all, we seem to be _adding_ the finalizers in the webhooks for the objects, not their owners…

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -7,52 +7,138 @@ package controllers
...
+	configMapName := limaVM.GetTemplateConfigMapName()
+	templateConfigMap := &corev1.ConfigMap{}
+	if err := r.Get(ctx, types.NamespacedName{Name: configMapName, Namespace: limaVM.Namespace}, templateConfigMap); err != nil {
+		logger.Error(err, "Failed to get template ConfigMap for owner reference setup")
+		return ctrl.Result{}, err
+	}
+
+	// Set owner reference if not already set.
+	// Note: ConfigMap finalizer is already set during creation by the mutating webhook.
+	if len(templateConfigMap.OwnerReferences) == 0 {
```
mook-as: Should this be something closer to `isOwnedByUID`? The difference being, what should happen if somebody else also set an owner ref on it?

pkg/service/cmd/service.go
mook-as: Please expand the commit message to say _why_ we're using logrus instead of klog, instead of just saying we should use logrus in CLIs.

mook-as (changes requested):

bats/tests/33-lima-controllers/limavm-cli.bats
jandubois: Change it to mention `$stderr`.

bats/tests/33-lima-controllers/limavm-cli.bats
jandubois: Changed them all

bats/tests/33-lima-controllers/limavm-cli.bats
jandubois: Sure, updated.

bats/tests/33-lima-controllers/limavm-lifecycle.bats
jandubois: Yeah, but there are a million things we don't do, so I don't get the point.

And testing that we don't modify the template when it is empty doesn't prove we don't modify it under other conditions.

pkg/controllers/base/finalizer.go
jandubois: Thanks, changed (also in RemoveFinalizer).

pkg/controllers/base/finalizer.go
jandubois: +1

pkg/controllers/base/finalizer.go
jandubois: Yes, this is wrong. I've already fixed this in my followup PR that implements the namespace cleanup; it also refactors `DeletedOwnedResources`.

Can we leave the fixing to the next PR?

Cherry-picking these parts out would be a lot of pointless work, and applying all the refactoring to this PR would add a lot of code to an already oversized PR.

pkg/controllers/base/finalizer.go
jandubois: I added the gvk type.

pkg/controllers/base/webhook.go
jandubois: Removed the comment

pkg/controllers/lima/limavm/controller.go
jandubois: +1

pkg/controllers/lima/limavm/controller.go
jandubois: I still don't get it. You want to use the `ConfigMapValidator` on the `spec.templateRef` ConfigMap during dry-run? I don't understand how that would make the code simpler, and I don't understand what additional checks it would do.

I guess we could check that no ConfigMap with the `templateConfigMap` name exists, because that would fail if it was not a dry-run. I've added a check for this to the dry-run branch (and a corresponding BATS test).

The `status.templateConfigMap` will be validated by our admission controller, which uses the `ConfigMapValidator` for implementation because it is triggered by the `TemplateConfigMapLabel`. So I don't understand what you mean we would not noticed if we screw up the config map definition here; it would get rejected by the Validator.

pkg/controllers/lima/limavm/controller.go
jandubois: I've reworded it to

```
// It is only invoked for ConfigMaps that have the TemplateConfigMapLabel set to "true".
```

But it still feels like the same to me; I don't understand the difference.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Sounds unlikely, but checking with `IsOwnedByUID` is more correct, so changing.

pkg/service/cmd/service.go
jandubois: I've added an example to show why klog doesn't work for CLI output.

jandubois (commented):
Sorry, I missed some questions in the first round. I believe I have addressed all of them now.

There are 2 issues I would like to address in a follow-up PR:

* deleting namespaces because it requires the new namespace controller
* fixing the finalizer handling because I've refactored all this code in the namespace controller PR

pkg/controllers/lima/limavm/controller.go
mook-as: I want to delay the difference between dry run or not, so that the only change is `d.Client.Create` vs `validator.ValidateCreate`, so that the dry run is more obviously as close to the real thing as possible.

pkg/controllers/base/finalizer.go
jandubois: Wait, it looks like I merged the code back in here already. I'm totally confused now

mook-as (approved):
Eh, we can make any further changes in follow-up PRs; this one is so large that it's unwieldy to review.

PR #52, opened 2025-10-27
Generate-CRDs: add header

mook-as:
This marks the CRD files as generated.


scripts/generate-crds.sh
```diff
@@ -16,9 +16,16 @@ TMPDIR=$(mktemp -d)
 readonly TMPDIR
 trap 'rm -rf "$TMPDIR"' EXIT
 
+cat > "$TMPDIR/header.txt" <<EOF
+# Generated by generate-crds.sh; DO NOT EDIT.
```
jandubois: I think this line should be below the license template. Once `ltag` supports YAML files it will require the comments to be on top unless we implement also a different document class. Right now shell scripts can have a #! line, and Go files a comment with compiler flags, but that's it.

jandubois (approved):
Thanks, LGTM

PR #53, opened 2025-10-28
CI: Fix getting write token

mook-as:
The setup to fetch the token is set up slightly differently; instead of using a GitHub application, the token is stored directly in vault. We also switch to running the workflow on `pull_request` instead of `pull_request_target` and just let the action fail if it's from a cross-fork pull request, since the main goal of this workflow is dependabot pull requests which come from the same repository.

Sample run from #43 (which will be recreated after this merges): https://github.com/rancher-sandbox/rancher-desktop-daemon/actions/runs/18884953018/job/53897848262

### DIFF
--- a/.github/actions/get-token/action.yaml
+++ b/.github/actions/get-token/action.yaml
@@ -12,7 +12,7 @@ inputs:
 outputs:
   token:
     description: The GitHub token retrieved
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #54, opened 2025-10-28
CI: Disable setup-go caching in places

mook-as:
The setup-go action only uses the contents of `go.sum` to calculate the cache key; this means that different workflows that download different modules will end up with cache key collisions.  This means that the fastest job to complete will end up being the one populating the cache; this typically means the jobs that end up downloading the smallest amount of modules will win (frequently, this means the license tag checking job).  This is not ideal, because it means the cache is not being utilized effectively as we will download more kubernetes modules more often.

To resolve this, manually mark the jobs that do not pull in the majority of modules with `cache: false` to reserve the cache for the heavy weights.

A second commit also pins `shfmt` to a specific version using `go tool`, so that we can be sure we are running that version if we need to re-run it in the future.


jandubois:
~Wouldn't it be better to use a different cache-key for each OS and AARCH instead of disabling the cache? As long as the total cache size is below 10GB this should work even better.~

```yaml
cache-dependency-path: go.sum
cache-key: go-${{ matrix.os }}-${{ matrix.goarch }}
```

It seems like `cache-key` does not actually exist and is just AI hallucination. 😞 

But it should be possible to do our own caching with `actions/cache` instead.

mook-as:
Yeah, I checked https://github.com/actions/setup-go/blob/main/action.yml for things that may be useful, and didn't spot anything.  https://github.com/actions/cache/blob/main/action.yml does have `key`, but we're not using that directly so it wouldn't work.

If you think it's worth the maintenance hassle to use `actions/cache` directly, I can look into that, but at first glance that doesn't seem to be worth 15s of downloading the non-Kubernetes modules.

jandubois:
> The setup-go action only uses the contents of `go.sum` to calculate the cache key; this means that different workflows that download different modules will end up with cache key collisions.

Looking at the action source code itself I see

```ts
const primaryKey = `setup-go-${platform}-${arch}-${linuxVersion}go-${versionSpec}-${fileHash}`;
```

Which looks like we should have a different cache key for each platform.

jandubois:
I'm reluctant to post another AI response, but this is what Copilot suggested as a manual alternative (with caching disabled in `actions/setup-go`):

```yaml
- name: Cache Go build
  uses: actions/cache@v4
  with:
    path: |
      ~/.cache/go-build
      ~/go/pkg/mod
    key: go-${{ matrix.os }}-${{ matrix.goarch }}-${{ hashFiles('**/go.sum') }}
    restore-keys: |
      go-${{ matrix.os }}-${{ matrix.goarch }}-
```

If it doesn't seem immediately applicable, then we can leave this for later (or don't bother at all, if the benefit is only small).

> doesn't seem to be worth 15s of downloading the non-Kubernetes modules.

Agreed. I thought we would still be looking at 5min+ downloads on the paths where we now disable caching.

mook-as:
We do have a different hash per platform, but Linux has multiple setup-go uses:
- Kubernetes (that is, the actual RDD code from BATS and golangci-lint)
- ltags
- shfmt
- The script to fix up go.mod for different Kubernetes components
- The script for spelling (to figure out the go package names)

Each would produce a different combination of packages in the go module cache; the first is ~800MB, the others are up to ~10MB.

jandubois (approved):
Thanks, LGTM

Let's see how this works on a daily basis, maybe it is all we need.

jandubois:
> Each would produce a different combination of packages in the go module cache

I thought you could just run something like `go get ./...` after the setup to always get the full set of dependencies. But let's ignore this for now; maybe what we have now is good enough.

PR #55, opened 2025-10-30
BATS: Turn on optional lints for shellcheck

mook-as:
This enables optional shellcheck checks, to further constrain how we can write the BATS tests to get better uniformity across our scripts.  It is recommended to review commit-by-commit, as one of the commits is done with automated fixes (using `shellcheck … --format=diff` and applying the output).


mook-as:
Rebased for spelling fixes.

jandubois (approved):
Thanks, LGTM

PR #56, opened 2025-10-30
Spelling: Fix excludes file

mook-as:
It turns out check-spelling takes empty lines in the excludes file to mean
all files should be excluded, so we ended up never checking anything.

Issue filed upstream:
https://github.com/check-spelling/check-spelling/issues/96


jandubois (dismissed):
Check Spelling is complaining about missing newline at EOF for `.github/actions/spelling.exclude.txt` itself. 😆 

Also the change from `checkout: true` to `false` is not explained in the PR summary.

mook-as:
Added the new line at EOF back after testing that it does _not_ exclude all files.

Also updated the commit message about `checkout` :p

jandubois (approved):
Thanks, LGTM

PR #57, opened 2025-11-01
limavm: Support `rdd limavm create --dry-run`

mook-as:
This adds the option to use (server-side) dry-run to create lima vms with full validation.  It actually still creates the config map as needed, as that is required to pass validation; the config map is deleted after if it is a dry-run.

We may want to consider embedding the configuration into the LimaVM object instead in the future, so that this would not be an issue.

Hopefully this explains what I was requesting in #51. (At the time I wasn't thinking about server-side dry-run, but this is better.)

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-cli.bats
+++ b/bats/tests/33-lima-controllers/limavm-cli.bats
@@ -29,6 +29,21 @@ assert_created() {
     assert_info "${msg}"
 }
 
+@test "lima create with missing ConfigMap template" {
+    # Create VM with ConfigMap template
+    run_e -1 rdd limavm create "test-missing-vm" "test-missing-template" --namespace "$LIMA_TEST_NS"
+    assert_stderr_line --partial 'denied the request'
+    assert_stderr_line --partial 'ConfigMap \"test-missing-template\" not found'
+
+    # Verify the ConfigMap was not created, or at least deleted after.
+    run -1 rdd ctl get configmap "test-missing-template" --namespace "$LIMA_TEST_NS" --output=name
+    assert_output --partial NotFound
+
+    # Verify the VM was not created
+    run -0 rdd ctl get limavm --namespace "$LIMA_TEST_NS" --output=name
+    refute_output # There should be no output at all
+}
+
 @test "lima create/start/stop/delete with ConfigMap template" {
     # Create a template ConfigMap first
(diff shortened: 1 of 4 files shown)


bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -29,6 +29,21 @@ assert_created() {
     assert_info "${msg}"
 }
 
+@test "lima create with missing ConfigMap template" {
```
jandubois, line 32: I feel like this test is not testing any CLI feature, so should probably be merged with the existing (rudimentary) test in `limavm-lifecycle.bats`:

```
@test "create LimaVM with nonexistent template ConfigMap fails" {
```

jandubois (dismissed):
Thanks, LGTM

I didn't get this idea from your previous feedback, but passing on the `DryRun` option to the ConfigMap admission controller is a great idea; so much simpler.

I struggle a bit with how to organize the LimaVM tests: What is the difference between the admission/cli/lifecycle tests.

In my mind I separated them like this:

* admission is about validating the template itself
* cli tests the functionality of the `rdd limavm` commands, not the controller
* lifecycle tests everything about managing ownership, finalizers, labels, but not validating the content of the template

I guess `admission` is a poor choice (didn't have the mutating `Defaulter` at first), and should be renamed to `validation`.

Anyways, we can always refactor it later; still a lot of work to go into the `LimaVM` controller...

jandubois:
> This adds the option to use (server-side) dry-run to create lima vms with full validation. It actually still creates the config map as needed, as that is required to pass validation; the config map is deleted after if it is a dry-run.

You mean the `spec.templateRef` ConfigMap is created by the `rdd limavm create` command? I think that is fine; this is just a debugging/testing tool, for the most part.

Eventually we will support URLs as an alternative to ConfigMap names, and we can probably support `file://` URLs too, as long as they use absolute pathnames.

> We may want to consider embedding the configuration into the LimaVM object instead in the future, so that this would not be an issue.

We can discuss this, but I feel this is not important, and I wanted to keep the template our of the LimaVM resource because it can be quite large, so `rdd ctl get limavms -o yaml` would be unwieldy. Any settings that you expect to change regularly should be made available via `params`, that will be directly on the `LimaVM`.

It is the `App` controller that should have all the settings directly on the resource itself, to make it easy to validate changes in the "Preferences" dialog as the user makes changes.

jandubois (approved):

PR #59, opened 2025-11-04
Refactor DeleteOwnedResources to check all resource types

jandubois:
Enumerating resource types and resources inside a namspace is just a local sqlite database query, and iterating over the results should be just as quick as filtering stuff "server-side", which is the same process on the same computer.

Not requiring the caller to pass in a list of resource types is more maintenance friendly and the additional time required should be small, given that resources are not constantly being deleted.

Label selectors are used only for setting up reconcilers and general watches, and deletion is only controlled by owner references now.

### DIFF
--- a/pkg/controllers/base/finalizer.go
+++ b/pkg/controllers/base/finalizer.go
@@ -9,10 +9,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
-	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/runtime"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/discovery"
+	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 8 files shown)


pkg/controllers/base/finalizer.go
```diff
@@ -97,58 +92,126 @@ func DeleteOwnedResources(ctx context.Context, c client.Client, owner client.Obj
...
 
-		// Use Kubernetes standard library to iterate over list items
-		_ = meta.EachListItem(resourceList, func(runtimeObj runtime.Object) error {
-			obj, ok := runtimeObj.(client.Object)
-			if !ok {
-				errs = append(errs, fmt.Errorf("item is not a client.Object: %s", runtimeObj.GetObjectKind().GroupVersionKind()))
-				return nil // Continue to next item
+		// Process each resource
+		for i := range list.Items {
+			item := &list.Items[i]
```
mook-as: I guess we're not using `for i, item := …` so that `item` can be a pointer?

pkg/controllers/base/finalizer.go
```diff
@@ -97,58 +92,126 @@ func DeleteOwnedResources(ctx context.Context, c client.Client, owner client.Obj
...
-				if err := c.Update(ctx, obj); err != nil {
-					errs = append(errs, fmt.Errorf("failed to remove finalizers from %s %s/%s: %w", gvk, obj.GetNamespace(), obj.GetName(), err))
-					return nil // Continue to next item
+			patch := client.MergeFrom(item.DeepCopy())
+			if controllerutil.RemoveFinalizer(item, FinalizerName) {
+				// Use Patch instead of Update for PartialObjectMetadata
+				if err := c.Patch(ctx, item, patch); err != nil {
+					if apierrors.IsNotFound(err) {
+						logger.V(1).Info("Resource already deleted during finalizer removal",
+							"resourceName", item.GetName())
```
mook-as: Do we need gvk / namespace here? It may be useful to set up a logger beforehand with the gvk set ahead of time, shared with the `Failed to read resources` line.

mook-as (dismissed):
I think there's enough code here to do more focused testing, even with the BATS testing for the limavm/configmap pair.  But I guess that's sort of optional given that it _is_ tested… it just feels inadequate somehow.

pkg/controllers/base/finalizer.go
jandubois: Yeah, but using this form looks like we care about modifying the elements of `list.Items`, which we don't, so maybe using `&item` when needed is clearer.

pkg/controllers/base/finalizer.go
jandubois: Good idea.

jandubois (commented):
I've amended my existing commit with the feedback changes, as they are quite small.

jandubois:
> I think there's enough code here to do more focused testing, even with the BATS testing for the limavm/configmap pair. But I guess that's sort of optional given that it _is_ tested… it just feels inadequate somehow.

Can you create a new issue for this with more details. I don't want to be blocked by it right now, but if you have ideas how we can better test this, let's make sure we don't lose them.

mook-as (approved):

PR #60, opened 2025-11-04
Global names should not start with the package name

jandubois:
limavm.LimaVMDefaulter → limavm.Defaulter
notary.NotaryValidator → notary.Validator

### DIFF
--- a/pkg/controllers/lima/limavm/controller.go
+++ b/pkg/controllers/lima/limavm/controller.go
@@ -118,7 +118,7 @@ func (c *Controller) setupWebhookWithRuntimeConfig(mgr ctrl.Manager) error {
 			admissionregistrationv1.Create, // Only on CREATE - UPDATE doesn't need to create ConfigMap again
 		},
 		SideEffects: &sideEffectsNoneOnDryRun, // Creates ConfigMap normally, but not during dry-run
-		Defaulter:   &LimaVMDefaulter{Client: mgr.GetClient()},
+		Defaulter:   &defaulter{Client: mgr.GetClient()},
 	}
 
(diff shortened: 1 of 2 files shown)


mook-as (dismissed):
Looks fine, but I think we should un-export those symbols too (i.e. `defaulter`, `validator`); as this PR shows, nobody references those things outside of the single file they're defined in.

jandubois:
> I think we should un-export those symbols too

Agreed, I've updated the PR

pkg/controllers/lima/limavm/controller.go
```diff
@@ -166,16 +166,16 @@ func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
 	return c.setupWebhookWithRuntimeConfig(mgr)
 }
 
-// LimaVMDefaulter handles LimaVM mutation via webhook.
+// defaulter handles LimaVM mutation via webhook.
```
mook-as, line 169: Should that be sentence cased?
```suggestion
// Defaulter handles LimaVM mutation via webhook.
```
Obviously it's also fine as-is too, just nit picking on grammar…

mook-as (approved):

pkg/controllers/lima/limavm/controller.go
jandubois: It is a godoc convention:

> "Comments should begin with the name of the thing being described"

If seems to be followed by the core runtime library, e.g. (randomly picked from `fmt/format.go`):

```go
// fmtBoolean formats a boolean.
func (f *fmt) fmtBoolean(v bool) {
	if v {
		f.padString("true")
	} else {
		f.padString("false")
	}
}
```

PR #62, opened 2025-11-05
controllers/base: Add test for DeleteOwnedResources

### LINKED ISSUE #61: Create test for DeleteOwnedResources
It would be good to test the code in #59 in detail; however, it isn't clear how to best go about doing so.

mook-as:
This tests DeleteOwnedResources because its logic seems complicated. Ideally we would also test a CRD, but creating a new resource kind at runtime is not working as expected at the moment; that will need to happen in a later commit.

Fixes #61

As usual, now that the test exists I'm wondering if that was worth it :D


jandubois:
Test doesn't pass for me:

```console
❯ g st
On branch mook-as/base/finalizer-delete-test
Your branch is up to date with 'upstream/mook-as/base/finalizer-delete-test'.

nothing to commit, working tree clean

❯ go test ./pkg/controllers/base
--- FAIL: TestDeleteOwnedResources (6.49s)
    --- FAIL: TestDeleteOwnedResources/delete_owned_config_map_config_map_without_finalizers (0.03s)
        finalizer_test.go:118: assertion failed: apierrors.IsNotFound(err) is true: owned config map was not deleted: configmaps "owned" not found
    --- FAIL: TestDeleteOwnedResources/delete_owned_config_map_config_map_with_finalizer (0.20s)
        finalizer_test.go:118: assertion failed: apierrors.IsNotFound(err) is true: owned config map was not deleted: configmaps "owned" not found
FAIL
FAIL	github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base	6.934s
FAIL
```

mook-as:
Thanks, I had issues porting from `testing` to `gotest.tools/v3` as requested by the linter, and accidentally inverted a condition.  That's fixed now, and I added `go test` to CI so it actually shows up next time.

pkg/controllers/base/finalizer_test.go
```diff
@@ -0,0 +1,135 @@
...
+				owner,
+				mgr)
+			assert.NilError(t, err, "failed to delete owned resources")
+
+			// Check that only the owned config map was deleted
+			err = c.Get(t.Context(), client.ObjectKeyFromObject(owned), owned)
+			assert.Check(t, apierrors.IsNotFound(err), "owned config map was not deleted: %s", err)
+			err = c.Get(t.Context(), client.ObjectKeyFromObject(owner), owner)
+			assert.Check(t, cmp.ErrorIs(err, nil), "failed to get owner object")
+			err = c.Get(t.Context(), client.ObjectKeyFromObject(unrelated), owner)
```
jandubois: Looks like a copy&pasto:

```suggestion
			err = c.Get(t.Context(), client.ObjectKeyFromObject(unrelated), unrelated)
```

pkg/controllers/base/finalizer_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("delete owned config map %s", tc.name), func(t *testing.T) {
+			// Clean up all the created config maps
+			defer assert.Check(
+				t,
+				cmp.ErrorIs(c.DeleteAllOf(t.Context(), &corev1.ConfigMap{}, client.InNamespace(metav1.NamespaceDefault)), nil),
+				"failed to delete all config maps")
```
jandubois: I can't quite wrap my head around this, but the cleanup probably won't work if the test fails because `DeleteAllOf()` doesn't handle finalizers, and there is no garbage collector.

So a "safer" way to do this would be to list all ConfigMaps, remove their finalizers, and then use `DeleteAllOf()` to remove them. Not sure if it is worth the bother.

jandubois (changes requested):
I'm pretty sure the cut&pasto should be fixed.

I'm fine with keeping the `defer` code because it works as long as the test doesn't fail. 😄

pkg/controllers/base/finalizer_test.go
mook-as: I did this first, and then thought that this was too much code and obscured the logic of the actual test.

So I just renamed all the config maps to be unique, and then _not_ do the cleanup instead.  This is a temporary cluster (from `envtest`) anyway, it only lasts until `TestDeleteOwnedResources` returns.

pkg/controllers/base/finalizer_test.go
```diff
@@ -0,0 +1,129 @@
...
+			finalizers: []string{FinalizerName},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("delete owned config map %s", tc.name), func(t *testing.T) {
+			// Create a new config map that will be the owner
+			owner := &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("owner-%d", len(tc.finalizers)),
```
jandubois: I don't understand the `len(tc.finalizers)` part. Is this supposed to be different for each test case? Why not let use the index from `for i, tc := range testCases`?

jandubois (dismissed):
Thanks, LGTM

I just don't understand the `len()` thing. If you think that's what is needed, then go ahead and merge (and leave an explanation).

pkg/controllers/base/finalizer_test.go
mook-as: … Actually yes that's way easier isn't it? Yeah, I'll do that.

jandubois (approved):
Thanks, LGTM

PR #64, opened 2025-11-06
Implement namespace controller

jandubois:
We can use the Kubernetes NamespaceLifecycle admission controller to set the finalizer on namespaces as they are being created, but we cannot use the Namespaces controller because it requires controller-manager/kubelet.

### DIFF
--- a/bats/tests/30-builtin-controllers/namespace.bats
+++ b/bats/tests/30-builtin-controllers/namespace.bats
@@ -1,0 +1,66 @@
+load '../../helpers/load'
+
+# Namespace deletion tests - tests the built-in namespace controller
+# that handles cleanup of resources when a namespace is deleted.
+
+local_setup_file() {
+    setup_rdd_control_plane "*"
+}
+
+@test 'namespace deletion with multiple resource types' {
+    rdd ctl create namespace "multi-ns"
+
+    # Create various core resources
+    rdd ctl create configmap test-cm -n multi-ns --from-literal=key=value
+    rdd ctl create secret generic test-secret -n multi-ns --from-literal=password=secret
+
+    # Create custom resources from different API groups
+    # ConfigMapReplicaSet (rdd.rancherdesktop.io)
+    rdd ctl apply -f - <<EOF
+apiVersion: rdd.rancherdesktop.io/v1alpha1
+kind: ConfigMapReplicaSet
+metadata:
+  name: test-cmrs
+  namespace: multi-ns
+spec:
+  replicas: 2
+  data:
+    key: value
+EOF
+
+    # Notary (rdd.rancherdesktop.io)
+    rdd ctl apply -f - <<EOF
+apiVersion: rdd.rancherdesktop.io/v1alpha1
+kind: Notary
+metadata:
+  name: test-notary
+  namespace: multi-ns
+spec:
+  value: "test-value"
+  configMapName: "notary-history"
+EOF
+
+    # LimaVM (lima.rancherdesktop.io)
+    rdd ctl create configmap lima-template -n multi-ns --from-literal=template='{"vmType":"vz"}'
+    rdd ctl apply -f - <<EOF
+apiVersion: lima.rancherdesktop.io/v1alpha1
+kind: LimaVM
+metadata:
+  name: test-vm
+  namespace: multi-ns
+spec:
+  running: false
+  templateRef:
+    name: lima-template
+EOF
+
+    # Verify all resources exist
+    rdd ctl get configmap test-cm -n multi-ns
+    rdd ctl get secret test-secret -n multi-ns
+    rdd ctl get configmapreplicaset test-cmrs -n multi-ns
+    rdd ctl get notary test-notary -n multi-ns
+    rdd ctl get limavm test-vm -n multi-ns
+
+    # Delete namespace
+    rdd ctl delete namespace "multi-ns" --timeout=20s
+}
--- a/pkg/controllers/builtin/namespace/controller.go
+++ b/pkg/controllers/builtin/namespace/controller.go
@@ -20,37 +20,32 @@ const ControllerName = "namespace"
 const APIGroup = "builtin"
 
 func init() {
-	base.RegisterController(NewController())
+	base.RegisterController(&controller{})
 }
 
-// Controller implements the base.Controller interface for namespace lifecycle management.
-type Controller struct{}
+// controller implements the base.Controller interface for namespace lifecycle management.
+type controller struct{}
 
-// Verify that Controller implements base.Controller interface.
-var _ base.Controller = &Controller{}
-
-// NewController creates a new Controller instance.
-func NewController() *Controller {
-	return &Controller{}
-}
+// Verify that controller implements base.Controller interface.
+var _ base.Controller = &controller{}
 
 // GetName returns the controller name.
-func (c *Controller) GetName() string {
+func (c *controller) GetName() string {
 	return ControllerName
 }
 
 // GetAPIGroup returns the API group this controller belongs to.
-func (c *Controller) GetAPIGroup() string {
+func (c *controller) GetAPIGroup() string {
 	return APIGroup
 }
 
 // GetCRDData returns empty string since namespace is a built-in Kubernetes resource.
-func (c *Controller) GetCRDData() string {
+func (c *controller) GetCRDData() string {
 	return ""
 }
 
 // RegisterWithManager implements the complete controller registration.
-func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
+func (c *controller) RegisterWithManager(mgr ctrl.Manager) error {
 	klog.InfoS("Setting up namespace controller watch", "controller", ControllerName)
 
 	// Register the controller
--- a/pkg/controllers/builtin/namespace/controller.go
+++ b/pkg/controllers/builtin/namespace/controller.go
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package namespace
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/builtin/namespace/controllers"
+)
+
+// ControllerName is the name of this controller.
+const ControllerName = "namespace"
+
+// APIGroup is the API group this controller belongs to.
+const APIGroup = "builtin"
+
+func init() {
+	base.RegisterController(NewController())
+}
+
+// Controller implements the base.Controller interface for namespace lifecycle management.
+type Controller struct{}
+
+// Verify that Controller implements base.Controller interface.
+var _ base.Controller = &Controller{}
+
+// NewController creates a new Controller instance.
+func NewController() *Controller {
+	return &Controller{}
+}
+
+// GetName returns the controller name.
+func (c *Controller) GetName() string {
+	return ControllerName
+}
+
+// GetAPIGroup returns the API group this controller belongs to.
+func (c *Controller) GetAPIGroup() string {
+	return APIGroup
+}
+
+// GetCRDData returns empty string since namespace is a built-in Kubernetes resource.
+func (c *Controller) GetCRDData() string {
+	return ""
+}
+
+// RegisterWithManager implements the complete controller registration.
+func (c *Controller) RegisterWithManager(mgr ctrl.Manager) error {
+	klog.InfoS("Setting up namespace controller watch", "controller", ControllerName)
+
+	// Register the controller
+	// Note: Resource discovery happens dynamically during each reconciliation
+	// to ensure we always have the most up-to-date list of namespaced resources
+	err := ctrl.NewControllerManagedBy(mgr).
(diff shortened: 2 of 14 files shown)


bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+  namespace: test-ns
+spec:
+  running: false
+  templateRef:
+    name: src-template
+EOF
+
+    # Verify LimaVMs exist and have finalizers
+    run -0 rdd ctl get limavm "vm1" -n "test-ns" -o jsonpath='{.metadata.finalizers}'
+    assert_output --partial "rdd.rancherdesktop.io/cleanup"
```
mook-as: I was trying to see if I can get `assert_line` to work instead, but the result was rather ugly, so I guess keep it as is:
```sh
rdd ctl get limavm foo -o go-template='{{ range .metadata.finalizers }}{{ printf "%s\n" .}}{{end}}'
```

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+
+    # Wait for namespace deletion to complete
+    # The namespace controller should:
+    # 1. Discover all namespaced resource types dynamically
+    # 2. Delete all resources (using Patch to remove their finalizers efficiently)
+    # 3. Remove the kubernetes finalizer from the namespace
+    # With internal looping (100ms delay between iterations), actual deletion completes in ~400-500ms.
+    # Test polls every 1s, so total time from delete to verification is ~1s.
+    # Timeout set to 30s to catch performance regressions while allowing CI variance.
+    try --max 30 --delay 1 -- eval "! rdd ctl get namespace test-ns 2>/dev/null"
```
mook-as: As in #65, can we do
```sh
rdd ctl wait --for=delete namespace test-ns
```
The timeout is 30s by default, but being explicit here may be better.

Also, why are we doing `rdd ctl delete … --wait=false`, and then waiting for the deletion to finish? Wouldn't just `--wait=true` be fine?

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+
+    # Verify all resources were deleted
+    run -1 rdd ctl get limavm vm1 -n test-ns
+    assert_output --partial "not found"
```
mook-as: If the namespace no longer exists, how can we get the limavm inside of it?

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+@test 'namespace deletion with no resources' {
+    # Create an empty namespace
+    rdd ctl create namespace "empty-ns"
+    run -0 rdd ctl get namespace empty-ns
+
+    # Delete it immediately
+    rdd ctl delete namespace "empty-ns" --wait=false
+
+    # Should be deleted quickly since no resources to clean up
+    try --max 10 --delay 1 -- eval "! rdd ctl get namespace empty-ns 2>/dev/null"
```
mook-as: `rdd ctl wait --for=delete…`

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+    assert_output --partial "not found"
+
+    run -1 rdd ctl get configmap src-template -n test-ns
+    assert_output --partial "not found"
+}
+
+@test 'namespace deletion with no resources' {
+    # Create an empty namespace
+    rdd ctl create namespace "empty-ns"
+    run -0 rdd ctl get namespace empty-ns
```
mook-as: Should we check that we didn't add any finalizers?

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+
+@test 'namespace deletion with resources without finalizers' {
+    # Create a namespace
+    rdd ctl create namespace "simple-ns"
+
+    # Create some ConfigMaps (no finalizers)
+    rdd ctl create configmap cm1 -n simple-ns --from-literal=key=value1
+    rdd ctl create configmap cm2 -n simple-ns --from-literal=key=value2
+
+    # Verify ConfigMaps exist
```
mook-as: Verify that the namespace didn't get any finalizers?

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+    run -0 rdd ctl get configmap cm1 -n simple-ns
+    run -0 rdd ctl get configmap cm2 -n simple-ns
+
+    # Delete namespace
+    rdd ctl delete namespace "simple-ns" --wait=false
+
+    # Should be deleted quickly since resources have no finalizers (~5-10s)
```
mook-as: ~5-10s is an order of magnitude bigger than the with finalizers case, according to the comment above‽

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,195 @@
...
+  name: test-vm
+  namespace: multi-ns
+spec:
+  running: false
+  templateRef:
+    name: lima-template
+EOF
+
+    # Verify all resources exist
+    run -0 rdd ctl get configmap test-cm -n multi-ns
```
mook-as: If we use the `configmap/test-cm` form, I think we can just stick them all into an array and loop over it? That makes sure the checks before and after delete use the same values…

pkg/service/cmd/service.go
```diff
@@ -687,11 +689,13 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
...
 	// Get all registered controllers from the registry
 	allControllers := base.GetAllControllers()
-
-	// Get enabled controllers from the --controllers flag directly
 	controllersSpec := completed.Controllers.Controllers
+
+	// Enable controllers: builtin controllers are always enabled, others based on --controllers flag
 	for _, controller := range allControllers {
-		if shouldEnableController(controller, controllersSpec) {
+		if controller.GetAPIGroup() == "builtin" {
```
mook-as: Should this live inside of `shouldEnableController` instead?

pkg/controllers/builtin/namespace/controller.go
```diff
@@ -0,0 +1,72 @@
...
+
+// APIGroup is the API group this controller belongs to.
+const APIGroup = "builtin"
+
+func init() {
+	base.RegisterController(NewController())
+}
+
+// Controller implements the base.Controller interface for namespace lifecycle management.
+type Controller struct{}
```
mook-as: Does this need to be exported? Same with `NewController`, which is only called by `init()` (does that even need to exist?)

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
```diff
@@ -0,0 +1,178 @@
...
+const (
+	// kubernetesFinalizer is the standard Kubernetes namespace finalizer.
+	// This finalizer is automatically added to all namespaces by the NamespaceLifecycle
+	// admission plugin (enabled in pkg/service/admission/config.go). The admission plugin
+	// adds it during CREATE, ensuring all namespaces (including system namespaces like
+	// default and rdd-system) have the finalizer from the start.
+	//
+	// This controller is responsible for removing the finalizer after cleaning up all
+	// resources in the namespace during deletion.
+	kubernetesFinalizer = "kubernetes"
```
mook-as, line 33: Oh, the admission plugin isn't ours, but belongs to upstream? That should probably be called out in the comment; I was wondering why the finalizer isn't scoped to our domain, and couldn't find the admission plugin… :D

I think a small tweak is fine:
> … added to all namespace by the upstream NamespaceLifecycle admission plugin …

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
```diff
@@ -0,0 +1,178 @@
...
+	log.Info("Processing namespace deletion", "namespace", namespace.Name)
+
+	resourceTypes, err := base.DiscoverNamespacedResources(ctx, r.Manager)
+	if err != nil {
+		log.Error(err, "Failed to discover namespaced resources")
+		return ctrl.Result{}, err
+	}
+
+	// Loop internally until all resources are deleted or no progress can be made.
+	// This avoids re-queueing and rate limiter delays when making progress.
```
mook-as, line 81: And this just requeues if it gets stuck, which is good in case new resource types get registered during the loop. 👍

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
```diff
@@ -0,0 +1,178 @@
...
+	// - GetAPIReader() bypasses the cache and reads directly from the API server, avoiding watches
+	apiReader := r.Manager.GetAPIReader()
+
+	totalRemaining := 0
+	for _, gvk := range resourceTypes {
+		list := &metav1.PartialObjectMetadataList{}
+		list.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   gvk.Group,
+			Version: gvk.Version,
+			Kind:    gvk.Kind + "List",
```
mook-as, line 141: I feel like this might bite us later if the list kind _isn't_ just the kind with _List_ appended, but it's probably fine for now.

mook-as (changes requested):

bats/tests/30-builtin-controllers/namespace.bats
jandubois: Yeah, we either have to assume there is only a single finalizer (not unreasonable in the test), or use `jq` or `yq` if we want to split them into lines. I think it is good as-is for now.

bats/tests/30-builtin-controllers/namespace.bats
jandubois: Yes, this doesn't make sense; I'll remove the test. It is redundant with the previous one.

bats/tests/30-builtin-controllers/namespace.bats
jandubois: I feel that now that the controller works, the first 3 tests are redundant. I will delete them and only keep test 4.

bats/tests/30-builtin-controllers/namespace.bats
jandubois: Some of the timings were done when the controller didn't actually work correctly. Or I might have messed things up when I updated the wrong comments.

bats/tests/30-builtin-controllers/namespace.bats
jandubois: As you pointed out above, the "after" checks are pointless because they fail when the namespace is gone. They cannot detect orphaned resources, so I removed them.

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
jandubois: Practically speaking, all RDD resource types should be registered during cluster start, so these would be theoretical additional resources a user would create. But still good to be robust.

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
jandubois: I tried to look it up, but it turns out to be more complicated than I thought. 

It also seems to be kind of hard-wired into the apiserver code. I suspect the `listKind` field in the CRD exists for historical reasons only, and think resources will not work properly with `kubectl` if you use a different name. I didn't try it though because I got tired of it.

pkg/controllers/builtin/namespace/controller.go
jandubois: My only requirement is that the code for all controllers looks the same. I tried before to change this, and ran into a problem, but I just noticed that the `demo_controller.go` file incorrectly used `controller` instead of `controllers` as the package name, and created a clash.

I may change this as a second commit, but given that it will change all the controllers it could have been a separate PR as well.

bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -0,0 +1,66 @@
...
+
+    # Verify all resources exist
+    rdd ctl get configmap test-cm -n multi-ns
+    rdd ctl get secret test-secret -n multi-ns
+    rdd ctl get configmapreplicaset test-cmrs -n multi-ns
+    rdd ctl get notary test-notary -n multi-ns
+    rdd ctl get limavm test-vm -n multi-ns
+
+    # Delete namespace
+    rdd ctl delete namespace "multi-ns" --timeout=20s
```
mook-as, line 65: I guess we basically rely on the delete succeeding here (because if it fails the timeout will expire).  Good enough.

pkg/controllers/builtin/namespace/controller.go
```diff
@@ -0,0 +1,67 @@
...
+
+func init() {
+	base.RegisterController(&controller{})
+}
+
+// controller implements the base.Controller interface for namespace lifecycle management.
+type controller struct{}
+
+// Verify that controller implements base.Controller interface.
+var _ base.Controller = &controller{}
```
mook-as, line 30: I'm sure `base.RegisterController` would have caught this, but it's harmless anyway so no need to change anything (and we can be explicit this way).

pkg/controllers/builtin/namespace/controllers/namespace_controller.go
mook-as: Yeah, a brief look didn't show how we'd get the list type without using the convention, so I gave up too.

mook-as (approved):
LGTM

pkg/controllers/builtin/namespace/controller.go
jandubois: It is mostly documentation, and gives me a way in the IDE to jump to the interface definition...

PR #65, opened 2025-11-06
BATS: lima/admission: fix wait

mook-as:
I was dumb and made a comment in https://github.com/rancher-sandbox/rancher-desktop-daemon/pull/63#discussion_r2496433447 and forgot about it and merged. So this fixes that.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-admission.bats
+++ b/bats/tests/33-lima-controllers/limavm-admission.bats
@@ -158,7 +158,7 @@ EOF
     assert_output --partial "deleted"
 
     # Wait for deletion to complete (asynchronous Kubernetes deletion)
-    try --max 10 --delay 1 --until-fail -- rdd ctl get limavm "my-vm" --namespace "test-ns1"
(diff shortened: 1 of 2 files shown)


bats/tests/33-lima-controllers/limavm-admission.bats
```diff
@@ -158,7 +158,7 @@ EOF
     assert_output --partial "deleted"
 
     # Wait for deletion to complete (asynchronous Kubernetes deletion)
-    try --max 10 --delay 1 --until-fail -- rdd ctl get limavm "my-vm" --namespace "test-ns1"
+    rdd ctl wait --for=delete limavm "my-vm" --namespace "test-ns1"
```
jandubois: To match the original command (and catch performance regressions):

```suggestion
    rdd ctl wait --for=delete limavm "my-vm" --namespace "test-ns1" --timeout=10s
```

jandubois: There is also another instance of this pattern in `limavm-lifecycle.bats` that should be treated the same way.

jandubois (approved):
Thanks, LGTM

PR #67, opened 2025-11-06
CI: BATS: run in parallel

mook-as:
Since we have multiple BATS tests, we can run them in parallel (as separate jobs) so the whole run completes faster.

### DIFF
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -58,6 +58,9 @@ ALLOCA
 ldflags
 sprintf
 
+# jq
+rtrimstr
+
 # make
 addprefix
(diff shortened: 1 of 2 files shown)


mook-as:
This basically ~halves the run time (~20m to ~10m), except for the run where `wsl --install` was falling over.  Not sure that's worth the extra compute, though.

We can re-visit this in the future once we have more test suites, I suppose?

jandubois:
> Not sure that's worth the extra compute, though.

Why does it require extra compute? Just to set up the additional runners?

I think reducing CI time from 20m to 10m is definitely worth it for increased turnaround time. Tests are only going to take longer as we add more. We don't even have tests starting VMs yet.

PR #69, opened 2025-11-07
readiness: Simplify checking CRD establishment

mook-as:
We are occasionally getting timeouts in CI when waiting for the CRDs to be established; try listing all CRDs and then watching based on that resource version to make sure we don't miss anything.

### DIFF
--- a/bats/tests/20-service/4-start-parameters.bats
+++ b/bats/tests/20-service/4-start-parameters.bats
@@ -2,52 +2,51 @@ load '../../helpers/load'
 
 # Controller testing using CRD presence validation
 
-assert_rdd_controllers() {
-    run -0 rdd ctl get CustomResourceDefinition notaries.rdd.rancherdesktop.io
-    assert_output --partial "notaries.rdd.rancherdesktop.io"
+CRD_GROUP=customresourcedefinition.apiextensions.k8s.io # spellchecker:ignore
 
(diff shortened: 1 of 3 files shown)


bats/tests/20-service/4-start-parameters.bats
```diff
@@ -2,52 +2,51 @@ load '../../helpers/load'
...
-    run -0 rdd ctl get CustomResourceDefinition demos.app.rancherdesktop.io
-    assert_output --partial "demos.app.rancherdesktop.io"
+CRD_GROUP=customresourcedefinition.apiextensions.k8s.io # spellchecker:ignore
+
+get_controllers() {
+    run -0 rdd ctl get CustomResourceDefinition --output=name
+    # If `$output` is empty, `refute_lines` will fail because `$lines` is unset.
+    if [[ -z "${output}" ]]; then
+        run -0 echo '<no output>'
+    fi
```
jandubois, line 12: Ok, maybe I'm just slow today, but the comment did not really register with me. 

At first I thought: Useless use of `run -0`, just use `echo ...`.

Then I realized it was about setting `$output`. So why not use `$output=...`? You can even combine it with the condition: `output=${output:-...}`.

But I see this is really about setting `$lines`, so I guess this is all ok, and actually what the comment says. Not sure why I had to go through all this to realize that.

jandubois (approved):
Thanks, LGTM

PR #77, opened 2025-11-08
rdd: implement `rdd svc log`

### LINKED ISSUE #71: Create `rdd svc logs` command
It would be helpful to have a simple command to show the `rdd.stderr.log` file for the current `$RDD_INSTANCE` without having to specify the full path.

Ideally it should support `-f` and `--follow` options to work like `tail -f`. In that case it should also just display the last 10 lines of output, so it doesn't flood the scrollback of the terminal.

Maybe @Nino-K wants to take this one to reconnect with his old https://github.com/hpcloud/tail module. 🤣 

In the future we will also want an `rdd limavm logs INSTANCE` command with the same functionality, just showing the `ha.stderr.log` of the Lima instance, so make the implementation reusable.

mook-as:
This new command prints out the last known logs of the corresponding RDD instance.  By default, stderr logs are printed; `rdd svc log --stdout` can be used to print the stdout logs instead.  `rdd svc log --follow` can be used to tail the logs until the user interrupts it.

We are using the pulumi fork of `github.com/nxadm/tail` because that has some fixes around stopping at the end of the file that would be important for our use case (when not using `--follow`) that has pull requests upstream that have not been merged.

Fixes #71

### DIFF
--- a/cmd/rdd/service.go
+++ b/cmd/rdd/service.go
@@ -7,6 +7,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"strconv"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


jandubois:
Needs another rebase

jandubois (dismissed):
Thanks, LGTM

Needs another rebase.

jandubois (approved):

PR #92, opened 2025-12-09
Use golangci-lint as a Go tool with a pinned version

jandubois:


.github/workflows/bats.yaml
```diff
@@ -98,6 +98,17 @@ jobs:
         # Disable first-boot script
         wsl --distribution openSUSE --exec sed -i '/^command =/d' /etc/wsl-distribution.conf
 
+        # Make sure WSL interop is enabled
```
mook-as, line 101: I still don't know why we suddenly need this, but since we do I guess we're stuck with it.

mook-as (approved):

PR #94, opened 2025-12-10
Use Powershell HEREDOC instead of bash HEREDOC

jandubois:
It looks like the command was not executed by bash; possibly because the EOF marker had a carriage return at the beginning from the previous line.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -99,14 +99,16 @@ jobs:
         wsl --distribution openSUSE --exec sed -i '/^command =/d' /etc/wsl-distribution.conf
 
         # Make sure WSL interop is enabled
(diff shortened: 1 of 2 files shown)


.github/workflows/bats.yaml
```diff
@@ -99,14 +99,16 @@ jobs:
...
         systemd=true
-
+        
         [interop]
         enabled=true
         appendWindowsPath=true
-        EOF'
+        "@
+             
+        $wslConf -replace "`r`n", "`n" | wsl --distribution openSUSE --exec sh -c 'cat > /etc/wsl.conf'
```
mook-as: Completely optional:
```suggestion
        $wslConf | wsl --distribution openSUSE --exec sh -c 'tr -d "\r" > /etc/wsl.conf'
```

mook-as (dismissed):
Can't be worse at least.

mook-as (approved):

Issue #88, opened 2025-11-28
Create API for container engine


mook-as:
We will need to have Kubernetes-ized APIs for the container engine (list containers, images, volumes, get logs, etc.) so that the front end can interact with it.  The design of this API is to be determined (need to see if there's appropriate APIs in the ecosystem we can use).

jandubois:
As a first step we can expose the socket locations on the App object status. It would be ideal if we just had the guest agent set up watches on the API and then update k8s resources to match container state etc, but for e.g. RDX support we may still need to expose the sockets to the GUI anyways.

jandubois:
I don't know of any APIs that would expose a container engine directly; it is really abstracted away by kubelet which manages containers internally. What we are probably looking for is something like (a subset of) CRI, but at the Kubernetes API level, and not as gRPC: https://github.com/kubernetes/cri-api/blob/v0.33.1/pkg/apis/runtime/v1/api.proto

mook-as:
Yeah, we'll need to expose the socket for both the extension API, as well as for the CLI (and other clients), we'll never be able to get away from that.

But I was assuming that we'd want the GUI to be using the rdd kube API for everything, and write controllers to do the management.  I don't want to be writing more TypeScript backend code for the GUI, if possible.

jandubois:
> I was assuming that we'd want the GUI to be using the rdd kube API for everything

yes

mook-as:
Looking into `logs` (and therefore `exec`), in terms of having an API endpoint that converts into websockets:

Per https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility:

Under _Other Subresources_, basically things of that nature are not supported by CRDs, and we need to use the aggregated API method for that (which means rewriting `containers.rancherdesktop.io` to use that instead).

There's also some documentation at https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/

https://github.com/kubernetes-sigs/apiserver-builder-alpha is the only code-generation tooling I can find, but it's in alpha, and hasn't been touched in four years.

At this point, we're probably going to have the log/exec APIs independent of the Kubernetes APIs, and have a completely separate endpoint (possibly on a different port).  We can still use the Kubernetes APIs to transmit information about this endpoint (e.g. container IDs).  However, this might complicate using RD from remote machines.

mook-as:
[...]

PR #95, opened 2025-12-10
Initial implementation of containers API

mook-as:
This adds the resource definitions for the containers API, for use by the front end.  While it implements the resources, it does _not_ implement any controllers.

This is the start of #88 — that would only get resolved by having controller implementations.  Once this is merged, I intend to first write controllers using static data so that the front end can be mocked; we will need to come back later to actually implement communicating with the proper container engines later.

As usual, the CRDs are generated; feel free to review that, or the api package sources, whichever is easier to read.

The container and volume APIs are based on what we actually use, per https://github.com/rancher-sandbox/rancher-desktop/blob/183991a092f3dad3fa7c0fea01c3bdd1ae1c168e/pkg/rancher-desktop/store/container-engine.ts#L63-L87 plus whatever the docker engine API has.


docs/design/api_containers.md
```diff
@@ -0,0 +1,104 @@
...
+`Container` objects reflect the running containers.  If the container engine
+supports namespaces, then they are in the namespace with the same name;
+otherwise, all containers are in the `default` namespace.
+
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: Container
+metadata:
+  name: 8eb6f2cf72b6616aa743cf9187f350af84c9749dab65474db2530f26745d2ef3 # container ID
+  namespace: default # matches containerd namespace, or `default` if not supported
```
jandubois: I think this should be a label, but all containers should be in the same namespace. The label gives us the same selection functionality, but prevents containers ending up in arbitrary namespaces that could be used by other applications.

There should be no possibility to have the same container id in multiple namespaces, so the name should be globally unique per container engine.

docs/design/api_containers.md
```diff
@@ -0,0 +1,104 @@
...
+`Image` objects reflect images in the container engine.  If the container engine
+supports namespaces, then they are in the namespace with the same name;
+otherwise, all images are in the `default` namespace.
+
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: Image
+metadata:
+  name: 'sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # Image ID
+  namespace: default # matches containerd namespace, or `default` if not supported
```
jandubois: Again, should be a label.

We don't have any plans to support multiple container engines, or multiple hosts right now, but if we ever do, then we need to use separate namespaces for each engine because the same image id can be stored by more than one engine, and `.metadata.name` must be unique in each namespace.

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -0,0 +1,146 @@
...
+
+const (
+	ContainerStatusCreated    ContainerStatusValue = "created"
+	ContainerStatusRunning    ContainerStatusValue = "running"
+	ContainerStatusPausing    ContainerStatusValue = "pausing"
+	ContainerStatusPaused     ContainerStatusValue = "paused"
+	ContainerStatusRestarting ContainerStatusValue = "restarting"
+	ContainerStatusRemoving   ContainerStatusValue = "removing"
+	ContainerStatusExited     ContainerStatusValue = "exited"
+	ContainerStatusDead       ContainerStatusValue = "dead"
```
jandubois, line 23: These look a bit like the Pod lifecycle phases, which have been deprecated in favour of conditions. It seems redundant to have both.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -0,0 +1,146 @@
...
+	ContainerStatusUnknown    ContainerStatusValue = "unknown"
+)
+
+// ContainerPortBinding describes one host port for the container to bind to.
+type ContainerPortBinding struct {
+	// hostIp is the host IP address that the container's port is mapped to.
+	// +required
+	HostIP string `json:"hostIp"`
```
jandubois: I don't know if there is a convention for JSON property names, but lowercasing the `P` just looks wrong to me:


```suggestion
	HostIP string `json:"hostIP"`
```

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// name of the port; in the form [port]/[protocol], e.g. "80/tcp".
+	// +required
+	Name string `json:"name"`
+	// bindings to the host port; can contain multiple entries to e.g. express
+	// IPv4 and IPv6 bindings.
+	// +required
+	Bindings []ContainerPortBinding `json:"bindings"`
+}
+
+// ContainerSpec defines the configuration the container was originally created with.
```
jandubois: I was wondering why it says "originally". The config is all immutable, so it should just say


```suggestion
// ContainerSpec defines the configuration the container was created with.
```

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -0,0 +1,146 @@
...
+	Error string `json:"error"`
+	// createdAt is the time this container was initially created.
+	// +optional
+	CreatedAt metav1.Time `json:"createdAt"`
+	// startedAt is the time this container was started; unset if the container is stopped.
+	// +optional
+	StartedAt metav1.Time `json:"startedAt"`
+	// finishedAt is the time this container was last stopped; unset if the container never ran.
+	// +optional
+	FinishedAt metav1.Time `json:"finishedAt"`
```
jandubois, line 95: I think it should be `StoppedAt`. "Finish" sounds like a Job thing. You don't start something again once it is finished.

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// conditions represent the calculated state of the container.
+	// Each condition has a unique type and reflects the status of a specific aspect of the resource.
+	//
+	// Known condition types include:
+	// - "Running": the container is running; this may also be paused.
+	// - "Paused": the container is paused.
+	// - "Restarting": the container is restarting.
+	// - "OOMKilled": a process within this container has been killed because it ran out of memory
+	//                since the container was last started.
+	// - "Dead": the container is dead.
```
jandubois, line 105: Please see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties for guidelines on condition names (should be adjectives, or past tense verbs, not present tense verbs).

jandubois (commented):
Looks good, just minor feedback.

Except for the `namespace` thing. I believe all objects for a single container engine should be stored in a single namespace. The container engine namespace just just be a label.

pkg/apis/containers/v1alpha1/containers_types.go
mook-as: That's what the docker engine API actually reports, I figured keeping the same name is clearer than trying to be clever.

pkg/apis/containers/v1alpha1/containers_types.go
mook-as: Those are actually names copied from the docker engine API; I was preferring keeping that in sync over a clean API, because looking it up would be easier.

pkg/apis/containers/v1alpha1/containers_types.go
jandubois: Yes, that is a good reason to keep the name; I thought it came from our internal Rancher Desktop 1.x data structures.

pkg/apis/containers/v1alpha1/containers_types.go
jandubois: Ok. We still have to see how these map to `containerd` names, which may be different.

mook-as: I don't intend to interact with `containerd` directly, but through `nerdctl`. In which case it presents a docker-compatible API (most of the time) anyway.  But we'll see how that goes.

jandubois: I assume you will create a controller that subscribes to events from the container engine and then updates the corresponding rdd resources to match. I assumed for that you would talk to the containerd API directly, kind of like our guestagent does. But I guess for things like images we still need to poll using `nerdctl`.

Longer term we could think about having an inotify client inside the VM that tells the controller when files have been modified in the image storage (just to trigger another poll, not to tell which images have changed). But let's keep it simple for now.

pkg/apis/containers/v1alpha1/containers_types.go
mook-as: This is just reflecting docker engine api `.status` field; it'll probably be used in the UI as the state of the container. For actual filtering and things, we do have conditions.

docs/design/api_containers.md
```diff
@@ -10,23 +10,25 @@ be rejected by the controllers.
 
 All objects are in the `containers.rancherdesktop.io` API group.
 
+If the container engine supports namespaces, it's listed as the `namespace`
+label rather than re-using the Kubernetes namespace.
```
jandubois: We should decide if we always want to specify the namespace label, even if the engine doesn't support/expose them. Just to simplify the code.

Then we need to decide if we should use `default` or `moby` when using `dockerd`. Because Docker does use the `moby` namespace in `containerd`.

pkg/apis/containers/v1alpha1/image_types.go
```diff
@@ -0,0 +1,70 @@
...
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ImageSpec defines the configuration the image was originally created with.
```
jandubois: There are several instances of "originally" that I think should all be deleted.

docs/design/api_containers.md
mook-as: Hmm, in that case that's just use `moby` then? We'll just hide it in the UI.

jandubois:
```
❯ rdctl shell dockerd --help | grep -A 1 containerd-namespace
      --containerd-namespace string           Containerd namespace to use
                                              (default "moby")
```

jandubois: Since we are going to use the `containerd-snapshotter` exclusively in RDD, I wonder if we can figure out for both `containerd` and `dockerd` to store images in the same location. Then we could try to set the `dockerd` namespace to `default` and maybe have the same images visible even when you switch container engines.

Doesn't work for k8s though, because when using `cri-dockerd` they end up in the default namespace and not `k8s.io`. So not sure if this is a good idea. It is an idea though. 😄

jandubois:
You need to recreate the PR from a branch inside the repo because the GitHub actions will not run for an external PR.

Issue #66, opened 2025-11-06
Notary events test flaky on Ubuntu


jandubois:
I've now seen this failure a second time (always on Ubuntu). Re-running fixes (hides) it:

```
RDD_INSTANCE=bats-rdd-controller ./lib/bats-core/bin/bats tests/31-rdd-controllers/
1..59
ok 1 create ConfigMapReplicaSet resource
...
ok 33 dry-run=server update rejects invalid values
not ok 34 verify event generation for spec updates
# (from function `try' in file helpers/utils.bash, line 328,
#  from function `wait_for_events_after_timestamp' in file tests/31-rdd-controllers/notary-events.bats, line 93,
#  in test file tests/31-rdd-controllers/notary-events.bats, line 131)
#   `wait_for_events_after_timestamp "events" "${timestamp}" "SpecUpdate"' failed
# notary.rdd.rancherdesktop.io/events created
#
#
#
# notary.rdd.rancherdesktop.io/events patched
#
#
# -- output equals, but it was expected to differ --
# output : 0
# --
ok 35 test no-change events with annotation updates
```

jandubois:
Saw it again:

```
RDD_INSTANCE=bats-rdd-controller ./lib/bats-core/bin/bats tests/31-rdd-controllers/
1..59
ok 1 create ConfigMapReplicaSet resource
ok 2 verify ConfigMapReplicaSet is created in Kubernetes
ok 3 wait for controller to create 3 ConfigMaps
ok 4 verify ConfigMaps have correct names
ok 5 verify ConfigMap content includes original data plus index
ok 6 verify ConfigMap owner references are set
ok 7 debug owner references before deletion
ok 8 scale ConfigMapReplicaSet up to 5 replicas
ok 9 wait for scaling up to complete (5 ConfigMaps)
ok 10 verify all 5 ConfigMaps exist after scaling up
ok 11 scale ConfigMapReplicaSet down to 2 replicas
ok 12 wait for scaling down to complete (2 ConfigMaps)
ok 13 verify only 2 ConfigMaps remain after scaling down
ok 14 update ConfigMapReplicaSet data
ok 15 verify ConfigMapReplicaSet has finalizer for cleanup
ok 16 delete ConfigMapReplicaSet
ok 17 verify parent resource deletion triggers finalizer cleanup
ok 18 wait for ConfigMaps to be cleaned up by finalizer
ok 19 create ConfigMapReplicaSet with zero replicas
[...]

PR #98, opened 2025-12-10
Skip flaky events test on Linux

jandubois:
Also includes running `make lint-fix`.

### DIFF
--- a/bats/tests/31-rdd-controllers/notary-events.bats
+++ b/bats/tests/31-rdd-controllers/notary-events.bats
@@ -67,7 +67,7 @@ get_events_after_timestamp() {
     local reason=$3
 
     run rdd ctl get events --field-selector involvedObject.name="${resource_name}" -o json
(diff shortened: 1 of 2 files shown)


bats/tests/31-rdd-controllers/notary-events.bats
```diff
@@ -106,6 +106,10 @@ get_latest_event_timestamp() {
 }
 
 @test 'verify event generation for spec updates' {
+    if is_linux; then
```
mook-as: Since this only fails on CI, should we also do `is_true $CI`?

mook-as (dismissed):
See also #66

bats/tests/31-rdd-controllers/notary-events.bats
jandubois: Sure, why not. Just hope CI stays green on the re-run now. 😄

mook-as (approved):

PR #99, opened 2025-12-11
Initial implementation of containers API

mook-as:
This is #95 but from the same repo due to GitHub Actions reasons. I'm leaving that open because it'll get closed once this merges.

This adds the resource definitions for the containers API, for use by the front end.  While it implements the resources, it does _not_ implement any controllers.

This is the start of #88 — that would only get resolved by having controller implementations.  Once this is merged, I intend to first write controllers using static data so that the front end can be mocked; we will need to come back later to actually implement communicating with the proper container engines later.

As usual, the CRDs are generated; feel free to review that, or the api package sources, whichever is easier to read.

The container and volume APIs are based on what we actually use, per https://github.com/rancher-sandbox/rancher-desktop/blob/183991a092f3dad3fa7c0fea01c3bdd1ae1c168e/pkg/rancher-desktop/store/container-engine.ts#L63-L87 plus whatever the docker engine API has.


jandubois (approved):
Thanks, LGTM

PR #102, opened 2025-12-16
Mock container data

mook-as:
This is a draft on top of #99 that implements the container API (with a webhook that prevents post-creation modification), plus a thing that generates a static set of data from a (hard-coded) JSON file.  This will be used to test out the API and build on top of before the backend is actually implemented (which would in turn depend on the VM being created).


bats/tests/34-containers-controllers/containers-mock.bats
```diff
@@ -0,0 +1,58 @@
...
+        read -r controller_pid <"${BATS_FILE_TMPDIR}/controller_pid"
+        kill "${controller_pid}" 2>/dev/null || true
+        wait "${controller_pid}" 2>/dev/null || true
+    fi
+}
+
+@test "containers are created" {
+    run -0 cat ../pkg/controllers/mock/testdata/containers.json
+    run -0 jq_output '.[].Id'
+    mapfile -t containers <<<"${output}"
```
jandubois: Do we have a test somewhere to make sure we have a newish `bash` version and are not using the macOS builtin one? Because bash 3.x has no support for `mapfile`.

bats/Makefile
```diff
@@ -16,7 +16,7 @@ check-bats:
 .PHONY: check-bats
 
 # Targets for the top level
-TOPLEVEL_TARGETS := build-rdd build-rdd-controller build-all-controllers
+TOPLEVEL_TARGETS := build-rdd build-rdd-controller build-mock-controller build-all-controllers
```
jandubois, line 19: Why is the mock controller not included `build-all-controllers`?

bats/tests/34-containers-controllers/containers-mock.bats
```diff
@@ -0,0 +1,58 @@
...
+local_teardown_file() {
+    if [[ -f "${BATS_FILE_TMPDIR}/controller_pid" ]]; then
+        read -r controller_pid <"${BATS_FILE_TMPDIR}/controller_pid"
+        kill "${controller_pid}" 2>/dev/null || true
+        wait "${controller_pid}" 2>/dev/null || true
+    fi
+}
+
+@test "containers are created" {
+    run -0 cat ../pkg/controllers/mock/testdata/containers.json
```
jandubois: The `../pkg/controllers/mock/testdata` path should be set in a global variable. 

And it should use `${PATH_BATS_ROOT}/../pkg/controllers` instead of relying on `bats` being invoked from the `bats/` directory.


```suggestion
    run -0 cat "${TESTDATA}/containers.json"
```

pkg/controllers/base/field_indexer.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get GVK for %T: %w", obj, err)
+	}
+	mapping, err := mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return fmt.Errorf("failed to get REST mapping for %T: %w", obj, err)
+	}
+	var crd apiextensionsv1.CustomResourceDefinition
+	// The client-side cache is typically not set up at this point, so we need to
+	// use ``.GetAPIReader()`` instead of .GetClient().
```
jandubois: Why are there 2 backquotes around `GetApiReader()` and none around `GetClient`?

This is not markdown; I think the trailing `()` makes it clear these are function calls:


```suggestion
	// use GetAPIReader() instead of GetClient().
```

pkg/controllers/base/field_indexer.go
```diff
@@ -0,0 +1,79 @@
...
+	for _, version := range crd.Spec.Versions {
+		if !version.Served {
+			continue
+		}
+		for _, field := range version.SelectableFields {
+			if field.JSONPath == "" {
+				continue
+			}
+			jp := jsonpath.New(field.JSONPath)
+			if err := jp.Parse("{" + field.JSONPath + "}"); err != nil {
```
jandubois, line 50: I was initially confused by this, so I think a comment explaining that `field.JSONPath` already starts with a `.` would be helpful. I guess once you know that it does, it is obvious that it has to.

pkg/controllers/containers/container/container_reconciler.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Find existing condition of this type
+	for i, condition := range container.Status.Conditions {
+		if condition.Type == conditionType {
+			// Update existing condition if status changed
+			if condition.Status != status || condition.Reason != reason || condition.Message != message {
+				container.Status.Conditions[i].Status = status
+				container.Status.Conditions[i].Reason = reason
+				container.Status.Conditions[i].Message = message
+				container.Status.Conditions[i].LastTransitionTime = now
```
jandubois: The `LastTransitionTime` is supposed to be set only when the `Status` changes but not when e.g. only the message is updated.

pkg/controllers/mock/container_reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+		if err := r.Get(ctx, namespacedName, &existingContainer); apierrors.IsNotFound(err) {
+			if err := r.Create(ctx, &targetContainer); err != nil {
+				errs = append(errs, fmt.Errorf("failed to create static container %s: %w", namespacedName, err))
+			}
+		} else if err != nil {
+			log.Error(err, "Failed to get static container", "name", namespacedName)
+			errs = append(errs, err)
+		} else {
+			targetContainer.ResourceVersion = existingContainer.ResourceVersion
+			if equality.Semantic.DeepEqual(existingContainer, targetContainer) {
```
jandubois: I think this will never be `true` because the `existingContainer` has all the metadata from the API server (like `ResourceVersion`) whereas the `targetContainer` has only the fields actually set here.

pkg/controllers/mock/container_reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+				continue // Nothing to do here.
+			}
+			if err := r.Update(ctx, &targetContainer); err != nil {
```
jandubois, line 136: I'm not sure if this will update the `Status` subresource. You probably need to also call `r.Status().Update()` as well.

pkg/controllers/mock/container_reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+
+// Reconcile implements [reconcile.TypedReconciler].
+func (r *containerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (reconcile.Result, error) {
+	log := log.FromContext(ctx)
+
+	var errs []error
+	var inspects []mobycontainer.InspectResponse
+	if err := json.Unmarshal(testContainers, &inspects); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to load static test data: %w", err)
+	}
```
jandubois: There is no reason to `Unmarshal()` the data on each `Reconcile()` cycle; it can be done once in an `init` function.

Same for the other mock reconcilers.

pkg/controllers/mock/mock_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	c.webhookPort = port
+}
+
+func (c *controller) GetCRDData() string {
+	return ""
+}
+
+// GetName implements [base.Controller].
+func (c *controller) GetName() string {
+	return controllerName + "-webhook"
```
jandubois: What is different about this controller? None of our other controllers seem to append "-webhook" to the name. I guess it doesn't matter, but seems inconsistent.

And fully nit-picking: `GetName` is normally the first method we define, before `GetAPIGroup`. 😄 


```suggestion
	return controllerName
```

jandubois (changes requested):
First round of feedback. The PR is kinda big, so I probably missed something.

bats/tests/34-containers-controllers/containers-mock.bats
mook-as: Ah, you're right, I'll switch this to `while read` instead.

bats/tests/34-containers-controllers/containers-mock.bats
mook-as: Makes sense; will fix.

bats/Makefile
mook-as: No, this is used as a dependency of `bats-containers`.  The same thing is already done for `build-rdd-controller` / `bats-rdd`.

pkg/controllers/base/field_indexer.go
mook-as: I think I meant to add it around both and messed up.  Will remove.

pkg/controllers/base/field_indexer.go
mook-as: I'll add a comment with an example (`.status.repoTag`).

pkg/controllers/containers/container/container_reconciler.go
mook-as: I'll make the logic a bit more complex but do the correct thing.

I'll also update the demo controller, where this logic was copied from.

pkg/controllers/mock/container_reconciler.go
mook-as: Good idea; I'll just stick it in `SetupWithManager` though.

pkg/controllers/mock/container_reconciler.go
mook-as: We copy the `ResourceVersion` on the previous line, but yeah that makes sense (especially `Status.Conditions`).

pkg/controllers/mock/container_reconciler.go
mook-as: Good catch, it probably wouldn't.

pkg/controllers/mock/mock_controller.go
mook-as: Fixed; I just accidentally copied the code from `GetWebhookServiceName` instead…

pkg/controllers/mock/container_reconciler.go
```diff
@@ -0,0 +1,188 @@
...
+			log.Info("Namespace not found, requeuing", "namespace", req.NamespacedName)
+			return ctrl.Result{RequeueAfter: time.Second}, nil
+		}
+		log.Error(err, "Failed to get namespace", "namespace", req.NamespacedName)
+		return ctrl.Result{}, err
+	}
+	gvk, err := r.Client.GroupVersionKindFor(&rddNamespace)
+	if err != nil {
+		log.Error(err, "Failed to get GVK for namespace", "namespace", &rddNamespace)
+	}
```
jandubois, line 71: So I think this probably cannot happen anyways, so it doesn't really matter. But if there really was an error, shouldn't we return the error instead of continuing with `gvk` in an undefined state? It would probably start end endless loop (with exponential backoff), but at least the error would be more obvious.

pkg/controllers/mock/mock_webhook.go
```diff
@@ -0,0 +1,82 @@
...
+
+// ValidateDelete implements ctrlwebhookadmission.CustomValidator.
+func (c *staticNamespaceNoDeleteValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (warnings ctrlwebhookadmission.Warnings, err error) {
+	ns, ok := obj.(*corev1.Namespace)
+	if !ok {
+		return nil, errors.New("object is not a Namespace")
+	}
+	if ns.Name == mockNamespaceName {
+		klog.FromContext(ctx).V(4).Info("Rejecting delete of namespace")
+		return nil, apierrors.NewConflict(
```
jandubois: This should return "forbidden", not "conflict". Conflict means a ResourceVersion mismatch.

jandubois (dismissed):
Thanks, LGTM, but needs a rebase

pkg/controllers/mock/container_reconciler.go
jandubois: Sorry, I didn't mention it, but the same issue applies to the image and volume reconcilers too, as they share the same code.

jandubois (approved):
Thanks, LGTM

PR #103, opened 2025-12-18
Allow multiple controller managers

mook-as:
This makes it possible to support multiple shared controller managers (each in a different process).  This is to make it easier to reuse the code; we can have multiple processes running different controllers in the future to e.g. allow testing-only controllers to be used with the normal build.

The config map is changed so that the key is the name of the controller process (with the default one named "embedded"), and the value is JSON-serialized `ControllerManagerInfo`.  JSON is chosen so that we can dig into it as part of the BATS tests.  We also no longer track the instance in the info, because that is fixed per Kubernetes API controller so it does not
actually provide any useful information.

To test:
- Build all the executable (with `make build`)
- Run the executable with a subset of controllers:
  `bin/rdd --controllers rdd,lima`
- Run the app controller externally:
  `bin/app-controller`

Everything should work as before, the only difference being multiple
controller processes co-existing.

#102 will be rebased on top of this once this lands.

### DIFF
--- a/bats/tests/31-rdd-controllers/notary-external.bats
+++ b/bats/tests/31-rdd-controllers/notary-external.bats
@@ -134,3 +134,57 @@ EOF
     # External controllers check every 2 seconds, allow up to 3 failures (6 seconds), plus buffer
     try --max 15 --delay 1 -- assert_process_exited "${controller_pid}"
 }
+
+@test "control plane starts with different controller" {
+    rdd service delete
+    rdd service create --controllers="lima"
+    rdd service start
+}
+
+@test "control plane starts without rdd controller" {
+    # Confirm apiserver is working
+    run -0 rdd ctl get namespaces -o name
+    assert_line namespace/default
+
+    # Verify that the embedded lima controller is running
+    run -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system --output jsonpath='{.data.embedded}'
+    assert_output
+    jq_output '.enabledControllers[]'
+    assert_line "limavm"
+    refute_line "notary"
+
+    # Verify that the external rdd controller is not running
+    run -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system --output jsonpath='{.data.rdd}'
+    refute_output
+}
+
+@test "external controller runs and registers" {
+    # Start external rdd-controller binary in background
+    "rdd-controller${EXE}" &
+    # Store PID to verify it auto-exits later
+    echo "$!" >"${BATS_FILE_TMPDIR}/controller_pid"
+
+    # Wait for external controller to register in discovery system
+    try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system --allow-missing-template-keys=false --output jsonpath='{.data.rdd}'
+
+    # Verify the discovery ConfigMap contains notary controller
+    run -0 --separate-stderr rdd ctl get configmap rdd-controller-manager --namespace rdd-system --output jsonpath='{.data.rdd}'
+    run -0 jq_output '.enabledControllers[]'
+    assert_output
+    assert_line "notary"
+}
+
+@test "shut down external controller" {
+    # Get the stored PID from the controller start test
+    controller_pid=$(cat "${BATS_FILE_TMPDIR}/controller_pid")
+
+    # Verify the external controller process is currently running
+    run -0 kill -0 "${controller_pid}"
+
+    # Stop the control plane - this should trigger auto-cleanup of external controllers
+    run -0 rdd svc stop
+
+    # Wait for external controller to detect control plane shutdown and exit
+    # External controllers check every 2 seconds, allow up to 3 failures (6 seconds), plus buffer
+    try --max 15 --delay 1 -- assert_process_exited "${controller_pid}"
+}
--- a/bats/tests/31-rdd-controllers/notary-external.bats
+++ b/bats/tests/31-rdd-controllers/notary-external.bats
@@ -38,7 +38,7 @@ assert_process_exited() {
     try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system
 
     # Verify the discovery ConfigMap contains notary controller
-    run -0 --separate-stderr rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.rdd}'
+    run_e -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.rdd}'
     run -0 jq_output '.enabledControllers[]'
     assert_line "notary"
 }
@@ -179,10 +179,10 @@ EOF
     controller_pid=$(cat "${BATS_FILE_TMPDIR}/controller_pid")
 
     # Verify the external controller process is currently running
-    run -0 kill -0 "${controller_pid}"
+    kill -0 "${controller_pid}"
 
     # Stop the control plane - this should trigger auto-cleanup of external controllers
-    run -0 rdd svc stop
+    rdd service stop
 
     # Wait for external controller to detect control plane shutdown and exit
     # External controllers check every 2 seconds, allow up to 3 failures (6 seconds), plus buffer
--- a/bats/tests/31-rdd-controllers/notary-external.bats
+++ b/bats/tests/31-rdd-controllers/notary-external.bats
@@ -38,8 +38,9 @@ assert_process_exited() {
     try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system
 
     # Verify the discovery ConfigMap contains notary controller
-    run -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.enabledControllers}'
-    assert_output --partial "notary"
+    run -0 --separate-stderr rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.rdd}'
+    run -0 jq_output '.enabledControllers[]'
+    assert_line "notary"
 }
 
 @test "webhook configuration is created" {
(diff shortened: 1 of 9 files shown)


pkg/service/controllers/discovery.go
```diff
@@ -58,136 +65,169 @@ func (d *ControllerManagerDiscovery) RegisterControllerManager(ctx context.Conte
...
+		// Update existing ConfigMap as needed
+		configMap.Data[d.name] = string(serializedInfo)
+		_, err = d.client.CoreV1().ConfigMaps(d.namespace).Update(ctx, configMap, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to update existing controller manager configmap: %w", err)
+		}
+		return nil
+	} else if !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to get existing controller manager configmap: %w", err)
+	}
```
jandubois: There are 2 potential race conditions:

1. Two external controllers both try to create the ConfigMap at the same time. One creation will fail, and then it needs to retry updating the ConfigMap created by the other.

    In practice this will probably not happen because the ConfigMap should be created by the `rdd` main controller before the other controllers get started.

2. Two controllers may try to register simultaneously, but the Get/Update operation is not atomic. It would be better to do a StrategicMergePatch or JSONPatch. Or use server-side Apply.

The same issue also applies to the Unregister case.

jandubois (changes requested):
Thanks, looks mostly fine to me (and does what I intended it to do to begin with, but messed up implementing).

There is a race condition though that should be addressed.

pkg/service/controllers/discovery.go
jandubois: I just realized that for (2) you could also just retry the Get/Update operation when it fails because the ResourceVersion does not match. Maybe that is simpler? But I think a Patch or server-side update is more idiomatic.

mook-as:
Added fixes for the race condition, and a BATS test to make sure it actually works correctly.  This also means adding a40950e07a6b1e2a73d69a79469ef0972bf4954e to fix an issue where the external controllers end up trying to use the same port as the internal controllers (even though it's in use), and failing to start.

bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -38,8 +38,9 @@ assert_process_exited() {
     try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system
 
     # Verify the discovery ConfigMap contains notary controller
-    run -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.enabledControllers}'
-    assert_output --partial "notary"
+    run -0 --separate-stderr rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.rdd}'
```
jandubois:
```suggestion
    run_e -0 rdd ctl get configmap rdd-controller-manager --namespace rdd-system -o jsonpath='{.data.rdd}'
```

bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -133,3 +134,57 @@ EOF
...
+
+@test "shut down external controller" {
+    # Get the stored PID from the controller start test
+    controller_pid=$(cat "${BATS_FILE_TMPDIR}/controller_pid")
+
+    # Verify the external controller process is currently running
+    run -0 kill -0 "${controller_pid}"
+
+    # Stop the control plane - this should trigger auto-cleanup of external controllers
+    run -0 rdd svc stop
```
jandubois: `run -0` is redundant unless you look at `$output` afterwards.

```suggestion
    kill -0 "${controller_pid}"

    # Stop the control plane - this should trigger auto-cleanup of external controllers
    rdd svc stop
```

pkg/external/main.go
```diff
@@ -122,8 +126,8 @@ func RunControllers(apiGroupName string) int {
 
 // monitorControlPlane monitors the control plane lifecycle and cancels the context when it's no longer available.
 // This allows external controllers to automatically exit when `rdd svc stop` or `rdd svc delete` is called.
-func monitorControlPlane(ctx context.Context, config *rest.Config, log klog.Logger, cancel context.CancelFunc) {
-	discovery, err := controllers.NewControllerManagerDiscovery(config)
+func monitorControlPlane(ctx context.Context, name string, config *rest.Config, log klog.Logger, cancel context.CancelFunc) {
```
jandubois: I was wondering what exactly `name` is supposed to be. Should we call it `apiGroupName`? Because I don't think it can be anything else; we have one shared controller manager per API group.

That means the `NewSharedControllerManager()` parameter should also be `apiGroupName`.

pkg/external/main.go
```diff
@@ -194,7 +198,7 @@ func monitorControlPlane(ctx context.Context, config *rest.Config, log klog.Logg
 // shouldStartController checks if an external controller should start based on RDD discovery.
 // Returns true if no RDD controller manager is running or the specific controller is not enabled.
 func shouldStartController(ctx context.Context, config *rest.Config, controllerName string, log klog.Logger) (bool, error) {
-	discovery, err := controllers.NewControllerManagerDiscovery(config)
+	discovery, err := controllers.NewControllerManagerDiscovery(config, "")
```
jandubois: I think we should either have a named constant for the special `""` API group name to signify the builtin controller manager, or use a different function name to create a builtin ControllerManagerDiscovery object.

pkg/service/cmd/service.go
```diff
@@ -267,38 +256,36 @@ func checkReadiness(ctx context.Context) error {
...
-			klog.V(2).Infof("getRuntimeControllersFromCluster: %v", err)
+			klog.V(2).Info("Controllers expected but discovery configmap not found yet - waiting for external controllers to register...")
 		}
-		return err
+		return errors.New("waiting for controller manager registration")
 	}
+
 	klog.V(2).Infof("Waiting for %d runtime controllers to be ready", len(runtimeControllers))
 	if len(runtimeControllers) == 0 {
-		// This shouldn't happen since we handle it above, but keeping as fallback
```
jandubois, line 301: This doesn't make sense, this literally cannot happen with the logic above; why don't we delete this block?

```go
if len(runtimeControllers) == 0 {
    ...
    return ...
}
klog()
if len(runtimeControllers) == 0 {
    // this is unreachable
}
```

pkg/service/controllers/discovery.go
```diff
@@ -58,136 +66,211 @@ func (d *ControllerManagerDiscovery) RegisterControllerManager(ctx context.Conte
...
-	if errors.IsNotFound(err) {
-		_, err = d.client.CoreV1().ConfigMaps(d.namespace).Create(ctx, configMap, metav1.CreateOptions{})
-	}
-
+	_, err = d.client.CoreV1().ConfigMaps(d.namespace).Create(ctx, configMap, metav1.CreateOptions{})
 	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			// We hit a race condition, and somebody else created it first.  Try again.
+			// We're not expecting to recurse more than once, because if the config map
+			// exists then we won't hit IsNotFound on the patch attempt.
```
jandubois, line 157: I understand this should not happen, but would still feel better about passing a `retry: false` flag to avoid recursion if there is some other bug in the logic.

pkg/service/controllers/discovery.go
```diff
@@ -58,136 +66,211 @@ func (d *ControllerManagerDiscovery) RegisterControllerManager(ctx context.Conte
...
 		"controllers", len(controllers))
 
 	return nil
 }
 
 // UnregisterControllerManager removes the controller manager ConfigMap.
 func (d *ControllerManagerDiscovery) UnregisterControllerManager(ctx context.Context) error {
-	err := d.client.CoreV1().ConfigMaps(d.namespace).Delete(ctx, ControllerManagerConfigMapName, metav1.DeleteOptions{})
+	patchData, err := json.Marshal(map[string]any{
+		d.name: nil,
```
jandubois: I think this needs to be nested inside `data`:

```suggestion
		"data": map[string]string{
			d.name: nil,
		},
```

jandubois (changes requested):
Mostly small things, but I think the patch for deletion is wrong and doesn't work.

bats/tests/31-rdd-controllers/notary-external.bats
mook-as: I'll do this, though I do want to point out that `run_e` not exactly a very clear name.

pkg/external/main.go
mook-as: This isn't actually using the builtin manager; instead, it's just that the name is irrelevant here.  But that's confusing, so I'll do some refactoring so we don't need to provide an empty name.

Essentially, the problem is that `ControllerManagerDiscovery` is an amalgamation of two things:
- A thing to look up the discovery info
- A thing to also _update_ the discovery info.

mook-as (commented):
Comments addressed (I think), and added a test to make sure I don't re-break the unregister bit.

bats/tests/31-rdd-controllers/notary-external.bats
jandubois: > I'll do this, though I do want to point out that `run_e` not exactly a very clear name.

I agree, and better suggestions welcome. I just want the test scaffolding to fade into the background, and the lines should be mostly actual test logic. So `run --separate-stderr` was just too long of a prefix for me; it takes my eyes too long to see where the actual command starts.

I think I've intermittently also used `run -0 --separate-stderr -- actual command here` to use `--` to give an extra hint on where the command itself begins, but I still prefer the shorter `run_e` prefix. But we can discuss this some more if it doesn't work for you.

jandubois (approved):
Thanks, LGTM

jandubois (changes requested):
So `pkg/service/controllers/discovery_test.go‎` still needs copyright headers.

And I don't know why BATS failed in CI (I will just rerun it for now, as it passed for me locally).

jandubois (approved):
Thanks, LGTM

PR #106, opened 2026-01-02
Increase unit test timeouts for CI

jandubois:
Saw multiple failures to clean up after tests:

```
--- FAIL: TestDeleteOwnedResources (25.88s)
    finalizer_test.go:37: assertion failed: error is not nil: timeout waiting for process kube-apiserver to stop: failed to stop environment
FAIL
```

Only on CI; could not repro locally

### DIFF
--- a/pkg/controllers/base/finalizer_test.go
+++ b/pkg/controllers/base/finalizer_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"runtime"
 	"testing"
+	"time"
 
 	"gotest.tools/v3/assert"
 	"gotest.tools/v3/assert/cmp"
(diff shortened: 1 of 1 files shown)


pkg/controllers/base/finalizer_test.go
```diff
@@ -24,7 +25,8 @@ import (
 
 func TestDeleteOwnedResources(t *testing.T) {
 	env := &envtest.Environment{
-		DownloadBinaryAssets: true,
+		DownloadBinaryAssets:    true,
+		ControlPlaneStopTimeout: 60 * time.Second,
```
mook-as: Why not use `time.Minute` here? That's existed since `time.Duration` was a thing.

mook-as (dismissed):

mook-as (approved):

PR #107, opened 2026-01-05
controllers: tests: work around CI failures

mook-as:
In CI, we seem to be unable to stop kube-apiserver sometimes (for the test) but this cannot be reproduced locally.  Since it's just cleanup anyway, just ignore the specific error in CI.

### DIFF
--- a/pkg/controllers/base/finalizer_test.go
+++ b/pkg/controllers/base/finalizer_test.go
@@ -5,9 +5,10 @@ package base
 
 import (
 	"fmt"
+	"os"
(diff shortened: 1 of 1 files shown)


jandubois (dismissed):
Thanks, LGTM

Should we revert the extended timeout setting then, since we are ignoring the error in CI anyways? Makes the test finish 40 seconds earlier, and gives us a warning if the cleanup takes longer locally as well.

jandubois (approved):

PR #110, opened 2026-01-06
API: Add notes on API to manipulate containers &c

mook-as:
Jan brings up a good point that we don't have a single source of truth, so it's not clear if a containerd-side resource without a Kubernetes-side resource would mean the containerd-side needs to be deleted or the Kubernetes-side one needs to be created.

### DIFF
--- a/docs/design/api_containers.md
+++ b/docs/design/api_containers.md
@@ -16,11 +16,14 @@ When running `containerd`, the containerd namespace is listed as the `namespace`
 label rather than re-using the Kubernetes namespace.  When running `dockerd`,
 namespaces are not supported and we always use `moby` as the value for that label.
 
-For the `*Request` resources, once they have successfully completed, they will
-be removed after some timeout; this will be at least one minute (so the caller
-can read any data).
+For the `*Request` resources, they use the `Complete` and `Failed` conditions to
+express state; those are mutually exclusive (only one of the two can be set to
+`True` at once).  Once either is set to `True`, the request object is considered
+to be in a terminal state and will be removed after some timeout.  This will be
+at least one minute (so the caller can read any data), but the precise timing is
+unspecified.
 
-This is mainly for use by the Rancher Desktop front end; all other users are
+This API is mainly for use by the Rancher Desktop front end; all other users are
 strongly urged to use the relevant CLI or other API instead.
 
 ## Containers
@@ -110,8 +113,10 @@ status:
   # same Kubernetes namespace as the ContainerCreateRequest.
   name: 8eb6f2cf72b6616aa743cf9187f350af84c9749dab65474db2530f26745d2ef3
   conditions:
-  - type: Finished
+  - type: Complete
     status: True
+  - type: Failed
+    status: False
 ```
 
 If `.metadata.labels.namespace` / `.metadata.labels.name` duplicates an existing
@@ -140,7 +145,9 @@ at which point the `Container` object will actually be deleted.
 
 `Image` objects reflect images in the container engine.  Each tag is represented
 as a new `Image` object; therefore, there may be multiple `Image` objects for
-the same image ID (one per tag).
+the same image ID (one per tag).  If an image without any tags exists, that will
+be represented by an `Image` object without `.status.repoTag` and
+`.metadata.labels.namespace`.
 
 ```yaml
 apiVersion: containers.rancherdesktop.io/v1alpha1
@@ -164,14 +171,12 @@ status:
   size: 45150437
   labels:
     org.opensuse.base.vendor: openSUSE Project
-  conditions:
-  - type: Pushed # Used to record last push time.
-    status: False
+  conditions: []
 ```
 
 ### Image Actions
 
-#### Fetch image
+#### Pull image
 Create an `ImagePullRequest` object:
 ```yaml
 apiVersion: containers.rancherdesktop.io/v1alpha1
@@ -182,7 +187,9 @@ spec:
   repoTag: 'registry.opensuse.org/opensuse/leap:latest'
 status:
   conditions:
-  - type: Pulled
+  - type: Complete
+    status: True
+  - type: Failed
     status: False
 ```
 
@@ -205,7 +212,9 @@ spec:
   imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345'
 status:
   conditions:
-  - type: Finished
+  - type: Complete
+    status: True
+  - type: Failed
     status: False
 ```
 
@@ -222,8 +231,10 @@ spec:
   imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345'
 status:
   conditions:
-  - type: Finished
+  - type: Complete
     status: True
+  - type: Failed
+    status: False
   result:
     # Just dump the raw Trivy result JSON here (without converting to YAML).
     '{ ... }'
@@ -279,7 +290,9 @@ spec:
   driver: local
 status:
   conditions:
-  - type: Processed
+  - type: Complete
+    status: True
+  - type: Failed
     status: False
 ```
 Only local volumes are supported initially.
--- a/docs/design/api_containers.md
+++ b/docs/design/api_containers.md
@@ -43,6 +43,7 @@ spec:
         hostPort: 32768
   labels:
     org.opensuse.base.vendor: openSUSE Project
+  state: Running # Desired state
 status:
   status: Running
   pid: 5059
@@ -64,21 +65,45 @@ status:
     status: False
 ```
 
+If a `Container` object exists that does not actually exist in the container
+engine, it is automatically deleted.  Therefore, creating `Container` objects
+directly is not effective; to create containers, use `ContainerRequest`.
+
 ### Container actions
 
 We will need to support a variety of actions on containers:
 
 #### Create container
-Somehow we don't actually need this in the front end.  This will not be
-supported initially.
 
-If `.metadata.name` is the container ID (which we can't generate ahead of time),
-we can't actually create containers (because we don't know their name).
+To create a container, create a `ContainerRequest` object:
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: ContainerRequest
+metadata:
+  generateName: whatever
+  namespace: rdd-system
+  labels:
+    name: magical_gates
+    namespace: k8s.io # containerd namespace
+spec:
+  # See Container.spec
+status:
+  # Resulting .metadata.name, which is the container ID.  It must be in the
+  # same Kubernetes namespace as the ContainerRequest.
+  name: 8eb6f2cf72b6616aa743cf9187f350af84c9749dab65474db2530f26745d2ef3
+  conditions:
+  - type: Finished
+    status: True
+```
+
+The `ContainerRequest` will be deleted soon after the `Finished` condition
+transitions to `True`.
+
+If multiple `ContainerRequest` objects exist at the same time for the same
+contained `name`/`namespace` pair, the result is undefined.
 
 #### Change container state
-Set an annotation? `containers.rancherdesktop.io/desiredStatus` perhaps?
-Alternatively, add a `.spec.state` and make that the desired state, which would
-match Kubernetes APIs better.
+Set `.spec.state` to match how Kubernetes resources normally work.
 
 #### Fetch container logs
 The `@kubernetes/client-node` package has some hand-written code to deal with
@@ -89,7 +114,12 @@ Same as logs; there's some special code in `@kubernetes/client-node` that we may
 be able to fork.
 
 #### Delete container
-Delete the Kubernetes API object.
+Create a `ContainerRequest` with `.spec.state` set to `deleted`.
+We can't just delete the Kubernetes object, because it will be recreated as a
+reflection of the containerd object.
+
+Once the container has been removed from the container engine, the Kubernetes
+object will be removed via normal reflection.
 
 ## Images
 
@@ -126,6 +156,7 @@ metadata:
 spec:
   # refers to `Image` objects, which are not namespaced.
   imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b'
+  push: false # If set to true, the image is pushed; see below.
 status:
   conditions:
   - type: PullStarted
@@ -151,15 +182,15 @@ Not sure; do something with the `Resource` API maybe?
 We may need an `ImageBuildRequest` job-thing or something?
 
 #### Push image
-Annotation? Spec?
-After push, probably set the `Pushed` condition to `True` (which will
-automatically get a time stamp).
+Set `.spec.push` to `true`.  The image will be pushed by `.metadata.labels.name`.
+After the push, `.status[?(@.type=="Pushed")]` will be set to `True`, and
+`.spec.push` will be set to `false`.
 
 #### Scan image
 We will need a new object type for this; maybe something like
 ```yaml
 apiVersion: containers.rancherdesktop.io/v1alpha1
-kind: ImageScan
+kind: ImageScanRequest
 metadata:
   generateName: imageScan-
   namespace: rdd-system # not containerd namespace
@@ -174,8 +205,9 @@ status:
     # Just dump the Trivy result JSON here (without converting to YAML).
 ```
 
-#### Delete image
-Delete the Kubernetes API object.
+#### Untag image
+Update the `ImageTag` with the `.spec.imageRef` set to `deleted`; once processed,
+the `ImageTag` will be removed to reflect the container engine state.
 
 ## Volumes
 
@@ -200,7 +232,21 @@ spec:
 ### Volume Actions
 
 #### Create volume
-Create a volume with a name.  To start with, only local volumes are supported.
+Create a `VolumeRequest`:
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: Volume
+metadata: # same as Volume.metadata
+spec: # same as Volume.spec
+status:
+  conditions:
+  - type: Processed
+    status: False
+```
+Only local volumes are supported initially.
+Parts of `.spec` may be ignored.
 
 #### Delete volume
-Delete the Kubernetes object.
+Create a `VolumeRequest`, with `.spec.scope` set to `deleted`.
+Once the volume has been deleted, the `VolumeRequest` will also be deleted after
+a brief timeout.
--- a/docs/design/api_containers.md
+++ b/docs/design/api_containers.md
@@ -64,6 +64,33 @@ status:
     status: False
 ```
 
+### Container actions
+
+We will need to support a variety of actions on containers:
+
+#### Create container
+Somehow we don't actually need this in the front end.  This will not be
+supported initially.
+
+If `.metadata.name` is the container ID (which we can't generate ahead of time),
+we can't actually create containers (because we don't know their name).
+
+#### Change container state
+Set an annotation? `containers.rancherdesktop.io/desiredStatus` perhaps?
+Alternatively, add a `.spec.state` and make that the desired state, which would
+match Kubernetes APIs better.
+
+#### Fetch container logs
+The `@kubernetes/client-node` package has some hand-written code to deal with
+logs; maybe we can make a copy of that (with a different endpoint) for this?
+
+#### Exec (shell) in container
+Same as logs; there's some special code in `@kubernetes/client-node` that we may
+be able to fork.
+
+#### Delete container
+Delete the Kubernetes API object.
+
 ## Images
 
 `Image` objects reflect images in the container engine.
@@ -72,13 +99,9 @@ status:
 apiVersion: containers.rancherdesktop.io/v1alpha1
 kind: Image
 metadata:
-  name: 'sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # Image ID
-  namespace: default # Not related to containerd namespace
-  labels:
-    namespace: k8s.io # containerd namespace, or `moby` if using dockerd
+  name: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # Image ID, colon replaced with dot.
+  namespace: rdd-system # not the containerd namespace
 spec:
-  repoTags:
-  - registry.opensuse.org/opensuse/leap:latest
   repoDigests:
   - registry.opensuse.org/opensuse/leap@sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b
   createdAt: "2025-11-17T03:14:16Z"
@@ -89,15 +112,81 @@ spec:
     org.opensuse.base.vendor: openSUSE Project
 ```
 
+`ImageTag` objects are names for each `Image`.
+
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: ImageTag
+metadata:
+  generateName: registry.opensuse.org-opensuse-leap-latest-
+  namespace: rdd-system
+  labels:
+    name: 'registry.opensuse.org/opensuse/leap:latest'
+    namespace: moby # containerd namespace
+spec:
+  # refers to `Image` objects, which are not namespaced.
+  imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b'
+status:
+  conditions:
+  - type: PullStarted
+    status: True
+  - type: Ready
+    status: True
+  - type: Pushed
+    status: False
+```
+
+This is split into separate objects so that listing images by tag is easier,
+assuming we never want to list images with no tags.
+
+### Image Actions
+
+#### Fetch image
+Create an `ImageTag` object, but omit the `imageRef`.  The reconciler will pull
+the corresponding image and fill in the `imageRef` once available.
+
+#### Build image
+Not sure; do something with the `Resource` API maybe?
+
+We may need an `ImageBuildRequest` job-thing or something?
+
+#### Push image
+Annotation? Spec?
+After push, probably set the `Pushed` condition to `True` (which will
+automatically get a time stamp).
+
+#### Scan image
+We will need a new object type for this; maybe something like
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: ImageScan
+metadata:
+  generateName: imageScan-
+  namespace: rdd-system # not containerd namespace
+spec:
+  # refers to `Image` objects, which are not namespaced.
+  imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b'
+status:
+  conditions:
+  - type: Finished
+    status: True
+  result:
+    # Just dump the Trivy result JSON here (without converting to YAML).
+```
+
+#### Delete image
+Delete the Kubernetes API object.
+
 ## Volumes
 
 ```yaml
 apiVersion: containers.rancherdesktop.io/v1alpha1
 kind: Volume
 metadata:
-  name: volume-name
+  generateName: volume-name- # Based on the containerd name
   namespace: default # Not related to containerd namespace
   labels:
+    name: volume-name
     namespace: k8s.io # containerd namespace, or `moby` if using dockerd
 spec:
   createdAt: "2025-11-17T03:14:16Z"
@@ -107,3 +196,11 @@ spec:
   scope: local
   options: {}
 ```
+
+### Volume Actions
+
+#### Create volume
+Create a volume with a name.  To start with, only local volumes are supported.
+
+#### Delete volume
+Delete the Kubernetes object.
--- a/docs/design/api_containers.md
+++ b/docs/design/api_containers.md
@@ -10,9 +10,15 @@ be rejected by the controllers.
 
 All objects are in the `containers.rancherdesktop.io` API group.
 
+All times are in RFC3339 format, per usual Kubernetes conventions.
+
 When running `containerd`, the containerd namespace is listed as the `namespace`
 label rather than re-using the Kubernetes namespace.  When running `dockerd`,
 namespaces are not supported and we always use `moby` as the value for that label.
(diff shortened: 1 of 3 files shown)


jandubois:
### Feedback on the document, not limited to what is in this PR

#### Namespace should be `rancher-desktop` not `rdd-system`

See api_app.md

#### Delete containers with a finalizer

Creating a `ContainerRequest` resource to delete a `Container` resource feels too complicated. We can delete the actual container in the finalizer when a `Container` resource is deleted. That way it isn't being recreated.

#### `ContainerRequest` should support modifying containers

Well, modifying is not really a thing; we have to delete the old container, and create a new one with the modified settings (which will have a different container id and therefore a different resource name). To avoid port conflicts the old one needs to be removed (including port forwards) before the new one can be created. Moving this functionality into the controller avoids having to track state in the frontend.

I think we should have a `replaceExisting: BOOL` setting (defaults `false`). If it is set to true, and a container with the specified namespace/name already exists, it will be deleted before the new container is created.

#### What is `ContainerRequest.metadata.generateName`?

I know we talked about the `Container` resource names before, but I just realized they should always match the container id.

#### `Container.spec` should be moved to `.status`, except for `.spec.state`

The `.spec` is for desired state, and only `.spec.state` can be modified to specify the desired state. Everything else is reflected actual state of the running container and should be in `.status` instead.

Same thing for `Image.spec`.

#### Having both `Image` and `ImageTag` resources feels too complicated.

It means you have to perform a "table join" to actually get the list of tagged images with metadata.

I think it would be easier to have a separate `Image` object for each tag, even if this duplicates some metadata. Makes them easy to delete using the same finalizer mechanism proposed for `Container`. May need some additional work for "dangling images" without tags, but the current design has the same issue. This also closely matches how `docker images` works.

I wonder if we should have an `ImageRequest` resource similar to `ContainerRequest` to create images by pulling/building (and maybe also scanning of existing images).

mook-as:
> ### Feedback on the document, not limited to what is in this PR
> #### Namespace should be `rancher-desktop` not `rdd-system`
> 
> See api_app.md

Ah, okay, will change that.

> #### Delete containers with a finalizer
> 
> Creating a `ContainerRequest` resource to delete a `Container` resource feels too complicated. We can delete the actual container in the finalizer when a `Container` resource is deleted. That way it isn't being recreated.

Ah, I'll redo things to assume we can cause deletion by deleting objects, then; that makes the API nicer.

> #### `ContainerRequest` should support modifying containers
> 
> Well, modifying is not really a thing; we have to delete the old container, and create a new one with the modified settings (which will have a different container id and therefore a different resource name). To avoid port conflicts the old one needs to be removed (including port forwards) before the new one can be created. Moving this functionality into the controller avoids having to track state in the frontend.
> 
> I think we should have a `replaceExisting: BOOL` setting (defaults `false`). If it is set to true, and a container with the specified namespace/name already exists, it will be deleted before the new container is created.

Given how both container engines work, wouldn't it be better to make the user do that in two steps? Since it's immutable, trying wrap would just lead to a leaky abstraction (any temporary files would be deleted, etc.); I think would be better just avoid the issue by letting the user know what is actually happening.

> #### What is `ContainerRequest.metadata.generateName`?

See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names about `generateName`.  It's just server-side generation of a unique name.
 
> I know we talked about the `Container` resource names before, but I just realized they should always match the container id.

That's how it's specced right now.
(`ContainerRequest` names can't match the container ID, because we don't know what it is until the container has been created.)

> #### `Container.spec` should be moved to `.status`, except for `.spec.state`
> 
> The `.spec` is for desired state, and only `.spec.state` can be modified to specify the desired state. Everything else is reflected actual state of the running container and should be in `.status` instead.

When creating a container, `args`, `image`, etc. should be spec, too right?
That is, `ContainerRequest` should use `.spec`, but `Container` should use `.status`?

> 
> Same thing for `Image.spec`.
> #### Having both `Image` and `ImageTag` resources feels too complicated.
> 
> It means you have to perform a "table join" to actually get the list of tagged images with metadata.
> 
> I think it would be easier to have a separate `Image` object for each tag, even if this duplicates some metadata. Makes them easy to delete using the same finalizer mechanism proposed for `Container`. May need some additional work for "dangling images" without tags, but the current design has the same issue. This also closely matches how `docker images` works.

Internally, images have a list of tags; if we do that instead (that is, make it `.spec.repoTags[]`), would that be better? I think the tags would end up being namespaced, though, so it'd be a bit odd.

The current (77a8d8d1df7d60d2ff990f9f6e133404df1aad14) design would allow for dangling images by having `Image` without corresponding `ImageTag` objects.  My new proposal would just mean `Image` with empty `.spec.repoTags` arrays.

> 
> I wonder if we should have an `ImageRequest` resource similar to `ContainerRequest` to create images by pulling/building (and maybe also scanning of existing images).

I'll see what that would look like.  Conflating creation and scanning seems kind of odd, though (since, theoretically, we could use a different controller for scanning vs image management).

jandubois:
> > #### `ContainerRequest` should support modifying containers
> > I think we should have a `replaceExisting: BOOL` setting (defaults `false`). If it is set to true, and a container with the specified namespace/name already exists, it will be deleted before the new container is created.
> 
> Given how both container engines work, wouldn't it be better to make the user do that in two steps? Since it's immutable, trying wrap would just lead to a leaky abstraction (any temporary files would be deleted, etc.); I think would be better just avoid the issue by letting the user know what is actually happening.

I'm not sure who/what "the user" is going to be here. My scenario is:

* we have some kind of container dashboard where we then can drill down into container details
* the user can edit container details, like modifying environment variables, port mappings, volume mounts, etc
* when the user selects "Apply" the old container must be deleted and a new container will be created with the modified settings

I assumed it would be easier for the UI to simply send a `ContainerRequest` to create the new container with `replaceExisting: true`. Otherwise it will have to maintain its own little state machine to track the deleting of the old container, and only create the new one when the deletion is complete.

If you think it is better for the UI to do this explicitly, then we don't need the replace mechanism.
 
> > #### What is `ContainerRequest.metadata.generateName`?
> 
> See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names about `generateName`. It's just server-side generation of a unique name.

I'm not sure how this is relevant for us. The user can use whatever name they want for the `ContainerRequest`; it is not actually used/needed for anything.

Only the controller can create `Container` resources, and they will always be created with the container id as the name.

> > #### `Container.spec` should be moved to `.status`, except for `.spec.state`
> When creating a container, `args`, `image`, etc. should be spec, too right? That is, `ContainerRequest` should use `.spec`, but `Container` should use `.status`?

Yes, exactly.
 
> > #### Having both `Image` and `ImageTag` resources feels too complicated.
> Internally, images have a list of tags; if we do that instead (that is, make it `.spec.repoTags[]`), would that be better? I think the tags would end up being namespaced, though, so it'd be a bit odd.

I think having the image tag as a metadata label will make it easier to select a specific tag, but haven't thought it through.
 
> The current ([77a8d8d](https://github.com/rancher-sandbox/rancher-desktop-daemon/commit/77a8d8d1df7d60d2ff990f9f6e133404df1aad14)) design would allow for dangling images by having `Image` without corresponding `ImageTag` objects. My new proposal would just mean `Image` with empty `.spec.repoTags` arrays.

We could use `<none>:<none>` as the name, just like docker?

> > I wonder if we should have an `ImageRequest` resource similar to `ContainerRequest` to create images by pulling/building (and maybe also scanning of existing images).
> 
> I'll see what that would look like. Conflating creation and scanning seems kind of odd, though (since, theoretically, we could use a different controller for scanning vs image management).

I agree, but then we should have more specific names, like `ImageCreateRequest` and `ImageScanRequest`. And for consistency also use `ContainerCreateRequest`.

jandubois:
> If you think it is better for the UI to do this explicitly, then we don't need the replace mechanism.

Given that we currently don't have container editing capabilities, we can leave this unimplemented until we actually need it. I would still keep a reference to the "idea" though, so we can consider it once we actually may need it.

mook-as:
> > > #### `ContainerRequest` should support modifying containers
> > > I think we should have a `replaceExisting: BOOL` setting (defaults `false`). If it is set to true, and a container with the specified namespace/name already exists, it will be deleted before the new container is created.
> > 
> > 
> > Given how both container engines work, wouldn't it be better to make the user do that in two steps? Since it's immutable, trying wrap would just lead to a leaky abstraction (any temporary files would be deleted, etc.); I think would be better just avoid the issue by letting the user know what is actually happening.
> 
> I'm not sure who/what "the user" is going to be here.

I meant the physical person at the computer poking at the mouse.

> My scenario is:
> 
>     * we have some kind of container dashboard where we then can drill down into container details
> 
>     * the user can edit container details, like modifying environment variables, port mappings, volume mounts, etc
> 
>     * when the user selects "Apply" the old container must be deleted and a new container will be created with the modified settings

I don't think we should support editing containers in-place, because that isn't actually possible, and it would have unexpected side effects (temporary files deleted, in-memory things going away, maybe regenerate certificates or something).  I think it would be better to actually expose it as separate delete and create operations instead.

> 
> 
> I assumed it would be easier for the UI to simply send a `ContainerRequest` to create the new container with `replaceExisting: true`. Otherwise it will have to maintain its own little state machine to track the deleting of the old container, and only create the new one when the deletion is complete.
> 
> If you think it is better for the UI to do this explicitly, then we don't need the replace mechanism.

Yeah, I think the person at the computer would need to do the delete + create explicitly, because we can't do the edit safely.
> 
> > > #### What is `ContainerRequest.metadata.generateName`?
> > 
> > 
> > See https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names about `generateName`. It's just server-side generation of a unique name.
> 
> I'm not sure how this is relevant for us. The user can use whatever name they want for the `ContainerRequest`; it is not actually used/needed for anything.
> 
> Only the controller can create `Container` resources, and they will always be created with the container id as the name.

I'm wearing frontend-tinted glasses when writing the API; in effect we'll end up with "some name" and nobody really cares, _specifically_, what that is.

> 
> > > #### `Container.spec` should be moved to `.status`, except for `.spec.state`
> > 
> > 
> > When creating a container, `args`, `image`, etc. should be spec, too right? That is, `ContainerRequest` should use `.spec`, but `Container` should use `.status`?
> 
> Yes, exactly.

🫡 

> 
> > > #### Having both `Image` and `ImageTag` resources feels too complicated.
> > 
> > 
> > Internally, images have a list of tags; if we do that instead (that is, make it `.spec.repoTags[]`), would that be better? I think the tags would end up being namespaced, though, so it'd be a bit odd.
> 
> I think having the image tag as a metadata label will make it easier to select a specific tag, but haven't thought it through.

True. I'm just uneasy about the duplication, because the backend will then need to fetch the list of images from docker/containerd, then fetch _all_ images from the Kube API, and then figure out the difference.

> 
> > The current ([77a8d8d](https://github.com/rancher-sandbox/rancher-desktop-daemon/commit/77a8d8d1df7d60d2ff990f9f6e133404df1aad14)) design would allow for dangling images by having `Image` without corresponding `ImageTag` objects. My new proposal would just mean `Image` with empty `.spec.repoTags` arrays.
> 
> We could use `<none>:<none>` as the name, just like docker?

That's a _display_ option, not an _API_ option, in docker; though I don't think anything _prevents_ us from doing that anyway…
> 
> > > I wonder if we should have an `ImageRequest` resource similar to `ContainerRequest` to create images by pulling/building (and maybe also scanning of existing images).
> > 
> > 
> > I'll see what that would look like. Conflating creation and scanning seems kind of odd, though (since, theoretically, we could use a different controller for scanning vs image management).
> 
> I agree, but then we should have more specific names, like `ImageCreateRequest` and `ImageScanRequest`. And for consistency also use `ContainerCreateRequest`.

Yeah, now that we will not have delete requests, `ImageCreateRequest` etc. seems better.

jandubois:
> > > > #### `ContainerRequest` should support modifying containers
> 
> I don't think we should support editing containers in-place, because that isn't actually possible, and it would have unexpected side effects (temporary files deleted, in-memory things going away, maybe regenerate certificates or something). I think it would be better to actually expose it as separate delete and create operations instead.

This is easy enough to make safe by only allowing editing of stopped containers. But then the `ContainerRequest` could explicitly stop an existing container before deleting it.
 
> Yeah, I think the person at the computer would need to do the delete + create explicitly, because we can't do the edit safely.

That is a horrible UX and essentially amounts to saying: "Use the CLI with docker-compose" to manage containers.

At a minimum we would need a mechanism to duplicate an existing container's settings, and allow editing them before creating the new one in stopped state. Then the user would have to delete the old container manually. And then they have to duplicate it once more to rename it back to the old container name, and then delete the temporary duplicate. Finally they can start the modified container to replace the deleted one. Much more complex and error-prone than pretending you just edit the existing instance by removing the old instance automatically before recreating it.

Remember this is a development tool that should make it easy to try things, not something to manage a production server.

But we can discuss this later and leave out the `removeExisting` from the spec. I also just realized that it does not allow for renaming a container either, so if we want to support the edit operation, we need to specify the old name and not just a boolean.

> > > > #### What is `ContainerRequest.metadata.generateName`?
> I'm wearing frontend-tinted glasses when writing the API; in effect we'll end up with "some name" and nobody really cares, _specifically_, what that is.

Ah yes, using `generateName` will be useful for this. But this is then not really part of the `Container` resource spec, but a general feature of the control plane that works for any object.
 
> > > > #### Having both `Image` and `ImageTag` resources feels too complicated.
> > I think having the image tag as a metadata label will make it easier to select a specific tag, but haven't thought it through.
> 
> True. I'm just uneasy about the duplication, because the backend will then need to fetch the list of images from docker/containerd, then fetch _all_ images from the Kube API, and then figure out the difference.

Fetching data from the Kube API is really just a SQLite query. It will be fast, and the amount of data per image doesn't really matter much because there will be a limited number of images in the VM.

mook-as:
Updated to use one `Image` object per tag; we can still have `Image` objects with no labels for untagged images, I think?

Also removed the `generateName` stuff, because you're right we don't need that in the API docs examples.

I think that covers everything?

docs/design/api_containers.md
```diff
@@ -64,40 +65,159 @@ status:
...
+  namespace: rancher-desktop # not the containerd namespace
   labels:
-    namespace: k8s.io # containerd namespace, or `moby` if using dockerd
+    # If an image is untagged, `name` and `namespace` are missing.
+    name: 'registry.opensuse.org/opensuse/leap:latest' # image tag
+    namespace: moby # containerd namespace
 spec:
-  repoTags:
-  - registry.opensuse.org/opensuse/leap:latest
-  repoDigests:
```
mook-as, line 82: I seem to have lost these fields; they should be moved to `status`, I think?

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
 kind: Image
 metadata:
-  name: 'sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # Image ID
-  namespace: default # Not related to containerd namespace
+  name: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345' # Image ID, colon replaced with dot, with random suffix.
+  namespace: rancher-desktop # not the containerd namespace
   labels:
-    namespace: k8s.io # containerd namespace, or `moby` if using dockerd
+    # If an image is untagged, `name` and `namespace` are missing.
+    name: 'registry.opensuse.org/opensuse/leap:latest' # image tag
```
mook-as: This turns out to be invalid, because there are restrictions on the values:

> failed to create static image : Image.containers.rancherdesktop.io "ghcr.io-rancher-sandbox-rancher-desktop-rdx-proxy.latest-7t9v9" is invalid: metadata.labels: Invalid value: "ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest": a valid label must be an empty string or consist of alphanumeric characters, '-', '\_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

I can dump it in a `.status` field, but that means we won't be select by them, which seems kind of terrible.

See [documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

mook-as: Maybe we can use a [field selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#custom-resources-fields) for that?

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
+  # Resulting .metadata.name, which is the container ID.  It must be in the
+  # same Kubernetes namespace as the ContainerCreateRequest.
+  name: 8eb6f2cf72b6616aa743cf9187f350af84c9749dab65474db2530f26745d2ef3
+  conditions:
+  - type: Finished
+    status: True
+```
+
+The `ContainerCreateRequest` will be deleted soon after the `Finished` condition
+transitions to `True`.
```
jandubois: Maybe we should define what "soon" means. Could it be deleted within a second of the request being "Finished"? Or is there some guaranteed time it will be available?

Job objects have a TTL controller, and you can set `ttlSecondsAfterFinished` to specify when they should be deleted (by default they are not deleted at all).

We could implement something similar, maybe even a generic controller that can work on any kind of CRD if the resource has a corresponding cleanup label?

May be overkill initially, but if we already implement deleting "soon" then we are already most of the way there.

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
+
+If `.metadata.labels.namespace` / `.metadata.labels.name` duplicates an existing
+container, an `CreateFailed` status is set with some details.
```
jandubois:
```suggestion
container, a `CreateFailed` status is set with some details.
```

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
+
+If multiple `ContainerRequest` objects exist at the same time for the same
+contained `name`/`namespace` pair, the result is undefined.
```
jandubois: Can't we have an admission controller that makes this impossible?

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
+
+#### Change container state
+Set `.spec.state` to match how Kubernetes resources normally work.
+
+#### Fetch container logs
+The `@kubernetes/client-node` package has some hand-written code to deal with
+logs; maybe we can make a copy of that (with a different endpoint) for this?
```
jandubois, line 134: The challenge is creating the endpoint. I've looked at this before, and you essentially end up with a separate API/port; aggregating the custom endpoint with the k8s API was non-trivial, IIRC (but have forgotten most of it already).

This was when I looked at adding custom verbs like "restart", and decided that adding labels and letting the controller handle it is easier. But logs are obviously different.

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
+
+#### Delete container
+Delete the `Container` object; a finalizer will be used to delete the container,
+at which point the `Container` object will actually be deleted.
+
 ## Images
 
-`Image` objects reflect images in the container engine.
+`Image` objects reflect images in the container engine.  Each tag is represented
+as a new `Image` object (therefore there may be duplication).
```
jandubois: It is not clear what "duplication" means here. Technically there cannot be duplicate resources in k8s.

docs/design/api_containers.md
jandubois: Yes, I think putting it in `.status` and using a field selector is the best we can do. Assuming filtering all images for a specific tag is what we expect to need most.

I couldn't come up with anything better either.

docs/design/api_containers.md
```diff
@@ -64,21 +65,92 @@ status:
...
 metadata:
-  name: 'sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # Image ID
-  namespace: default # Not related to containerd namespace
+  name: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345' # Image ID, colon replaced with dot, with random suffix.
+  namespace: rancher-desktop # not the containerd namespace
   labels:
-    namespace: k8s.io # containerd namespace, or `moby` if using dockerd
+    # If an image is untagged, `name` and `namespace` are missing.
+    name: 'registry.opensuse.org/opensuse/leap:latest' # image tag
+    id: 'sha256:999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b' # image ID
```
jandubois: Doesn't this have the same problem?

We can move it to `.metadata.annotations` if we just want to store the value, but won't be able to select on it. Or we can also move it to `.status`.

docs/design/api_containers.md
```diff
@@ -87,17 +159,75 @@ spec:
...
+
+#### Build image
+Not sure; do something with the `Resource` API maybe?
+
+We may need an `ImageBuildRequest` job-thing or something?
+
+#### Push image
+Set `.spec.push` to `true`.  The image will be pushed by `.metadata.labels.name`.
+After the push, `.status[?(@.type=="Pushed")]` will be set to `True`, and
+`.spec.push` will be set to `false`.
```
jandubois: I think we need an `ImagePushRequest` for this. The `Image` controller should not modify the `.spec` part, only the `.status`.

(And yes, we will need an `ImageBuildRequest` if we want to support building images).

docs/design/api_containers.md
```diff
@@ -87,17 +159,75 @@ spec:
...
+
+#### Scan image
+We will need a new object type for this; maybe something like
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: ImageScanRequest
+metadata:
+  name: image-scan-12345
+  namespace: rdd-system # not containerd namespace
```
jandubois:
```suggestion
  namespace: rancher-desktop # not containerd namespace
```

docs/design/api_containers.md
```diff
@@ -87,17 +159,75 @@ spec:
...
+spec:
+  # The `.metadata.name` of an `Image` object.
+  imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345'
+status:
+  conditions:
+  - type: Finished
+    status: True
+  result:
+    # Just dump the Trivy result JSON here (without converting to YAML).
```
jandubois: There may be a length limit problem here, but we can figure this out at a later time.

Although this may not apply here because we are using sqlite instead of etcd. Needs testing. Still, dumping the whole result into the status might be excessive.

docs/design/api_containers.md
```diff
@@ -87,17 +159,75 @@ spec:
...
+```
+
+#### Untag image
+Delete the `Image` object; the finalizer will untag the image.  If all tags of
+an image are removed, _and_ it is not in use by a container (running or not),
+then the image itself is removed.
```
jandubois, line 246: I assume it becomes a "dangling" image after all tags have been deleted, if it is still in use. Eventually we may want a way to delete those too (the `--force` mechanism), but we can punt on it for now.

docs/design/api_containers.md
```diff
@@ -87,17 +159,75 @@ spec:
...
 apiVersion: containers.rancherdesktop.io/v1alpha1
 kind: Volume
 metadata:
-  name: volume-name
+  name: volume-name-12345 # based on containerd name / namespace?
   namespace: default # Not related to containerd namespace
   labels:
+    name: volume-name
     namespace: k8s.io # containerd namespace, or `moby` if using dockerd
 spec:
```
jandubois: Most (all?) of the `.spec` fields need to move to `.status`.

jandubois (changes requested):
Still have some requests...

docs/design/api_containers.md
mook-as: I was initially thinking of just waiting a minute (in case the entity creating the request wanted to examine it).  I just didn't put it in the API because I didn't think that the specific amount of time should be guaranteed.

Implementing it as a helper controller does seem to make sense (presumably using an annotation?)

docs/design/api_containers.md
mook-as: Good point, we should do that instead (cribbing off the thing that make App a singleton).

docs/design/api_containers.md
mook-as: Yeah, the endpoint will need to upgrade to web sockets.

Possible reference: https://kubernetes.io/docs/tasks/extend-kubernetes/configure-aggregation-layer/

We'll need to figure out how to make a `corev1.Service` that points outside the Kubernetes cluster (at our binary), though. Should be doable if we manually make the Endpoints for it?

docs/design/api_containers.md
mook-as: Will expand. (There will be more than on `Image` object per image ID, in the case of multiple image tags.)

docs/design/api_containers.md
mook-as: Yep, currently fighting this; I haven't done custom field selectors before…

docs/design/api_containers.md
mook-as: Yeah, there needs to be a revision.  I'll do that once I get something that _works_, though, so I don't have more surprises caused by things we can't implement.

docs/design/api_containers.md
mook-as: Makes sense.

docs/design/api_containers.md
mook-as: Think of it more as _I don't want to deal with that schema right now_.

mook-as (commented):
Just responding to some comments now; don't need to look at it immediately, since I need to actually make changes to the document for that to be worthwhile.

docs/design/api_containers.md
```diff
@@ -135,36 +139,33 @@ at which point the `Container` object will actually be deleted.
...
   size: 45150437
   labels:
     org.opensuse.base.vendor: openSUSE Project
   conditions:
-  - type: PullStarted
-    status: True
-  - type: Ready
-    status: True
-  - type: Pushed
+  - type: Pushed # Used to record last push time.
```
jandubois: The comment says "push time", but the status is just boolean.

I wonder if this condition belongs here. Why do we need it? I think just having it on `ImagePullRequest` should be enough.

Should `Pushed" be true if we pulled the tag instead of building it locally?

I would drop this condition until we know we need it, and we can discuss the exact semantics at that time.

docs/design/api_containers.md
jandubois: Should we call it "Pull image" for consistency?

docs/design/api_containers.md
```diff
@@ -190,9 +192,22 @@ Not sure; do something with the `Resource` API maybe?
...
+metadata:
+  name: image-push-12345
+  namespace: rancher-desktop
+spec:
+  # `.metadata.name` of the image tag to push.
+  imageRef: 'sha256.999adf320e40662dc96119a14f07459af9959a081d10ccab7c405257030ab96b-12345'
+status:
+  conditions:
+  - type: Finished
+    status: False
```
jandubois: For the `ImagePullRequest` we have a `Pulled` condition, but for `ImagePushRequest` we have `Finished`. It should either be `Pushed`, or we should use `Finished` for all our `*Request` resources.

I think we should copy the conditions used by the `Job` resource: `Complete` and `Failed`.

Note that conditions are supposed to be "state adjectives", not "past-tense verbs". I agree that "Failed" makes this confusing because it is kind of both.

docs/design/api_containers.md
```diff
@@ -210,14 +225,23 @@ status:
...
+    '{ ... }'
 ```
 
 #### Untag image
 Delete the `Image` object; the finalizer will untag the image.  If all tags of
 an image are removed, _and_ it is not in use by a container (running or not),
 then the image itself is removed.
 
+If a container is using an image via a tag, then removing that tag may end up
+creating a new `Image` object to represent the untagged image.
```
jandubois, line 249: This is saying that we create dangling image references "on demand" only? What if we start up the container engine, and it already has a dangling image that is not in use by a container anymore?

jandubois (changes requested):
Almost ready, but I still have a few minor comments.

docs/design/api_containers.md
mook-as: That's because conditions inherently have a last transition time:
https://github.com/kubernetes/apimachinery/blob/8931c298fc6d5097879d7f34f8cf8ac50a05975c/pkg/apis/meta/v1/types.go#L1617

But I think you're right, having it on `ImagePushRequest` is probably fine (not the pull request).

docs/design/api_containers.md
mook-as: `Complete` sounds good, I'll try to use that everywhere.

Note that according to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

> Condition type names should describe the current observed state of the resource, rather than describing the current state transitions. This typically means that the name should be an adjective ("Ready", "OutOfDisk") or a past-tense verb ("Succeeded", "Failed") rather than a present-tense verb ("Deploying").

So it's just supposed to avoid present-tense verbs; but then it immediately contradicts with
> For state transitions which take a long period of time (e.g. more than 1 minute), it is reasonable to treat the transition itself as an observed state.  In these cases, the Condition (such as "Resizing") itself […]

docs/design/api_containers.md
mook-as: That's not my intention; I'm saying that we will create dangling image references if there's a dangling (untagged) image.  Deleting the last tag of an image with a running container makes it dangling, much in the way that starting with a dangling image makes it dangling (by tautology).

docs/design/api_containers.md
```diff
@@ -164,14 +171,12 @@ status:
   size: 45150437
   labels:
     org.opensuse.base.vendor: openSUSE Project
-  conditions:
-  - type: Pushed # Used to record last push time.
-    status: False
+  conditions: []
```
jandubois, line 174: Not sure if having an empty `conditions` list makes sense.

jandubois (approved):
Thanks, LGTM

docs/design/api_containers.md
mook-as: I left it there because it's actually possible to set conditions from elsewhere (as long as they have write access); that, and we'll probably want it in the future anyway. It doesn't hurt for now.

PR #117, opened 2026-01-21
Implement new container API

mook-as:
This implements the API from #110 for the stub resource types.

There is, of course, still no actual backend implementation; it just makes those resources usable so they can be mocked.

Notes:
- The CRDs are still auto-generated; reviewing them can probably be skipped (as long as the source files are reviewed).  Maybe run `make generate` to ensure I didn't accidentally change them.
- A controller may now return CRD data that consists of multiple CRD documents concatenated together in the normal YAML way; this is so e.g. `ContainerCreateRequest` lives in the `Container` controller.

Excluding the generated files, this contains:
>  9 files changed, 423 insertions(+), 95 deletions(-)


pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -141,6 +141,78 @@ type ContainerList struct {
 	Items           []Container `json:"items"`
 }
 
+type ContainerCreateSpec struct {
+	// path is the path to the executable (within the image) for the process.
```
jandubois: The name in the comments is usually the Go name, not the JSON name. So all names here should have a capital first letter.


```suggestion
	// Path is the path to the executable (within the image) for the process.
```

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -141,6 +141,78 @@ type ContainerList struct {
...
+	// Defaults to the image's default command if not specified.
+	// +optional
+	Path string `json:"path"`
+	// args is the arguments to the executable.
+	// +optional
+	Args []string `json:"args"`
+	// image is the image the container was created with.
+	// May be either a tag or a digest.
```
jandubois: Maybe specify how the digest is provided; just plain hex string, or with digest name as a prefix "sha256:..."?

pkg/apis/containers/v1alpha1/containers_types.go
```diff
@@ -141,6 +141,78 @@ type ContainerList struct {
...
+	// +required
+	Image string `json:"image"`
+	// ports describes the exposed ports of the container.
```
jandubois: What is the format for multiple ports? Comma separated, space separated, are ranges supported?

pkg/apis/containers/v1alpha1/image_types.go
```diff
@@ -65,6 +74,158 @@ type ImageList struct {
...
+type ImagePullRequestSpec struct {
+	// RepoTag is the image to pull.
+	// +required
+	RepoTag string `json:"repoTag"`
+}
+
+type ImagePullRequestStatus struct {
+	// conditions represent the state of the image pull request.
+	// Current known condition types include:
+	//  - "Pulled": the image pull request has been completed.
```
jandubois: Too lazy to look it up, but didn't we agree to use `Failed` and `Complete` as the conditions for all request objects instead of using a custom name for each?

pkg/apis/containers/v1alpha1/image_types.go
```diff
@@ -65,6 +74,158 @@ type ImageList struct {
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ImageScanRequest `json:"items"`
+}
+
 func init() {
-	SchemeBuilder.Register(&Image{}, &ImageList{})
+	SchemeBuilder.Register(
+		&Image{}, &ImageList{},
+		&ImagePullRequest{}, &ImagePullRequestList{},
```
jandubois, line 257: This is missing an entry for `ImagePushRequest`.

jandubois (changes requested):
Thanks, looks good! Just some naming convention feedback, and one CRD is not getting registered.

pkg/apis/containers/v1alpha1/containers_types.go
mook-as: I don't understand; this is a list, so it's just individual items:
```yaml
ports:
- { name: "65536/sctp", bindings: [ … ] }
- { name: "0/atp", bindings: [ … ] }
```

pkg/apis/containers/v1alpha1/image_types.go
mook-as: Thanks, I did forget to update conditions.  Actually, all of them; to quote the markdown docs:

> For the `*Request` resources, they use the `Complete` and `Failed` conditions to express state; those are mutually exclusive (only one of the two can be set to `True` at once).  Once either is set to `True`, the request object is considered to be in a terminal state and will be removed after some timeout.  This will be at least one minute (so the caller can read any data), but the precise timing is unspecified.

jandubois (approved):
Thanks, LGTM

PR #118, opened 2026-01-21
Validate Lima templates with limayaml

jandubois:
Parse and validate ConfigMap template data using Lima's limayaml package. This catches invalid YAML at admission time rather than at VM creation. Previously, validation only verified the key existed and was non-empty.

Move `validateTemplateData()` to `validation.go` and make `ValidateLimaVM` private since nothing external uses it.

This PR also includes a commit to use a shorter timeout for external controllers to detect when the control plane goes away. Before it might have taken 30s to actually fail the connection, which in turn would fail the BATS test waiting for the external controller to shut down.


bats/tests/30-builtin-controllers/namespace.bats
```diff
@@ -41,7 +41,7 @@ spec:
 EOF
 
     # LimaVM (lima.rancherdesktop.io)
-    rdd ctl create configmap lima-template -n multi-ns --from-literal=template='{"vmType":"vz"}'
+    rdd ctl create configmap lima-template -n multi-ns --from-literal=template='images: [{"location":"https://foo"}]'
```
mook-as: Using `example.invalid` instead of `foo` should ensure things fail quickly (instead of attempting to actually look up `foo`).

bats/tests/33-lima-controllers/limavm-lifecycle.bats
```diff
@@ -172,6 +181,21 @@ EOF
...
+    run -1 create_limavm "test-vm-bad-yaml" "bad-yaml"
+    assert_output --partial "failed to parse template"
+}
+
+@test "create LimaVM with invalid Lima schema fails" {
+    # arch parses fine as a string but fails Lima validation (not a valid architecture)
+    rdd ctl create configmap "bad-lima" --namespace "${NAMESPACE}" --from-literal='template=arch: "invalid"'
+
+    run -1 create_limavm "test-vm-bad-lima" "bad-lima"
+    assert_output --partial "failed to validate template"
```
mook-as, line 196: It would be good to check that the validation says something about what failed; that is, it must contain both `arch` (the key) and `invalid` (the value).  Maybe change the value to something like `mips` to make it clearer it's not part of the fixed message.

pkg/controllers/lima/limavm/validation.go
```diff
@@ -63,3 +66,26 @@ func ValidateLimaVM(ctx context.Context, c client.Client, limavm *v1alpha1.LimaV
...
+	}
+
+	// Call the Lima validator
+	y, err := limayaml.Load(ctx, []byte(templateData), "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse template: %w", err)
+	}
+	// Second parameter controls whether to log warnings via logrus.Warn
+	if err := limayaml.Validate(y, false); err != nil {
+		return nil, fmt.Errorf("failed to validate template: %w", err)
```
mook-as, line 87: The error is somewhat hard to read, but I guess there isn't much we could sensibly do here:

>  Error from server (Forbidden): error when creating "STDIN": admission webhook "limavm-defaulter.lima.rancherdesktop.io" denied the request: admission webhook "limavm-configmap-validator.lima.rancherdesktop.io" denied the request: failed to validate template: field `arch` must be one of [x86_64 aarch64 armv7l ppc64le riscv64 s390x]; got "invalid"

pkg/controllers/lima/limavm/validation_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{
+			name:        "valid minimal template",
+			data:        map[string]string{v1alpha1.TemplateConfigMapKey: `images: [{location: "https://example.com/image.qcow2"}]`},
+			expectError: false,
+		},
+		{
+			name:        "valid template with memory",
+			data:        map[string]string{v1alpha1.TemplateConfigMapKey: `{"memory":"2GB","images":[{"location":"https://example.com/image.qcow2"}]}`},
+			expectError: false,
+		},
```
mook-as, line 55: Should there be a test given a valid template plus an unknown field? That seems to pass currently.

mook-as (dismissed):
The change is fine, but there's improvements we can make before merge too.

bats/tests/30-builtin-controllers/namespace.bats
jandubois: I don't know; nothing should be looking of `foo`; this is just text that is going to be ignored. Or do you mean the human reader of the code?

Anyways, will change to `example.invalid`, as it makes no difference either way.

pkg/controllers/lima/limavm/validation.go
jandubois: I don't know what you want to do. This is just the result of recursively wrapping the error result at each caller level.

The whole `admission webhook "..." denied the request` thing doesn't even come from us; we only provide the Lima validation error; the rest is Kubernetes scaffolding.

pkg/controllers/lima/limavm/validation_test.go
jandubois: I added it, but this is potentially "subject to change" from the Lima side.

bats/tests/30-builtin-controllers/namespace.bats
mook-as: Yes, I mean me reading the code.

pkg/controllers/lima/limavm/validation_test.go
mook-as: I'm happy to know when they change it, since we might care at that point.

That's why I phrased it as _seems to pass currently_ :D

pkg/controllers/lima/limavm/validation.go
mook-as: Yeah, I don't think there's anything we can do here. Just lamenting at the injustice of the world, or whatever.

PR #130, opened 2026-01-28
Create Lima instance after template validation

jandubois:
The reconciler creates a Lima instance on disk after the template ConfigMap passes validation. Each reconcile performs at most one mutation, then returns to let the next reconcile proceed with fresh state. This avoids ResourceVersion conflicts from multiple mutations within a single reconcile.

Instance creation calls both `Create` (config files) and `Prepare` (download images, create disks) for early error detection. If `Prepare` fails, the reconciler deletes the partially created instance to prevent future reconciles from seeing it as complete. Instance creation is idempotent: if an instance already exists, the reconciler sets the `InstanceCreated` condition without recreating it. Leftover instances from failed deletions are cleaned up before any mutations.

Changes:
- Add `PathDir()` and `LimaHome()` helpers to `pkg/instance`
- Set `LIMA_HOME` at controller startup
- Create instance via `limainstance.Create()` and `Prepare()`
- Clean up partial instance on Prepare failure
- Delete instance in `handleDeletion()` before resource cleanup
- Track creation state via `InstanceCreated` condition
- Import platform-specific Lima drivers
- Add BATS tests for instance lifecycle and failure scenarios


bats/helpers/paths.bash
```diff
@@ -43,4 +43,7 @@ if is_windows; then
     PATH_DISTRO_DATA="${PATH_APP_HOME}/distro-data"
 fi
 
-LIMA_HOME="${PATH_APP_HOME}/lima"
+# PATH_RD is the "path directory" (e.g., ~/.rd2) as documented in docs/design/cmd_service.md.
```
mook-as: I don't see anything about `~/.rd2` in that file; could you elaborate?

Note that this directory sounds very incorrect on Windows (very few things should go into the home directory).  And if we can figure out the path length issues, we should move things out on mac too.

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+# LimaVM instance tests verify that Lima instances are created on disk
+# after template verification, and deleted when the LimaVM is deleted.
+
+NAMESPACE="instance-test-ns"
+VM_NAME="alpine-test"
+
+# Use a real Alpine ISO template that instance.Create() can use.
+# This is a minimal template with actual image URLs.
+# vmType defaults to vz on macOS and qemu on Linux.
+ALPINE_TEMPLATE='images:
```
mook-as: To make this more readable, maybe a new line before `images:` (so it's essentially a blank line at the start of the YAML file)?

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+- location: https://github.com/lima-vm/alpine-lima/releases/download/v0.2.47/alpine-lima-std-3.23.0-x86_64.iso
+  arch: x86_64
+  digest: sha512:c71e21dfb152642dd79af281497f86e7f690998997f787307978d83594e5e47addbc61e7d8ee405b0afc4230688de9eeb98fa44d6e74654e8d9d8b70151fb8da
+- location: https://github.com/lima-vm/alpine-lima/releases/download/v0.2.47/alpine-lima-std-3.23.0-aarch64.iso
+  arch: aarch64
+  digest: sha512:44659e71c1e277361bc10ecc813fba799d0371c2bc291db811e05e43429fd31aaa7ebe185331d02dccadccddfe9d54184376cdceeb1c444b58e3e9a4e690ce33
+mounts:
+- location: "~"
```
mook-as: Do we need this mount for testing? It seems better to go without, so we can deal with possible mount issues in a different test.

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+    done
+}
+
+local_setup() {
+    skip_on_windows
+}
+
+create_limavm() {
+    local name=$1
+    local template=$2
```
mook-as: I think calling this `template_name` would be clearer; when reading it it sounds like `template` should contain YAML instead of the name of a config map.

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+    assert_file_exists "${LIMA_HOME}/${VM_NAME}/lima.yaml"
+}
+
+@test "verify Lima instance has basedisk" {
+    assert_file_exists "${LIMA_HOME}/${VM_NAME}/basedisk"
+}
+
+@test "verify Lima instance has diffdisk" {
+    assert_file_exists "${LIMA_HOME}/${VM_NAME}/diffdisk"
+}
```
mook-as: Not sure we should be testing the contents of the directory, other than `lima.yaml`.  I realize this is because we build it, but the format shouldn't matter as long as the VM runs (and we'll test that later).

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+
+@test "verify InstanceCreated condition has message" {
+    run -0 rdd ctl get limavm "${VM_NAME}" --namespace "${NAMESPACE}" \
+        --output jsonpath='{.status.conditions[?(@.type=="InstanceCreated")].message}'
+    # Message is "Lima instance created successfully" or "Lima instance exists"
+    assert_output --partial "Lima instance"
+}
+
+@test "delete LimaVM resource" {
+    rdd ctl delete limavm "${VM_NAME}" --namespace "${NAMESPACE}" --grace-period=0
```
mook-as: Would zero grace period mean (once we run the VM) we'll end up not shutting it down / deleting correctly?  That is, make the next test step flaky?

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+@test "verify Lima instance is deleted from disk" {
+    try --max 30 --delay 1 --until-fail -- lima_instance_exists "${VM_NAME}"
+}
+
+# Test that leftover instances from failed deletions are cleaned up
+
+@test "create fake leftover Lima instance" {
+    mkdir -p "${LIMA_HOME}/${VM_NAME}"
+    echo "leftover: true" >"${LIMA_HOME}/${VM_NAME}/lima.yaml"
+    echo "0.0.0" >"${LIMA_HOME}/${VM_NAME}/lima-version"
```
mook-as: Writing files is annoying on Windows.

See `create_file`: https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/e7bbbc90a0d1182888c7c6e4461b916f945190da/bats/helpers/utils.bash#L388

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+# after template verification, and deleted when the LimaVM is deleted.
+
+NAMESPACE="instance-test-ns"
+VM_NAME="alpine-test"
+
+# Use a real Alpine ISO template that instance.Create() can use.
+# This is a minimal template with actual image URLs.
+# vmType defaults to vz on macOS and qemu on Linux.
+ALPINE_TEMPLATE='images:
+- location: https://github.com/lima-vm/alpine-lima/releases/download/v0.2.47/alpine-lima-std-3.23.0-x86_64.iso
```
mook-as, line 18: This fails on Windows because:
> 'failed to resolve vm for "C:\\Users\\Bob\\.rdbats-lima-controller\\lima\\alpine-test\\lima.yaml": unsupported image type for vmType: wsl2, tarball root file system required: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.47/alpine-lima-std-3.23.0-x86_64.iso"'

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+
+@test "verify Lima instance is deleted from disk" {
+    try --max 30 --delay 1 --until-fail -- lima_instance_exists "${VM_NAME}"
+}
+
+# Test that leftover instances from failed deletions are cleaned up
+
+@test "create fake leftover Lima instance" {
+    mkdir -p "${LIMA_HOME}/${VM_NAME}"
+    echo "leftover: true" >"${LIMA_HOME}/${VM_NAME}/lima.yaml"
```
mook-as: This needs a comment; until I got to the other tests I assumed `leftover:` was a special key in `lima.yaml`.

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -0,0 +1,217 @@
...
+
+@test "cleanup LimaVM after leftover test" {
+    rdd ctl delete limavm "${VM_NAME}" --namespace "${NAMESPACE}" --grace-period=0
+    try --max 30 --delay 1 --until-fail -- lima_instance_exists "${VM_NAME}"
+}
+
+# Test that invalid image URL causes InstanceCreated condition to be False
+
+INVALID_TEMPLATE='images:
+- location: https://invalid.example.com/nonexistent.iso
```
mook-as: Consider using `https://example.test/…` because that TLD shouldn't resolve. Due to… decisions on IANA's side, `example.com` is an actually real FQDN now.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -22,44 +29,49 @@ import (
...
 // +kubebuilder:rbac:groups=lima.rancherdesktop.io,resources=limavms,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=lima.rancherdesktop.io,resources=limavms/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=lima.rancherdesktop.io,resources=limavms/finalizers,verbs=update
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
+// Reconcile moves the cluster state toward the desired state.
 //
 // Webhook and reconciler responsibilities:
```
mook-as, line 94: I think this needs a mermaid diagram somewhere, and this should have a pointer to that.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -70,23 +82,27 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		}
 	}
 
-	// Fetch the template ConfigMap to set owner reference
-	// The mutating webhook creates it during admission, but cannot set the owner reference
-	// (requires LimaVM UID which is only available after the resource is persisted)
+	// Get the template ConfigMap (created by the mutating webhook)
 	configMapName := limaVM.GetTemplateConfigMapName()
 	templateConfigMap := &corev1.ConfigMap{}
 	if err := r.Get(ctx, types.NamespacedName{Name: configMapName, Namespace: limaVM.Namespace}, templateConfigMap); err != nil {
```
mook-as: This line is a bit long; consider splitting the `err :=…` assignment and `if err …` into two statements, so we can wrap the first as needed.  Or just create the `NamspacedName` separately.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, err
 		}
 		logger.Info("Set owner reference on template ConfigMap", "ConfigMap.Name", configMapName)
+		return ctrl.Result{}, nil
```
mook-as, line 188: Does this need to set `RequeueAfter: time.Second` or something along those lines?  Same for the other places this function returns early without an error.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	}
 
-	// Update status.templateConfigMap to mark initialization as complete
-	limaVM.Status.TemplateConfigMap = configMapName
+	// Record the template ConfigMap name in status
+	if limaVM.Status.TemplateConfigMap == "" {
+		limaVM.Status.TemplateConfigMap = configMapName
+		if err := r.Status().Update(ctx, &limaVM); err != nil {
+			logger.Error(err, "Failed to update status.templateConfigMap")
+			return ctrl.Result{}, err
```
mook-as, line 196: I suspect we should check if `err` is conflict, and requeue without error in that case. (Just returning the error would mean exponential backoff, and eventually failing the reconcile unnecessarily.)

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	// Instance exists on disk (perhaps from a previous reconcile); record the condition
+	existingInst, err := store.Inspect(ctx, limaVM.Name)
+	if err == nil && existingInst != nil {
+		logger.Info("Lima instance already exists", "instance", limaVM.Name)
+		r.setCondition(&limaVM, ConditionInstanceCreated, metav1.ConditionTrue, ReasonCreated, "Lima instance exists")
+		if statusErr := r.Status().Update(ctx, &limaVM); statusErr != nil {
+			logger.Error(statusErr, "Failed to update status for existing instance")
+			return ctrl.Result{}, statusErr
+		}
+		return ctrl.Result{}, nil
```
mook-as, line 216: We will probably need to inspect the state too? For example, we seem to create the instance later before downloading the images, so if the download files it'll get into a wedged state instead of retrying the download.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	}
+
+	// Extract template data
+	templateData, ok := templateConfigMap.Data[v1alpha1.TemplateConfigMapKey]
+	if !ok || templateData == "" {
+		err := errors.New("template ConfigMap missing template data")
+		logger.Error(err, "Template ConfigMap has no template key")
```
mook-as: Logging `err` here doesn't seem very useful, since the message is just duplicated.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		r.setCondition(&limaVM, ConditionInstanceCreated, metav1.ConditionFalse, ReasonCreateFailed, err.Error())
+		if statusErr := r.Status().Update(ctx, &limaVM); statusErr != nil {
+			logger.Error(statusErr, "Failed to update status")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Create the Lima instance
+	inst, err := limainstance.Create(ctx, limaVM.Name, []byte(templateData), false)
```
mook-as, line 231: This can potentially take a while, right?

We should probably set `ConditionInstanceCreated` to unknown first.  Maybe do it right after we check the previous condition?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		if statusErr := r.Status().Update(ctx, &limaVM); statusErr != nil {
+			logger.Error(statusErr, "Failed to update status after instance creation failure")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Prepare the instance: download images and create disks.
+	// The guestAgent path is unused during Prepare (only stored for later);
+	// limactl start will look up the real path when called.
+	if _, err := limainstance.Prepare(ctx, inst, "unused"); err != nil {
```
mook-as: Dumb suggestion: use `placeholder` instead of `unused`, to make it more obvious.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		logger.Error(err, "Failed to prepare Lima instance")
+		// Clean up the partially created instance so the next reconcile doesn't
+		// see it as existing and skip creation.
+		if delErr := limainstance.Delete(ctx, inst, true); delErr != nil {
+			logger.Error(delErr, "Failed to clean up instance after prepare failure")
```
mook-as, line 258: We might get wedge here.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -97,36 +113,160 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+}
+
+// handleDeletion cleans up when a LimaVM is being deleted.
 func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.LimaVM) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
-	// Clean up all owned resources using the base helper
+	// Delete the Lima instance
+	existingInst, err := store.Inspect(ctx, limaVM.Name)
+	if err == nil && existingInst != nil {
```
mook-as: Shouldn't we do something if `err` is set? At least log it? I assume that if the instance doesn't exist, both `err` and `existingInst` are nil.

pkg/controllers/lima/limavm/controller.go
```diff
@@ -96,9 +98,11 @@ func (c *controller) GetCRDData() string {
 // setupReconciler sets up the LimaVMReconciler with the manager.
 func (c *controller) setupReconciler(mgr ctrl.Manager) error {
 	return (&controllers.LimaVMReconciler{
-		Client:  mgr.GetClient(),
-		Scheme:  mgr.GetScheme(),
-		Manager: mgr,
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Manager:  mgr,
+		Recorder: mgr.GetEventRecorderFor(ControllerName + "-controller"), //nolint:staticcheck // new API requires events.k8s.io
```
mook-as: Since this is new code, any reason not to just use `events.Recorder` here?

pkg/controllers/lima/limavm/drivers_vz.go
mook-as, line 1: Sounds like this file should be call `drivers_vz.go`, and `drivers_windows.go` should be `drivers_wsl2.go`.

pkg/instance/instance.go
```diff
@@ -73,3 +73,21 @@ var PIDFile = sync.OnceValue(func() string {
 var TLSDir = sync.OnceValue(func() string {
 	return filepath.Join(Dir(), "tls")
 })
+
+// PathDir returns the path directory for this instance (e.g., ~/.rd2).
+// This is distinct from Dir() which returns the service directory.
+// See docs/design/cmd_service.md for directory documentation.
+var PathDir = sync.OnceValue(func() string {
```
mook-as: Calling it _path directory_ is confusing; we could use a better name. Semantically, this is closer to _short instance directory_.

Actually, isn't this going to break horribly if multiple RDD instances run with different `RDD_INSTANCE` environment variables (and they each need their own `lima` directory)?

.golangci.yml
```diff
@@ -83,6 +83,8 @@ linters:
       # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md
       rules:
       - name: blank-imports
+        exclude:
+        - "**/drivers_*.go"
```
mook-as: The path should be much more specific (`pkg/controllers/lima/limavm/drivers_*.go`).

There should also be a justification comment, so we can figure out if we still need it later.

mook-as (changes requested):

bats/helpers/paths.bash
jandubois: Sorry, the filename is defined in `README.md`:

---

### Directories

The control plane has both a [service directory](cmd_service.md#service-directory) `$APPDATA/rancher-desktop-2` and a [path directory](cmd_service.md#path-directory) `~/.rd2`[^lima]. `$APPDATA` is the platform-specific application data directory, e.g. `~/Library/Application Support` on macOS.

[^lima]: RDD will set `LIMA_HOME` to `~/.rd2/lima` instead of `$APPDATA/rancher-desktop-2/lima` because of socket name length constraints.

It is currently not a goal to support RDD running as a system service; it is always tied to a user session[^WSL2]. Design decisions should still attempt to not make that harder, in case this ever becomes a goal.

[^WSL2]: On Windows WSL2 also depends on the user being logged in. It would require a Hyper-V Lima driver before it could run as a system service.

---

But the directory content is defined in `cmd_service.md`:

---

### Path Directory

The path directory contains the following files and directories:

| File | Description |
| --- | --- |
| `bin`         | contains utilities like `docker`, `helm`, etc. May be symlinks |
| `kube.config` | Kubernetes config only containing the `rancher-desktop-2` context |
| `lima`        | `LIMA_HOME` is located here because of socket name length restrictions |

---

> if we can figure out the path length issues, we should move things out on mac too.

The max length of socket filenames is a restriction of macOS/Linux; there isn't really anything we can do about it.

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: I know, the whole PR does not support Windows yet, which is why all the tests are being skipped.

To keep PRs small I would prefer to implement the Windows bits in a separate PR.

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: We don't; will remove.

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: I added it to verify that `Prepare()` has been called successfully. I originally didn't and wondered why I didn't have a disk image. Twocows doesn't call it either. It still works because `Start()` will still call it. But I wanted to call it as part of the creation, to get errors early.

I can remove the tests because I don't think we'll accidentally remove the call to `Prepare()`. 😄

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: Good point, I need to research that.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Ok, but maybe in a separate PR? I only condensed the existing info because I thought it was too verbose.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: No, the `r.Update()` modifies the resource which triggers another reconcile.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: I don't understand. Why would we not want the exponential backoff? Either it works and will succeed in the next few retries, or is seriously broken and probably will never succeed.

I think retries without backoff should be considered a bug.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: The files would be downloaded as part of `Prepare`. That's why we delete the instance if `Prepare` returns an error, so we don't end up with a broken instance that we assume is correct.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: We can do this. There seem to be 2 patterns:

1. Create the condition right away and set to "Unknown" so the user knows the request is being processed.
2. Don't create the condition until the status is either "True" or "False".

We currently do the latter, even though we define the "Unknown" setting as a possible value.

I don't know which pattern is more useful from the client side. Conditions is just a slice, and a specific condition just doesn't exist until it has been set the first time. So I think "Unknown" is really only useful if the condition can switch back to "Unknown" after it has been "True" or "False" already. Otherwise the absence of the condition provides the same signal.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yeah, I don't know what to do. We can try to create a marker file. That might be better anyways, so we can deal with it with the already existing code on top that deletes an old existing instance. And then it will just retry the deletion there until it succeeds.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Good point. I have to double-check the API anyways; it has very weird error semantics, e.g. I think `existingInst.errors` can be another list of errors.

pkg/controllers/lima/limavm/drivers_vz.go
jandubois: Makes sense; I copied the names from Lima:

```
❯ ls cmd/limactl/main*
cmd/limactl/main_darwin.go	cmd/limactl/main_windows.go
cmd/limactl/main_qemu.go	cmd/limactl/main.go
```

pkg/instance/instance.go
jandubois: Yes, I don't like the name either. I'm fine with `ShortDir` unless you have a better suggestion.

There is no conflict because the short dir includes the instance suffix, e.g. `~/.rd2`, `~/.rd3` etc. corresponding to `rancher-desktop-2` and `rancher-desktop-3` application directories.

jandubois (commented):
Just some quick feedback on the review comments before I start working on them...

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: The grace period is just set on the resource metadata. For Pods it is used to send a SIGTERM first, and if that doesn't stop the pod before the time is up, it sends a SIGQUIT. We should do the same for LimaVM instances: do regular delete first, and if it doesn't shut down within the grace period, do a force-delete.

The finalizers will always execute, even when grace-period is 0.

But for now this just sets metadata that is being ignored, so I'm removing it again from the test until we actually need it.

bats/helpers/paths.bash
mook-as: Ah, alright, then.  I guess keep this PR as-is, and we can evolve the directories in separate PRs later.

bats/tests/33-lima-controllers/limavm-instance.bats
mook-as: Sounds good; please make sure there's an issue filed to fix this for Windows.  Otherwise we'll probably forget about the `skip_on_windows` bit.

bats/tests/33-lima-controllers/limavm-instance.bats
mook-as: I'm okay with either commenting that this is to test `Prepare`, or removing it.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
mook-as: Yeah, separate PR sounds good.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
mook-as: Per https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties:

> Controllers should apply their conditions to a resource the first time they visit the resource, even if the `status` is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.

For things that can take some time, setting it to unknown can be useful to let the user know that things are in progress.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
mook-as: I _think_ the solution is to do more inspection initially around `existingInst`, but that might need more thought.

I guess file an issue for now, and think about it later?

pkg/instance/instance.go
mook-as: I'm fine with `ShortDir`. Or maybe `ShortInstDir`.

If you want to maintain momentum, I guess we can do that in a separate PR (that includes changes to the markdown docs).

mook-as (commented):
I guess for a lot of things I'm okay with filing an issue and doing them in separate PRs.

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: #134

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: I added a diagram in a separate commit, but I'm not convinced it really helps. It is kind of hard to read, and will be challenging to keep in sync with the code.

docs/design/api_lima.md
```diff
@@ -49,6 +53,89 @@ status:
...
+
+Deleting a `LimaVM` object triggers the finalizer to delete the Lima instance from disk and remove the template ConfigMap.
+
+### LimaVM Reconciler
+
+The reconciler creates and manages Lima instances on disk. Each reconcile performs at most one mutation, then returns to let the next reconcile proceed with fresh state.
+
+A `.preparing` sentinel file marks preparation in progress. If a reconcile fails after creating the instance but before updating the status, the next reconcile detects the sentinel and cleans up the incomplete instance.
+
+```mermaid
```
mook-as, line 74: That's… way more detail than I expected (I was just expecting those comments to be moved), but okay

pkg/controllers/lima/limavm/controllers/limavm_controller.go
mook-as: Ah, okay, that means I did the mock controller wrong; let me go make a PR to fix that.

mook-as (approved):

PR #131, opened 2026-01-28
Fix external controller shutdown timing

jandubois:
The external controller exit test failed intermittently because shutdown could exceed the 25-second timeout. Two changes address this:

1. Reduce manager graceful shutdown timeout from 30s to 10s. The default controller-runtime timeout is excessive when the control plane has already disappeared.

2. Increase test timeout from 25s to 35s to accommodate worst-case timing:
   - 2s: wait for next health check tick
   - 9s: three consecutive API timeouts (3×3s)
   - 10s: manager graceful shutdown
   - 5s: cleanup timeout

Add shutdown duration logging to aid future debugging.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -154,6 +154,8 @@ jobs:
 
     - run: make -C bats
       shell: run-in-wsl {0}
+      env:
+        RDD_TRACE: true
 
     - name: Collect RDD service logs
       if: always()
@@ -161,7 +163,7 @@ jobs:
       run: |
         mkdir -p rdd-logs
         for instance in $(make --no-print-directory -C bats bats-instances); do
-          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" 2>&1 || true
+          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" || true
         done
 
     - name: Upload RDD service logs
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -154,3 +154,20 @@ jobs:
 
     - run: make -C bats
       shell: run-in-wsl {0}
+
+    - name: Collect RDD service logs
+      if: always()
+      shell: run-in-wsl {0}
+      run: |
+        mkdir -p rdd-logs
+        for instance in $(make --no-print-directory -C bats bats-instances); do
+          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" 2>&1 || true
+        done
+
+    - name: Upload RDD service logs
+      if: always()
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+      with:
+        name: rdd-logs-${{ matrix.runs-on }}
+        path: rdd-logs/
+        if-no-files-found: ignore
--- a/bats/tests/31-rdd-controllers/notary-external.bats
+++ b/bats/tests/31-rdd-controllers/notary-external.bats
@@ -30,7 +30,7 @@ assert_process_exited() {
 
 @test "external controller starts and registers" {
     # Start external rdd-controller binary in background, capturing logs for debugging
-    "rdd-controller${EXE}" >"${BATS_FILE_TMPDIR}/controller.log" 2>&1 &
+    "rdd-controller${EXE}" >"${BATS_FILE_TMPDIR}/controller1.log" 2>&1 &
     # Store PID to verify it auto-exits later
     echo "$!" >"${BATS_FILE_TMPDIR}/controller_pid"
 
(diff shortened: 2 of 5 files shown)


.github/workflows/bats.yaml
```diff
@@ -154,3 +154,20 @@ jobs:
...
     - run: make -C bats
       shell: run-in-wsl {0}
+
+    - name: Collect RDD service logs
+      if: always()
+      shell: run-in-wsl {0}
+      run: |
+        mkdir -p rdd-logs
+        for instance in $(make --no-print-directory -C bats bats-instances); do
+          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" 2>&1 || true
```
mook-as: I believe `rdd svc log` actually only prints to stdout. By default it reads the stderr logs, but `--stdout` makes it read stdout logs instead.

Also, question I've had a while now: why is it `rdd svc` and not `rdd service`, especially in committed code?

bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -29,8 +29,8 @@ assert_process_exited() {
 }
 
 @test "external controller starts and registers" {
-    # Start external rdd-controller binary in background
-    "rdd-controller${EXE}" &
+    # Start external rdd-controller binary in background, capturing logs for debugging
+    "rdd-controller${EXE}" >"${BATS_FILE_TMPDIR}/controller.log" 2>&1 &
```
mook-as: Perhaps this should live in `${PATH_LOGS}` instead.

bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -128,11 +128,19 @@ EOF
     run -0 kill -0 "${controller_pid}"
 
     # Stop the control plane - this should trigger auto-cleanup of external controllers
+    echo "# Stopping control plane at $(date +%T || true)" >&3
```
mook-as: We should have a helper function for printing things (and have it default on for CI).

bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -159,8 +167,8 @@ EOF
 }
 
 @test "external controller runs and registers" {
-    # Start external rdd-controller binary in background
-    "rdd-controller${EXE}" &
+    # Start external rdd-controller binary in background, capturing logs for debugging
+    "rdd-controller${EXE}" >"${BATS_FILE_TMPDIR}/controller2.log" 2>&1 &
```
mook-as, line 171: This clobbers the previous log.

Perhaps we should have a `teardown` script that copies the logs out.  Or just it in the bit in `load.bash`…

pkg/external/main.go
```diff
@@ -74,11 +75,16 @@ func RunControllers(apiGroupName string) int {
 	defer cancelMonitor()
 
 	// Start control plane monitoring in background
+	// Track when shutdown is initiated for timing logs
```
mook-as: The comment is joined together and confusing now. Add a blank line?

pkg/external/main.go
```diff
@@ -128,8 +140,9 @@ func RunControllers(apiGroupName string) int {
 // This allows external controllers to automatically exit when `rdd svc stop` or `rdd svc delete` is called.
 func monitorControlPlane(ctx context.Context, apiGroupName string, config *rest.Config, log klog.Logger, cancel context.CancelFunc) {
 	// Use a short timeout for monitoring so we detect shutdown quickly.
+	// 1 second is sufficient for local API server connections.
 	monitorConfig := rest.CopyConfig(config)
-	monitorConfig.Timeout = 3 * time.Second
+	monitorConfig.Timeout = 1 * time.Second
```
mook-as:
```suggestion
	monitorConfig.Timeout = time.Second
```
No need to multiply by one.

pkg/service/controllers/manager.go
```diff
@@ -132,13 +132,17 @@ func (scm *SharedControllerManager) Start(ctx context.Context) error {
 	configCopy.ContentType = "application/json"
 
 	// Create the shared controller-runtime manager
+	gracefulShutdownTimeout := 10 * time.Second
```
mook-as: Should this be a const near the top of the file? Just having a random variable lying about seems weird.

mook-as (changes requested):
Just some small comments. Also triggering CI a bunch to see if anything pops up; so far it's just the boring notary-events one (that we should get rid of).

pkg/service/controllers/manager.go
jandubois: It is only a variable because it has to be passed by pointer. Otherwise I would have put the expression directly in the `struct` constructor.

And we cannot make it a `const` because you cannot take the address of a `const`. So I'm not sure if introducing a global variable far away from the single place where it is used as a constant it actually better.

.github/workflows/bats.yaml
jandubois: > I believe `rdd svc log` actually only prints to stdout. By default it reads the stderr logs, but `--stdout` makes it read stdout logs instead.

Yeah, but this would capture any errors we would get from `rdd svc logs`. I guess we still see them in the CI output, so it doesn't really matter. I will remove the `2>&1`.

> Also, question I've had a while now: why is it `rdd svc` and not `rdd service`, especially in committed code?

It probably should be, but we are using `rdd svc` all over the place. To me it is a perfect complement to `rdd ctl`. Of course we never even defined `rdd control` and made `ctl` just an alias (and I don't think we should).

So I'm not sure if we should get rid of `rdd service` and continue to use `rdd svc` everywhere. In some ways having aliases is just a complication, so just having a short canonical subcommand might be better. I definitely don't want to type `rdd service` all the time.

bats/tests/31-rdd-controllers/notary-external.bats
jandubois: It does not because the previous log was called `controller.log` and this one is `controller2.log`. It doesn't matter though, as the previous log is not used anymore; it is shown directly in the test output when shutdown fails.

I don't want to deal with this "properly" right now and leave it as a hack in case the test continues to fail. We can reorganize this later.

pkg/service/controllers/manager.go
mook-as: Ah, sorry, missed the reference. In that case use [`k8s.io/utils/ptr.To`](https://pkg.go.dev/k8s.io/utils/ptr#To)?

mook-as (approved):

PR #132, opened 2026-01-29
Remove the notary-events tests

jandubois:
They keep failing in CI, at least on Linux, and we don't currently have plans to use events anyways. So removing the tests unblocks us from having to deal with them.


mook-as (approved):
LGTM; we'll need to remove the notary controller at some point, but that's not in a hurry.

PR #135, opened 2026-01-30
mock: Don't manual requeue after conflict

mook-as:
If we get a conflict when updating a resource, do not manually requeue; instead, just pass the error back to get the normal exponential backoff behaviour.

Also drop manual requeuing if we fail to find CRDs or namespaces; in those cases we also did not progress on the reconcile.


jandubois (approved):
Thanks, LGTM

PR #136, opened 2026-01-30
Implement Lima VM start and stop with hostagent

jandubois:
Add start/stop logic to LimaVM controller using Lima's instance package. The controller manages VM lifecycle through the InstanceRunning condition.

Key changes:
- Add InstanceRunning condition with Started/Stopped/StartFailed/StopFailed
- Controller starts VM when spec.running=true, stops when false
- Add hostagent command to rdd and lima-controller for VM lifecycle management
- Sign binaries with virtualization entitlements on macOS
- Change spec.running from pointer to plain bool (always required)
- Replace detailed reconciler flowchart with component interaction diagram

Still ToDo in followup PRs:
- Watch for ConfigMap changes and stop/edit/start the instance accordingly
- Goroutine watcher for the host-agent, to detect state changes automatically


Makefile
```diff
@@ -45,12 +45,26 @@ ldflags:
 build: build-rdd build-all-controllers
 .PHONY: build
 
+# Code signing for macOS - required for Virtualization.framework
+# On macOS, sign binaries with virtualization entitlement using ad-hoc signature
+# On other platforms, this is a no-op
+ifeq ($(shell uname -s),Darwin)
+define SIGN_BINARY
```
mook-as: I think I'd prefer `SIGN_VZ_BINARY` to be more explicit (since we may have other scripts for final code signing before distribution), but I'm not very hung up about it; that's rather far off still.

Makefile
```diff
@@ -45,12 +45,26 @@ ldflags:
...
+codesign -f -v --entitlements macos-entitlements.plist -s - $(1)
```
mook-as: Since I don't use `codesign` much, I'd prefer long options here:
```suggestion
codesign --force --verbose --entitlements macos-entitlements.plist --sign - $(1)
```

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,217 @@
...
+    local name=$1
+    local expected=$2
+    run -0 rdd ctl get limavm "${name}" --namespace "${NAMESPACE}" \
+        --output jsonpath='{.status.conditions[?(@.type=="InstanceRunning")].reason}'
+    assert_output "${expected}"
+}
+
+lima_instance_running() {
+    local name=$1
+    assert_file_exists "${LIMA_HOME}/${name}/ha.pid"
```
mook-as, line 92: Should we assert that the process identified by that file is running?

Actually, maybe not, because that wouldn't work on Windows…

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,217 @@
...
+}
+
+assert_instance_created_condition() {
```
mook-as: I feel like this should live with the other two `assert_instance_*_condition` functions.

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,217 @@
...
+    assert_file_exists "${pid_file}"
+}
+
+@test "trigger reconcile and verify recovery from broken state" {
+    # Annotate to trigger a reconcile while spec.running is still true
+    rdd ctl annotate limavm "${VM_NAME}" --namespace "${NAMESPACE}" \
+        reconcile-trigger="$(date +%s)"
+
+    # The reconciler should detect broken state, force stop, then restart
+    try --max 600 --delay 5 -- assert_instance_running_reason "${VM_NAME}" "Started"
```
mook-as: Shouldn't we first get the last transition time stamp of the running condition, and then check that it changed, and then check that it's true?  Otherwise the test could technically pass if the reconciler just died.

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,217 @@
...
+@test "verify hostagent is alive after recovery" {
+    local pid_file="${LIMA_HOME}/${VM_NAME}/ha.pid"
+    assert_file_exists "${pid_file}"
+    local pid
+    pid=$(cat "${pid_file}")
+    assert [ -n "${pid}" ]
+    kill -0 "${pid}"
+}
+
+@test "verify BrokenStateRecovered event was emitted" {
```
mook-as, line 229: I'm not a fan of testing events, because they may get dropped. But it's probably okay here?

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,217 @@
...
+    # Clean up any remaining Lima instances
+    if [[ -d "${LIMA_HOME}/${VM_NAME}" ]]; then
+        rm -rf "${LIMA_HOME:?}/${VM_NAME}"
+    fi
+}
+
+local_setup() {
+    skip_on_windows
+    if [[ ! -f "/usr/local/share/lima/lima-guestagent.Linux-${ARCH}.gz" ]]; then
+        skip "Running tests require lima-guestagent.Linux-${ARCH}.gz"
```
mook-as: Presumably in some future PR this will be fixed to not require that. (I don't have one on my location machine, for example.)

Actually, this seems to cause CI to skip this whole test file (because that doesn't exist), so that's not very great… we may need to fix it earlier.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -199,12 +233,13 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, nil
 	}
 
-	// Instance already created
+	// Instance already created - proceed to handle running state
 	if r.hasCondition(&limaVM, ConditionInstanceCreated, metav1.ConditionTrue) {
-		return ctrl.Result{}, nil
+		return r.handleRunningState(ctx, &limaVM)
```
mook-as, line 238: No need to actually do it (and if you do please do it in a different PR), but I wonder if it makes sense for reconcile to just be a chain:

```go
  if changed, result, err := r.handleConfigMap(…); changed {
    return result, err
  }
  if changed, result, err = r.createInstance(…); changed {
    return result, err
  }
  …
```

(Obviously the code as-is doesn't make sense, but I hope it's illustrative.)

pkg/controllers/lima/limavm/controllers/limavm_controller_test.go
```diff
@@ -0,0 +1,345 @@
...
+	err := os.MkdirAll(instanceDir, 0o755)
+	assert.NilError(t, err, "failed to create instance directory")
+
+	// Override sentinelPath for testing by creating the sentinel in our temp dir
+	sentinelFile := filepath.Join(instanceDir, preparingSentinel)
+
+	t.Run("hasSentinel returns false when sentinel does not exist", func(t *testing.T) {
+		_, err := os.Stat(sentinelFile)
+		if !os.IsNotExist(err) {
+			t.Skip("Sentinel file already exists")
```
mook-as: I don't understand how this can be true…

pkg/controllers/lima/limavm/controllers/limavm_controller_test.go
```diff
@@ -0,0 +1,345 @@
...
+		}
+		// hasSentinel uses instance.LimaHome() which we can't easily override,
+		// so we test the file operations directly
+		if _, err := os.Stat(sentinelFile); !os.IsNotExist(err) {
+			t.Errorf("Expected sentinel to not exist")
```
mook-as: I'm even more confused here. We're just running the same thing again?

pkg/controllers/lima/limavm/controllers/limavm_controller_test.go
```diff
@@ -0,0 +1,345 @@
...
+				},
+			},
+		}
+		r.setCondition(limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStopped, "stopped")
+
+		assert.Equal(t, len(limaVM.Status.Conditions), 2, "expected 2 conditions")
+	})
+}
+
+func TestSentinelFunctions(t *testing.T) {
```
mook-as: I don't understand this test function. It just tries to test the standard library‽

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.LimaVM) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	// Stop and delete the Lima instance
+	existingInst, err := store.Inspect(ctx, limaVM.Name)
+	if err != nil {
+		logger.Error(err, "Failed to inspect Lima instance for deletion")
+	}
+	if existingInst != nil {
```
mook-as, line 40: We're not returning on error here? If that's intentional, it probably warrants a comment.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		// Stop the instance if running
+		if existingInst.Status == limatype.StatusRunning {
+			logger.Info("Stopping Lima instance before deletion", "instance", limaVM.Name)
```
mook-as, line 43: Should this involve setting the status conditions?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+			if err := limainstance.StopGracefully(ctx, existingInst, false); err != nil {
+				logger.Error(err, "Failed to stop Lima instance gracefully, forcing stop")
+				limainstance.StopForcibly(existingInst)
+			}
+		}
+		logger.Info("Deleting Lima instance", "instance", limaVM.Name)
+		if err := limainstance.Delete(ctx, existingInst, true); err != nil {
```
mook-as, line 50: All this stopping and deleting can take a while right? Should we do them asynchronously (and requeue)?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	shouldRun := limaVM.Spec.Running
+	isRunning := inst.Status == limatype.StatusRunning
+
+	logger.Info("Checking running state", "shouldRun", shouldRun, "isRunning", isRunning, "status", inst.Status)
+
+	if shouldRun && !isRunning {
+		// Check if hostagent is already starting (PID file exists)
+		haPIDPath := filepath.Join(inst.Dir, filenames.HostAgentPID)
```
mook-as, line 135: We may want to split this out to a new method, and also check that the process exists (i.e. the pid file is not stale).

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to get executable path")
+		if updateErr := r.updateCondition(ctx, limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStartFailed, err.Error()); updateErr != nil {
+			logger.Error(updateErr, "Failed to update status after start failure")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Prepare the instance (disk setup, downloads, etc.)
+	prepared, err := limainstance.Prepare(ctx, inst, guestAgentPath())
```
mook-as, line 196: If we need to not block, surely we shouldn't do the download synchronously?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		logger.Error(err, "Failed to prepare Lima instance")
+		if updateErr := r.updateCondition(ctx, limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStartFailed, err.Error()); updateErr != nil {
+			logger.Error(updateErr, "Failed to update status after prepare failure")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Build hostagent command arguments
+	haPIDPath := filepath.Join(inst.Dir, filenames.HostAgentPID)
+	haSockPath := filepath.Join(inst.Dir, filenames.HostAgentSock)
```
mook-as, line 207: Huh, the host agent socket can be a long path but the ssh one can't be?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	haStdoutPath := filepath.Join(inst.Dir, filenames.HostAgentStdoutLog)
+	haStderrPath := filepath.Join(inst.Dir, filenames.HostAgentStderrLog)
+
+	var args []string
+	if logrus.GetLevel() >= logrus.DebugLevel {
```
mook-as: Seems odd to check `logrus` instead of [`logger.GetV`](https://pkg.go.dev/github.com/go-logr/logr#Logger.GetV).

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		args = append(args, "--debug")
+	}
+	args = append(args,
+		"hostagent",
+		"--pidfile", haPIDPath,
+		"--socket", haSockPath)
+	if prepared.Driver.Info().Features.CanRunGUI {
```
mook-as: I don't think we'll ever want this?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		args = append(args, "--run-gui")
+	}
+	if prepared.GuestAgent != "" {
+		args = append(args, "--guestagent", prepared.GuestAgent)
+	}
+	if prepared.NerdctlArchiveCache != "" {
```
mook-as: I don't think we'll ever want this either.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+			logger.Error(updateErr, "Failed to update status after prepare failure")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Build hostagent command arguments
+	haPIDPath := filepath.Join(inst.Dir, filenames.HostAgentPID)
+	haSockPath := filepath.Join(inst.Dir, filenames.HostAgentSock)
+	haStdoutPath := filepath.Join(inst.Dir, filenames.HostAgentStdoutLog)
+	haStderrPath := filepath.Join(inst.Dir, filenames.HostAgentStderrLog)
```
mook-as, line 209: Any reason to not use our log directory instead?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		logger.Error(err, "Failed to remove old stdout log")
+		return ctrl.Result{}, err
+	}
+	if err := os.RemoveAll(haStderrPath); err != nil {
+		logger.Error(err, "Failed to remove old stderr log")
+		return ctrl.Result{}, err
+	}
+
+	// Create log files for hostagent output
+	haStdoutW, err := os.Create(haStdoutPath)
```
mook-as, line 225: [`os.Create`](https://pkg.go.dev/os@go1.25.6#Create) already truncates the file; I guess the `RemoveAll` is to deal with symlinks?  But then that's adversarial, and there's a TOCTTOU issue here.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to create stdout log file")
+		if updateErr := r.updateCondition(ctx, limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStartFailed, err.Error()); updateErr != nil {
+			logger.Error(updateErr, "Failed to update status after log file creation failure")
+		}
+		return ctrl.Result{}, err
+	}
+	haStderrW, err := os.Create(haStderrPath)
+	if err != nil {
+		haStdoutW.Close()
```
mook-as: There should just be a `defer haStdoutW.Close()` right after it's opened. The sub-process gets its own file handle, so there's no need to worry about that.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	go func() {
+		if err := haCmd.Wait(); err != nil {
+			logger.Error(err, "Hostagent process exited with error")
+		}
+		haStdoutW.Close()
+		haStderrW.Close()
+	}()
+
+	// Wait for PID file to be created (indicates hostagent has started)
+	if err := r.waitForPIDFile(haPIDPath, haStderrPath); err != nil {
```
mook-as, line 266: This sounds like it should requeue instead?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		time.Sleep(100 * time.Millisecond)
+	}
+}
+
+// stopInstance stops the Lima VM.
+func (r *LimaVMReconciler) stopInstance(ctx context.Context, limaVM *v1alpha1.LimaVM, inst *limatype.Instance) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("Stopping Lima instance", "instance", limaVM.Name)
+
+	if err := limainstance.StopGracefully(ctx, inst, false); err != nil {
```
mook-as, line 302: As above, this should be asynchronous.

mook-as (changes requested):
The main issue I think is that the reconciler is doing a lot synchronously, that's probably not a great idea.

Not for this PR, but I just found [`k8s.io/apimachinery/pkg/api/meta#SetStatusCondition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition) — Maybe we could have used that, ah well.

jandubois:
> The main issue I think is that the reconciler is doing a lot synchronously, that's probably not a great idea.

I agree, but this will be a lot of work to do correctly, and I feel we can postpone this until later.

For Rancher Desktop 2.0 tech preview, we will only have a single VM, so there are no other reconciles that could run in parallel. The image is likely already downloaded and only needs to be copied once or twice, which shouldn't take that much time.

Longer term I think the whole cache handling should be moved out of Lima into a separate cache/download resource type, that we can then wait on.

Similarly, stopping the instance asynchronously will become a lot simpler once we have a hostagent watcher that monitors all the hostagent events. We can send a SIGTERM to the hostagent, and then add a timer to the watcher, and when the timer expires, we do a forceful termination.

Both these mechanisms will be quite large and should not be part of this PR. The HA watcher is already mentioned above, and should be done "soonish". The cache/download controller will probably not happen for a while; it is not really necessary for the initial tech preview.

I'm going to address the smaller issues only for now.

Makefile
jandubois: I disagree with this. We just sign the binary with whatever is in `macos-entitlements.plist`. The fact that the current entitlements are needed for VZ is not relevant to the function name.

bats/tests/33-lima-controllers/limavm-running.bats
jandubois: I didn't want to do it either, but I don't know how we want to record that the state was broken, and has been "fixed" by a forced shutdown. I think using an event here is fine; it is a notification, and this is just a test. No actual controller logic depends on it.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: We are deleting the resource, so any status updates will be lost when the resource is done, so I don't think it is worth it updating the status anymore.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Yes, in separate PRs. I've created some additional issues.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Good point, but this will be refactored anyways once we implement the HA watcher #150

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Agreed, but the comment is misleading. We already download/copy the disk image during creation; this call to Prepare() is only creating the cidata.iso. I'll update the comment.

Of course the creation needs to move the Prepare() call to an asynchronous goroutine, but that shall be a separate PR.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: No, it can't. Maybe this is confusing because `inst` is the Lima instance here, so the result is going to be `~/.rd2/lima/0/ha.sock`.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: For now I want to keep the layout as close to upstream as possible. We may eventually move them, but we also have to move our own logs too because currently we store them in the appdata directory and not the logs directory.

Makefile
mook-as: I still think this will be really confusing when we do actual release signing. Call it `SIGN_DEV_BINARY` or something then? The focus isn't the VZ part (though it's why this exists), it's that this _isn't_ for all signing (and we _will_ need something to sign files for distribution).

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -0,0 +1,238 @@
...
+    fi
+}
+
+local_setup() {
+    skip_on_windows
+    # The guestagent binary is installed by Lima (brew install lima). This is a
+    # workaround until we bundle our own guestagent.
+    if [[ ! -f "/usr/local/share/lima/lima-guestagent.Linux-${ARCH}.gz" ]]; then
+        skip "Running tests require lima-guestagent.Linux-${ARCH}.gz (install Lima: brew install lima)"
+    fi
```
mook-as, line 44: This still falls over on Linux.
https://github.com/rancher-sandbox/rancher-desktop-daemon/actions/runs/21731626866/job/62687385233?pr=136#step:10:419

I know #151 exists, but this is still causing inadequate CI testing. Please install Lima on Linux too.

I added a comment in #151 to also remove the installation of Lima, because that's a workaround that shouldn't exist long term.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
mook-as: This was filed as #154.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
mook-as: `StopGracefully` may take a while.  But if we're punting on that and assuming all operations complete instantly, I guess we can ignore that and fix it later.

mook-as (changes requested):
Just some small stuff, since we're punting the bulk to later PRs.

mook-as (dismissed):

jandubois:
I've fought all day with CI to get this PR green again, so I have a couple extra commits now. I'm not sure all are necessary, but they all seem sensible to me, and we are green again. So I'm leaving them until/unless told otherwise. 😄

pkg/service/cmd/options/options.go
```diff
@@ -83,6 +84,9 @@ func NewOptions(ctx context.Context, rootDir string) *Options {
 	_ = utilfeature.DefaultMutableFeatureGate.OverrideDefault(features.ServiceAccountTokenNodeBindingValidation, false)
 	_ = utilfeature.DefaultMutableFeatureGate.OverrideDefault(features.ServiceAccountTokenNodeBinding, false)
 
+	// Disable apiserver identity leases because this is a single-instance control plane.
+	_ = utilfeature.DefaultMutableFeatureGate.OverrideDefault(genericfeatures.APIServerIdentity, false)
```
mook-as, line 88: I don't see any documentation on when this feature gate might go away; it's probably fine (it's been around for 5+ years), but I hope we don't have to deal with that too much in the future.

mook-as (approved):
LGTM. GitHub is having issues (again), let's try to merge this while it's still green.

PR #137, opened 2026-01-31
`rdd limavm delete`: Add `--wait` option

mook-as:
This adds an option so that `rdd limavm delete` can wait until the resource is deleted before returning.  This can be used to fix a flaky BATS tests, because the command to check that the VM was deleted was completing before the resource was actually done being deleted.


jandubois (approved):
Thanks, LGTM

PR #143, opened 2026-02-04
Implement API pass through support

mook-as:
This declares a new optional interface, `PassThroughController`, that can be supported by controllers.  If they do, they can implement HTTP handlers that will be exposed via the Kubernetes API server at `/passthrough/...` (completely independent of the rest of Kubernetes).  This can be used to implement endpoints that require websocket support, for example.

Note that this is still behind the normal Kubernetes authentication, though we do not currently do any access control.

### DIFF
--- a/bats/tests/32-app-controllers/passthrough.bats
+++ b/bats/tests/32-app-controllers/passthrough.bats
@@ -1,0 +1,31 @@
+load '../../helpers/load'
+
+local_setup_file() {
+    setup_rdd_control_plane
+}
+
+@test 'Check that passthrough controllers work' {
+    run -0 rdd ctl get --raw '/passthrough/hello/'
+    assert_line 'Hello, world!'
+    assert_line --regexp 'X-Forwarded-For = .*127\.0\.0\.1.*'
+}
+
+@test 'Check that passthrough to web socket works' {
+    # We need to use curl for websocket support; however, some versions of curl
+    # do not support ws:// URLs directly, so we check for that and skip the
+    # test if not supported.
+    run -0 curl --version
+    if ! [[ "${output} " =~ Protocols:.*\ ws ]]; then
+        skip "curl does not support websockets"
+    fi
+    run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.clusters[].cluster.server}'
+    local server_url="${output}"
+    run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.users[].user.token}'
+    local token="${output}"
+    run -0 curl --silent --verbose --header "Authorization: Bearer ${token}" --insecure \
+        "ws${server_url#http}/passthrough/websocket/"
+    assert_line 'hello from websocket'
+    assert_line --partial 'HTTP/1.1 101 Switching Protocols'
+    assert_line --partial 'Connection: Upgrade'
+    assert_line --partial 'Upgrade: websocket'
+}
--- a/bats/tests/32-app-controllers/passthrough.bats
+++ b/bats/tests/32-app-controllers/passthrough.bats
@@ -5,9 +5,10 @@ local_setup_file() {
 }
 
 @test 'Check that passthrough controllers work' {
-    run -0 rdd ctl get --raw '/passthrough/hello/'
+    run -0 rdd ctl get --raw '/passthrough/demo/hello/thing'
     assert_line 'Hello, world!'
     assert_line --regexp 'X-Forwarded-For = .*127\.0\.0\.1.*'
+    assert_line 'Path = "/thing"'
 }
 
 @test 'Check that passthrough to web socket works' {
@@ -15,15 +16,17 @@ local_setup_file() {
     # do not support ws:// URLs directly, so we check for that and skip the
     # test if not supported.
     run -0 curl --version
-    if ! [[ "${output} " =~ Protocols:.*\ ws ]]; then
+    # `ws` may be the last protocol and precede a new line, so we only test for
+    # space to the left of it.
+    if ! [[ "${output}" =~ Protocols:.*\ ws ]]; then
         skip "curl does not support WebSocket"
     fi
     run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.clusters[].cluster.server}'
-    local server_url="${output}"
+    local server_url=${output}
     run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.users[].user.token}'
-    local token="${output}"
+    local token=${output}
     run -0 curl --silent --verbose --header "Authorization: Bearer ${token}" --insecure \
-        "ws${server_url#http}/passthrough/websocket/"
+        "${server_url/http/ws}/passthrough/demo/websocket/"
     assert_line --partial 'hello from websocket'
     assert_line --partial 'HTTP/1.1 101 Switching Protocols'
     assert_line --partial 'Connection: Upgrade'
--- a/bats/tests/32-app-controllers/passthrough.bats
+++ b/bats/tests/32-app-controllers/passthrough.bats
@@ -11,12 +11,12 @@ local_setup_file() {
 }
 
 @test 'Check that passthrough to web socket works' {
-    # We need to use curl for websocket support; however, some versions of curl
+    # We need to use curl for WebSocket support; however, some versions of curl
     # do not support ws:// URLs directly, so we check for that and skip the
     # test if not supported.
     run -0 curl --version
     if ! [[ "${output} " =~ Protocols:.*\ ws ]]; then
-        skip "curl does not support websockets"
+        skip "curl does not support WebSocket"
     fi
     run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.clusters[].cluster.server}'
     local server_url="${output}"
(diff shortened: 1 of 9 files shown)


jandubois:
I only browsed this so far, but it looks cool because it is so generic. I especially like that it still goes through authentication.

For spelling I would call it a "Passthrough" instead of a "PassThrough" because I think it is a compound word, not two words.

bats/tests/32-app-controllers/passthrough.bats
```diff
@@ -0,0 +1,31 @@
...
+    assert_line 'Hello, world!'
+    assert_line --regexp 'X-Forwarded-For = .*127\.0\.0\.1.*'
+}
+
+@test 'Check that passthrough to web socket works' {
+    # We need to use curl for WebSocket support; however, some versions of curl
+    # do not support ws:// URLs directly, so we check for that and skip the
+    # test if not supported.
+    run -0 curl --version
+    if ! [[ "${output} " =~ Protocols:.*\ ws ]]; then
```
jandubois: Why do you need the trailing space on the left side if you are not testing for a trailing space on the right side?

bats/tests/32-app-controllers/passthrough.bats
```diff
@@ -0,0 +1,31 @@
...
+        skip "curl does not support WebSocket"
+    fi
+    run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.clusters[].cluster.server}'
+    local server_url="${output}"
```
jandubois: There is no need for quoting here; there is no splitting on whitespace here.

```suggestion
    local server_url=${output}
```

bats/tests/32-app-controllers/passthrough.bats
```diff
@@ -0,0 +1,31 @@
...
+    run -0 rdd ctl config view --minify --flatten --output=jsonpath='{.users[].user.token}'
+    local token="${output}"
+    run -0 curl --silent --verbose --header "Authorization: Bearer ${token}" --insecure \
+        "ws${server_url#http}/passthrough/websocket/"
```
jandubois: I find this clearer, but maybe just personal preference:

```suggestion
        "${server_url/http/ws}/passthrough/websocket/"
```

pkg/controllers/base/controller.go
```diff
@@ -52,6 +53,28 @@ type WebhookController interface {
...
+// This is intended for controllers that need to provide more complex HTTP
+// functionality that cannot be provided using custom resources, such as
+// handling WebSocket connections to stream data.
+type PassthroughController interface {
+	// GetPassthroughEndpoints returns the list of HTTP endpoints provided by
+	// this controller.  These endpoints will be exposed on the API server under
+	// the `/passthrough/<endpoint>` path; for example, an endpoint named "logs"
+	// will be accessible at `/passthrough/logs`.  The endpoint names must be
+	// unique across the API server, and must be a valid path segment (i.e.,
+	// no slashes).
```
jandubois: Give the uniqueness requirement, should we have some naming scheme to prefix each endpoint with the controller name to avoid collisions? If we do, then the example should follow that convention.

pkg/service/cmd/service.go
```diff
@@ -733,16 +734,21 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
 
 // createServerChain creates the apiserver instances connected via delegation.
 func createServerChain(config options.CompletedConfig) (*aggregatorapiserver.APIAggregator, error) {
-	// 1. Basic not found handler
+	// Basic not found handler
 	notFoundHandler := notfoundhandler.New(config.ControlPlane.Generic.Serializer, genericapifilters.NoMuxAndDiscoveryIncompleteKey)
 
+	// Mux for use with [base.PassthroughController] controllers; see the
```
jandubois: The comment says "for use with ...", but then the Mux is created and used even when the interface is not implemented?

pkg/controllers/app/demo/controller.go
```diff
@@ -48,6 +69,40 @@ func (c *controller) GetCRDData() string {
...
+	}
+	conn, err := upgrader.Upgrade(w, r, nil)
+	if err != nil {
+		log.V(5).Info("Failed to upgrade to WebSocket", "error", err)
+		http.Error(w, "Failed to upgrade to WebSocket: "+err.Error(), http.StatusBadRequest)
+		return
+	}
+	defer conn.Close()
+
+	err = conn.WriteMessage(websocket.TextMessage, []byte("\nhello from websocket\n"))
```
jandubois: Why does it need a leading newline?

bats/tests/32-app-controllers/passthrough.bats
```diff
@@ -0,0 +1,31 @@
+load '../../helpers/load'
+
+local_setup_file() {
+    setup_rdd_control_plane
```
jandubois, line 4: Should this have some comment here that the `hello` and `websocket` endpoints are provided by the `Demo` controller?

pkg/service/controllers/passthrough.go
```diff
@@ -0,0 +1,44 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		log := klog.FromContext(r.Context())
+		endpoint, _, _ := strings.Cut(strings.TrimPrefix(r.URL.Path, "/"), "/")
+
+		target, err := discovery.LookupPassthroughEndpoint(r.Context(), endpoint)
+		if err != nil {
+			log.V(5).Info("Pass through endpoint not found", "endpoint", endpoint)
+			http.NotFound(w, r)
+			return
+		}
```
jandubois, line 44: Any error returned by `LookupPassthroughEndpoint` is an internal server error. The "endpoint not found" condition is signalled by `target == ""`:

```suggestion
		if err != nil {
			log.Error(err, "Failed to lookup passthrough endpoint", "endpoint", endpoint)
			http.Error(w, "Internal server error", http.StatusInternalServerError)
			return
		}
		if target == "" {
			log.V(5).Info("Passthrough endpoint not found", "endpoint", endpoint)
			http.NotFound(w, r)
			return
		}
```

pkg/service/controllers/passthrough.go
```diff
@@ -0,0 +1,44 @@
...
+	"net/url"
+	"strings"
+
+	"k8s.io/klog/v2"
+)
+
+// NewPassthroughHandler creates a new HTTP handler that proxies requests to the
+// appropriate controller endpoint based on the discovery information.
+//
+// Note that this assume the target endpoints do not contain path segments; that
```
jandubois:
```suggestion
// Note that this assumes the target endpoints do not contain path segments; that
```

pkg/service/controllers/passthrough.go
```diff
@@ -0,0 +1,44 @@
...
+		targetURL, err := url.Parse(target)
+		if err != nil {
+			log.V(5).Info("Failed to parse pass through target URL",
+				"endpoint", endpoint, "target", target, "error", err)
+			http.Error(w, "Bad target URL", http.StatusInternalServerError)
+			return
+		}
+
+		log.V(5).Info("Proxying pass through request", "endpoint", endpoint, "target", target)
+		httputil.NewSingleHostReverseProxy(targetURL).ServeHTTP(w, r)
```
jandubois, line 58: I don't know if these are expensive, but this is creating a new reverse proxy for every request. Wouldn't it make sense to create one per target, and reuse it?

jandubois (changes requested):
Looks good. Still have some nit-picking. But I think the missing error handling for the `target == ""` case is a real issue.

bats/tests/32-app-controllers/passthrough.bats
mook-as: … Oh, I just missed it, I should fix that. I even added that space with `"${output} "`… though that probably actually contains `…http https … ws wss\n `, so I should just drop the extra space, and add a comment instead.

bats/tests/32-app-controllers/passthrough.bats
mook-as: Will add. I wonder if I should add an (external only) `test` controller for stuff like this?

bats/tests/32-app-controllers/passthrough.bats
mook-as: That _does_ look better.

pkg/controllers/app/demo/controller.go
mook-as: It was failing on Windows because I had `assert_line` and it came out as `} hello from…`. But now that I use `--partial` I should just drop that.

pkg/controllers/base/controller.go
mook-as: Hmm, maybe we should just prefix the controller name since that's already a thing? `demo-hello`, `demo-websocket`…

Or even just make it `/passthrough/demo/hello/`? Yeah, I think I'll go for that instead if that seems fine.

pkg/service/cmd/service.go
mook-as: I guess I should rephrase that, then. The mux is always created because at this point we haven't even looked at the list of controllers yet.

pkg/service/controllers/passthrough.go
mook-as: I checked the implementation, and `NewSingle…` just sets some variables. But sure, I can implement caching; I just didn't want to think about updating things if an external controller gets restarted (but I probably wouldn't implement that for now).

jandubois: Don't bother making any changes if creating a proxy is cheap. I just didn't know.

bats/tests/32-app-controllers/passthrough.bats
mook-as: Now that I do `run -0 rdd ctl get --raw '/passthrough/demo/hello/'` etc instead, I did not end up adding the comment, because it's now obvious in the URL.

jandubois (approved):
Thanks, LGTM

PR #144, opened 2026-02-04
Add `rdd lima shell` command

jandubois:
Reuse Lima's `sshutil` package to implement shell access to running VMs. The command validates the VM exists in the API server, sets `LIMA_HOME` for the correct RDD instance, then uses Lima's SSH utilities to connect.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -116,6 +116,12 @@ assert_limavm_not_exists() {
     run ! rdd ctl get limavm "${name}" --namespace "${NAMESPACE}"
 }
 
+assert_shell_succeeds() {
+    local name=$1
+    shift
+    rdd lima shell "${name}" "$@"
+}
+
 @test "create source template ConfigMap for running tests" {
     rdd ctl create configmap "alpine-source" --namespace "${NAMESPACE}" --from-literal="template=${ALPINE_TEMPLATE}"
     run -0 rdd ctl get configmap "alpine-source" --namespace "${NAMESPACE}" --output jsonpath='{.data.template}'
@@ -172,6 +178,36 @@ assert_limavm_not_exists() {
     # The hostagent writes its first stdout event only after the VM's SSH port
     # becomes accessible, which can lag behind the InstanceRunning condition.
     try --max 60 --delay 5 -- assert_stdout_logs_contain "${VM_NAME}" "sshLocalPort"
+}
+
+@test "shell executes command in running VM" {
+    # SSH inside the guest may not be ready immediately after InstanceRunning=True
+    try --max 12 --delay 5 -- assert_shell_succeeds "${VM_NAME}" uname -s
+    assert_line "Linux"
+}
+
+@test "shell fails when VM is not running" {
+    # Create a stopped VM to test shell error
+    rdd ctl apply -f - <<EOF
+apiVersion: lima.rancherdesktop.io/v1alpha1
+kind: LimaVM
+metadata:
+  name: stopped-vm
+  namespace: ${NAMESPACE}
+spec:
+  templateRef:
+    name: alpine-source
+    namespace: ${NAMESPACE}
+  running: false
+EOF
+    # Wait for instance to be created
+    try --max 120 --delay 1 -- assert_instance_created_condition "stopped-vm" "True"
+
+    run -1 rdd lima shell "stopped-vm"
+    assert_output --partial "is not running"
+
+    # Cleanup
+    rdd ctl delete limavm "stopped-vm" --namespace "${NAMESPACE}" --grace-period=0
 }
 
 @test "stop the VM by setting running=false" {
--- a/cmd/rdd/limavm_shell.go
+++ b/cmd/rdd/limavm_shell.go
@@ -1,0 +1,191 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+// SPDX-FileCopyrightText: The Lima Authors
+
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"strconv"
+	"strings"
+
+	"al.essio.dev/pkg/shellescape"
+	"github.com/coreos/go-semver/semver"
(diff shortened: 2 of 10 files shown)


bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -157,6 +157,39 @@ assert_limavm_not_exists() {
...
+  name: stopped-vm
+  namespace: ${NAMESPACE}
+spec:
+  templateRef:
+    name: alpine-source
+    namespace: ${NAMESPACE}
+  running: false
+EOF
+    # Wait for instance to be created
+    try --max 120 --delay 1 -- bash -c "
```
mook-as: The escaping here is ugly, but I suspect `bash -c '…'"${NAMESPACE}"'…` would be even uglier…

cmd/rdd/limavm_shell.go
```diff
@@ -0,0 +1,191 @@
...
+	// Get the Lima instance from the store
+	inst, err := store.Inspect(ctx, args[0])
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("instance %q does not exist on disk", args[0])
+		}
+		return err
+	}
+	if inst.Config == nil {
+		if len(inst.Errors) > 0 {
```
mook-as: If `inst.Config` is non-nil, but `inst.Errors` is not empty, what should happen? (I think you might have mentioned that already, but I'm not sure…)

If that should never exist, then we should move the `if len(inst.Errors) > 0` check above the `inst.Config == nil` check.

cmd/rdd/limavm_shell.go
```diff
@@ -0,0 +1,191 @@
...
+	workDir, err := cmd.Flags().GetString("workdir")
+	if err != nil {
+		return err
+	}
+	if workDir != "" {
+		changeDirCmd = fmt.Sprintf("cd %s || exit 1", shellescape.Quote(workDir))
+	} else if len(inst.Config.Mounts) > 0 {
+		hostCurrentDir, err := os.Getwd()
+		if err == nil {
+			changeDirCmd = fmt.Sprintf("cd %s", shellescape.Quote(hostCurrentDir))
```
mook-as, line 94: Ugh, I find lima's automatic directory changing to be really annoying. Probably because I mostly need to do system-wide things, not jump into lima to run `nerdctl build` or whatever.

Unless everybody else hates it too, though, I'm not sure we should change.

mook-as (dismissed):
I suspect we'll eventually just use `wsl -d …` for Windows, instead of this weird SSH dance for no reason.

I hate the lima automatic-directory-changing thing, but I assume you're copying it to stay closer to upstream.

bats/tests/33-lima-controllers/limavm-running.bats
jandubois: Oops, sorry, that comes from AI, I forgot to rewrite it as an assert_* function.

cmd/rdd/limavm_shell.go
jandubois: Yes, we should. The whole file is basically just a copy of `cmd/limactl/shell.go` with minimal changes.

cmd/rdd/limavm_shell.go
jandubois: I don't know; it all depends on the use case. For `limactl` it does make sense; it is just like `wsl` which also switches to your current directory.

We don't need to be compatible with the `limactl` subcommand, and for now it is just a developer command. I would maybe wait a bit to see what turns out to be most convenient for us. Maybe we make it a setting on the resource: ssh as regular user or root, chdir to VM home directory, or cwd on host, etc. But seems like feature creep at this point.

mook-as (approved):

cmd/rdd/limavm_shell.go
```diff
@@ -0,0 +1,191 @@
...
+
+	shellCmd.Flags().SetInterspersed(false)
+	shellCmd.Flags().String("shell", "", "Shell interpreter, e.g. /bin/bash")
+	shellCmd.Flags().String("workdir", "", "Working directory")
+
+	return shellCmd
+}
+
+func limaVMShellAction(cmd *cobra.Command, args []string) error {
+	logrus.SetLevel(logrus.InfoLevel)
```
Nino-K, line 47: This overrides global logging level unconditionally, even if the caller configured debug or trace. Is that intentional?

cmd/rdd/limavm_shell.go
```diff
@@ -0,0 +1,191 @@
...
+		return err
+	}
+	if shell == "" {
+		shell = `"$SHELL"`
+	} else {
+		shell = shellescape.Quote(shell)
+	}
+
+	// Build script
+	script := fmt.Sprintf("%s ; exec %s --login", changeDirCmd, shell)
```
Nino-K, line 126: If `changeDirCmd` ends up being `false`, does it mean we still proceed to `exec shell --login`?

cmd/rdd/limavm_shell.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
+
+	// Determine shell
+	shell, err := cmd.Flags().GetString("shell")
+	if err != nil {
+		return err
+	}
+	if shell == "" {
+		shell = `"$SHELL"`
```
Nino-K, line 120: I think in Alpine, `$SHELL` might be unset, just clarifying.

cmd/rdd/limavm_shell.go
jandubois: Yes, this is intentional because we don't want to get any debug output mixed in when we expect only output from the shell.

cmd/rdd/limavm_shell.go
jandubois: Yes, the script is not executed with `errexit`. This also comes directly from Lima, so we know it works there.

In hindsight we should probably set `changeDirCmd` to `true` in that case, but it doesn't really matter. So for now we keep it in sync with upstream.

cmd/rdd/limavm_shell.go
jandubois: This code has been battle-tested in Lima, so I know it doesn't fail on Alpine.

Variable seems to be set also:

```console
❯ limactl shell alpine-iso printenv SHELL
/bin/ash
```

Nino-K (approved):

PR #146, opened 2026-02-04
Replace CustomDefaulter/CustomValidator

### LINKED ISSUE #125: Migrate from CustomValidator/CustomDefaulter to typed interfaces
`controller-runtime` 0.23.0 deprecated `CustomValidator` and `CustomDefaulter`. These are now type aliases for `Validator[runtime.Object]` and `Defaulter[runtime.Object]`. The typed interfaces `Validator[T]` and `Defaulter[T]` provide compile-time type safety and eliminate runtime type assertions.

### Current state

```go
var _ admission.CustomValidator = &validator{}

func (v *validator) ValidateCreate(ctx context.Context, obj runtime.Object) (Warnings, error) {
    notary, ok := obj.(*v1alpha1.Notary)
    if !ok {
        return nil, fmt.Errorf("expected Notary but got %T", obj)
    }
    // ...
}
```

### Target state

```go
var _ admission.Validator[*v1alpha1.Notary] = &validator{}

func (v *validator) ValidateCreate(ctx context.Context, notary *v1alpha1.Notary) (Warnings, error) {
    // No type assertion needed
}
```

Scope:
- `pkg/controllers/base/webhook.go`: Make WebhookConfig and SetupWebhookForResource generic
- `pkg/controllers/lima/limavm/controller.go`: Update defaulter and ConfigMapValidator
- `pkg/controllers/rdd/notary/controller.go`: Update validator

### Considerations

The `WebhookConfig` struct stores validators and defaulters for any resource type. Making it generic requires either parameterizing the entire webhook setup or accepting any at the config level. Evaluate whether the type safety benefit justifies the architectural change.

---

Note: Issue is mostly AI generated.

mook-as:
The admission webhook types have been deprecated, and replaced with generic versions that specify the type of the object.  To make this work in our code, `WebhookManager` has been turned into an interface so that we can have typed versions for the implementation.

The main fallout for the controllers is that `GetWebhookManagers` no longer return a slice of pointers to struct, but a slice of interfaces instead.

Fixes #125

### DIFF
--- a/pkg/controllers/base/controller.go
+++ b/pkg/controllers/base/controller.go
@@ -49,7 +49,7 @@ type WebhookController interface {
 
 	// GetWebhookManagers returns all WebhookManagers for parallel setup.
 	// Returns nil or empty slice if the controller doesn't use webhooks.
-	GetWebhookManagers() []*WebhookManager
+	GetWebhookManagers() []WebhookManager
 }
 
 // Registry holds all registered controllers.
--- a/pkg/controllers/base/webhook.go
+++ b/pkg/controllers/base/webhook.go
@@ -15,6 +15,7 @@ import (
 
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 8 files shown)


jandubois (approved):
Thanks, LGTM

PR #148, opened 2026-02-05
base: Replace deprecated `c.Patch(..., client.Apply)`

### LINKED ISSUE #126: Migrate from client.Patch with client.Apply to client.Apply method
`controller-runtime` 0.23.0 deprecated the `client.Apply` patch type. The new approach uses `client.Apply()` with `ApplyConfiguration` builder types from `k8s.io/client-go/applyconfigurations`.

### Current state

```go
webhook := &admissionregistrationv1.ValidatingWebhookConfiguration{
    ObjectMeta: metav1.ObjectMeta{Name: name},
    Webhooks:   []admissionregistrationv1.ValidatingWebhook{...},
}
c.Patch(ctx, webhook, client.Apply, client.ForceOwnership, client.FieldOwner("rdd-webhook-manager"))
```

### Target state

```go
import admregv1ac "k8s.io/client-go/applyconfigurations/admissionregistration/v1"

webhook := admregv1ac.ValidatingWebhookConfiguration(name).
    WithWebhooks(admregv1ac.ValidatingWebhook().
        WithName(webhookName).
        WithClientConfig(...).
        WithRules(...))
c.Apply(ctx, webhook, client.ForceOwnership, client.FieldOwner("rdd-webhook-manager"))
```

Scope:
- `pkg/controllers/base/webhook.go`: Refactor `createValidatingWebhook` and `createMutatingWebhook`

### Considerations

`ApplyConfiguration` types use a builder pattern that differs from direct struct construction. The refactor is straightforward but verbose. The current Patch-based approach works and has no announced removal timeline.

---

Note: Issue is mostly AI generated.

mook-as:
Use `c.Apply()` instead; that involves changing to using the `admissionregistrationv1apply` APIs instead.

I don't really love this change, but the other easy path would be using `client.MergeFrom()`, which doesn't work if the object didn't already exist.  Those function signatures look _ridiculous_ now…

Fixes #126

### DIFF
--- a/pkg/controllers/base/webhook.go
+++ b/pkg/controllers/base/webhook.go
@@ -288,15 +288,19 @@ func (wm *webhookManagerImpl[T]) createWebhookConfiguration() error {
 
 	switch wm.webhookType {
 	case MutatingWebhook:
-		return wm.createMutatingWebhook(ctx, c, webhookURL, clientConfig, rules, failurePolicy, sideEffects)
+		err = wm.createMutatingWebhook(ctx, c, clientConfig, rules, failurePolicy, sideEffects)
 	case ValidatingWebhook:
-		return wm.createValidatingWebhook(ctx, c, webhookURL, clientConfig, rules, failurePolicy, sideEffects)
+		err = wm.createValidatingWebhook(ctx, c, clientConfig, rules, failurePolicy, sideEffects)
 	default:
-		return fmt.Errorf("invalid webhook type: %s", wm.webhookType)
+		err = fmt.Errorf("invalid webhook type: %s", wm.webhookType)
 	}
+	if err == nil {
+		klog.Infof("Successfully applied webhook configuration for %s", webhookURL)
+	}
+	return err
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/base/webhook.go
```diff
@@ -299,63 +296,43 @@ func (wm *webhookManagerImpl[T]) createWebhookConfiguration() error {
...
-				ClientConfig:            clientConfig,
-				Rules:                   rules,
-				FailurePolicy:           &failurePolicy,
-				SideEffects:             &sideEffects,
-				AdmissionReviewVersions: []string{"v1", "v1beta1"},
-				ObjectSelector:          wm.config.ObjectSelector,
-			},
-		},
-	}
+func (wm *webhookManagerImpl[T]) createMutatingWebhook(ctx context.Context, c client.Client, webhookURL string, clientConfig *admissionregistrationv1apply.WebhookClientConfigApplyConfiguration, rules []*admissionregistrationv1apply.RuleWithOperationsApplyConfiguration, failurePolicy admissionregistrationv1.FailurePolicyType, sideEffects admissionregistrationv1.SideEffectClass) error {
```
jandubois: I know this is just a refactoring, but we can probably drop the `webhookURL` parameter and use `*clientConfig.URL` instead. I think they are the same thing.

jandubois (dismissed):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

PR #156, opened 2026-02-05
Replace pulumi dependency with vendored nxadm/tail

### LINKED ISSUE #119: Remove Pulumi dependency
We only use it to get its version of `tail.go` that has 2 PRs merged because `nxadm/tail` seems abandoned. It still relies on `nxadm/tail` for the `nxadm/tail/util` and `nxadm/tail/watch` packages.

By including the Pulumi `tail.go` file (plus platform-specific files) we could drop about 60 dependencies from our `go.mod` (if I checked this correctly) and would no longer get the frequent dependabot updates for Pulumi, which never touch `tail.go` anyways.

jandubois:
Pulumi vendors `github.com/nxadm/tail` with patches from nxadm/tail#70 and nxadm/tail#71. Copy that patched code into `pkg/util/nxadmtail/` and drop the `pulumi/pulumi/sdk/v3` dependency.

The vendored files retain their original MIT license. A NOTICE file documents the license difference; MIT is compatible with the project's Apache-2.0 license.

Fixes #119


pkg/util/nxadmtail/LICENSE
```diff
@@ -0,0 +1,21 @@
+# The MIT License (MIT)
+
+# © Copyright 2015 Hewlett Packard Enterprise Development LP
+Copyright (c) 2014 ActiveState
```
mook-as, line 4: I'd have thought that there'd be some added copyright from nxadm or whoever forked it, but that's the upstream license file so I guess that's fine…

mook-as (approved):
I'd have left the files in `pkg/util/tail` (instead of creating `nxadmtail`), but I can understand why this is the case too.

PR #160, opened 2026-02-09
Embed Lima guest agent in rdd binary

### LINKED ISSUE #151: Build and embed the guestagent
Once embedded, we can extract it as needed to the filesystem when we are starting a LimaVM instance.

The long term goal is to be able to provide e.g. an `io.Reader` to `instance.Start()` instead of a filename, but that requires upstream changes.

jandubois:
Build the guest agent from our existing Lima Go dependency and embed the compressed binary via go:embed. This eliminates the requirement to have Lima installed, both for developers and in CI.

The Makefile cross-compiles the guest agent for Linux (matching the host GOARCH), compresses it with gzip, and places it where go:embed picks it up. At runtime, the hostagent extracts it for provisioning.

A blank import in deps_linux.go ensures go mod tidy retains the guest agent's Linux-only transitive dependencies in go.sum.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,8 @@ LDFLAGS := \
 
 TAGS := sqlite_omit_load_extension
 
-GUESTAGENT_BINARY := pkg/guestagent/lima-guestagent.gz
+GUESTAGENT_BINARY := pkg/guestagent/lima-guestagent
+GUESTAGENT_GZ := $(GUESTAGENT_BINARY).gz
 
 default: build-rdd
 .PHONY: default
@@ -66,11 +67,13 @@ GOLANG_SOURCES := $(shell find . -name '*.go') go.mod go.sum
 $(GUESTAGENT_BINARY): go.mod go.sum
 	WSLENV=${WSLENV}:CGO_ENABLED:GOOS \
 	CGO_ENABLED=0 GOOS=linux \
-	go$(EXE) build -ldflags="-s -w" -o pkg/guestagent/lima-guestagent \
+	go$(EXE) build -ldflags="-s -w" -o $@ \
 		github.com/lima-vm/lima/v2/cmd/lima-guestagent
-	gzip --force pkg/guestagent/lima-guestagent
 
-bin/rdd$(EXE): $(GOLANG_SOURCES) $(GUESTAGENT_BINARY)
+$(GUESTAGENT_GZ): $(GUESTAGENT_BINARY)
+	gzip --to-stdout $< > $@
+
+bin/rdd$(EXE): $(GOLANG_SOURCES) $(GUESTAGENT_GZ)
 	WSLENV=${WSLENV}:CGO_CFLAGS:CGO_ENABLED \
 	CGO_CFLAGS="-DSQLITE_ENABLE_DBSTAT_VTAB=1 -DSQLITE_USE_ALLOCA=1" CGO_ENABLED=1 \
 	go$(EXE) build -tags="$(TAGS)" -gcflags="all=${GCFLAGS}" -ldflags="$(LDFLAGS)" -o $@ ./cmd/rdd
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -60,7 +62,15 @@ endif
...
 GOLANG_SOURCES := $(shell find . -name '*.go') go.mod go.sum
 
-bin/rdd$(EXE): $(GOLANG_SOURCES)
+# Build the Lima guest agent from our Go dependency and compress it for embedding.
+$(GUESTAGENT_BINARY): go.mod go.sum
+	WSLENV=${WSLENV}:CGO_ENABLED:GOOS \
+	CGO_ENABLED=0 GOOS=linux \
+	go$(EXE) build -ldflags="-s -w" -o pkg/guestagent/lima-guestagent \
+		github.com/lima-vm/lima/v2/cmd/lima-guestagent
+	gzip --force pkg/guestagent/lima-guestagent
```
mook-as: Other options is to have

```make
$(GUESTAGENT_BINARY).gz: $(GUESTAGENT_BINARY)
	gzip --force --to-stdout $< > $@
```
This is especially confusing because we don't actually require the guest agent binary; we require the gzipped version for `embed`.  So as currently written, if the guest agent binary exists but the gzipped version was deleted, the build would fail.

Why does `gzip` need `--force` again? It creates `pkg/guestagent/lima-guestagent.gz`, so there shouldn't be an overwrite involved, right?

In general, using `$@` more (e.g. for `go build -o …`) would be good.

Makefile
```diff
@@ -126,15 +136,15 @@ run: bin/rdd$(EXE)
...
+test: $(GOLANG_SOURCES) $(GUESTAGENT_BINARY)
 	go$(EXE) test ./...
 .PHONY: test
 
 lint-bats:
 	$(MAKE) -C bats lint
 .PHONY: lint-bats
 
-lint-rdd:
+lint-rdd: $(GUESTAGENT_BINARY)
```
mook-as: Ah, linting requires the guest agent binary so that the `embed` directive can find the file to embed?

pkg/guestagent/guestagent.go
```diff
@@ -0,0 +1,38 @@
...
+	if _, err := f.Write(guestAgentGZ); err != nil {
+		f.Close()
+		os.Remove(f.Name())
+		return "", nil, fmt.Errorf("writing guest agent temp file: %w", err)
+	}
+	if err := f.Close(); err != nil {
+		os.Remove(f.Name())
+		return "", nil, fmt.Errorf("closing guest agent temp file: %w", err)
+	}
+	return f.Name(), func() { os.Remove(f.Name()) }, nil
```
mook-as, line 37: Do we need to worry about the return value of `os.Remove`? (I thought the linter would have complained, actually.)

But then, I guess if the linter is fine with it it's probably fine.

mook-as (changes requested):
Basically fine, just need to fix the case where `lima-guestagent.gz` was deleted and `make` wouldn't rebuild it.

pkg/guestagent/guestagent.go
jandubois: The linter is configured to ignore `os.Remove()` error checking: https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/97b098a472d0501641a84d0adf42d479a12370ad/.golangci.yml#L132

See https://golangci-lint.run/docs/linters/false-positives/#preset-std-error-handling

```
(?i)Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*print(f|ln)?|os\.(Un)?Setenv). is not checked
```

mook-as (dismissed):

jandubois:
Sorry, had to rebase on `main` to resolve merge conflict; please approve again!

mook-as (approved):

PR #162, opened 2026-02-09
Replace hand-rolled condition helpers with meta.SetStatusCondition

jandubois:
The three controller setCondition methods reimplemented standard library behavior from k8s.io/apimachinery/pkg/api/meta. Replace the bodies with meta.SetStatusCondition while keeping the wrapper methods for event recording.

Delete hasCondition and conditionExists from the limavm controller, replacing call sites with meta.IsStatusConditionPresentAndEqual and meta.FindStatusCondition. Move hasConditionWithReason to the base package as a shared helper since inlining it would produce hard-to-read triple-negation at call sites.

Delete the unit tests that tested standard library behavior.


pkg/controllers/base/conditions.go
```diff
@@ -0,0 +1,17 @@
...
+
+package base
+
+import (
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// HasConditionWithReason reports whether conditions contains a condition
+// of the given type with the given status and reason.
```
mook-as, line 13: Ah, so this doesn't overlap with [`IsStatusConditionPresentAndEqual`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#IsStatusConditionPresentAndEqual) because we also need to match the reason.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -165,7 +165,7 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// Handle instances with a preparing sentinel file. The sentinel indicates that
 	// a previous reconcile started preparation but didn't complete successfully.
 	if hasSentinel(limaVM.Name) {
-		if r.hasCondition(&limaVM, ConditionInstanceCreated, metav1.ConditionTrue) {
+		if apimeta.IsStatusConditionPresentAndEqual(limaVM.Status.Conditions, ConditionInstanceCreated, metav1.ConditionTrue) {
```
mook-as:
```suggestion
		if apimeta.IsStatusConditionTrue(limaVM.Status.Conditions, ConditionInstanceCreated) {
```

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -234,7 +234,7 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	// Instance already created - proceed to handle running state
-	if r.hasCondition(&limaVM, ConditionInstanceCreated, metav1.ConditionTrue) {
+	if apimeta.IsStatusConditionPresentAndEqual(limaVM.Status.Conditions, ConditionInstanceCreated, metav1.ConditionTrue) {
```
mook-as:
```suggestion
	if apimeta.IsStatusConditionTrue(limaVM.Status.Conditions, ConditionInstanceCreated, metav1.ConditionTrue) {
```

mook-as (changes requested):

jandubois:
Since it was just 2 lines, I've amended the previous commit.

mook-as (approved):

PR #163, opened 2026-02-10
Improve logging infrastructure

jandubois:
Improve CI log collection by giving each RDD instance a dedicated log directory and numbered log files that survive instance deletion.

- Add `instance.LogDir()` for OS-specific log directories separate from instance data, so logs persist when an instance is deleted
- Add `rdd svc paths` command that prints all instance paths in table, JSON, or shell-export format; BATS helpers now source paths from this command instead of duplicating per-platform logic
- Add `pkg/util/logfile` package: each restart creates the next sequentially numbered file (`{name}.{N}.log`) and updates a symlink (`{name}.log`); old files beyond the retention count are pruned
- Use numbered log files for the RDD service and hostagent
- Collect the entire log directory per instance in CI artifacts instead of a single log file
- Run all BATS test suites even when one fails, so CI always produces complete logs

### DIFF
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -13,7 +13,6 @@ pulumi
 Trivy
 
 # golangci-lint & its linter / formatter names
-golangci
 bodyclose
 copyloopvar
 depguard
@@ -26,6 +25,7 @@ gocritic
 godot
 gofmt
 gofumpt
+golangci
 govet
 ineffassign
 intrange
@@ -35,14 +35,14 @@ perfsprint
 usetesting
 
 # golangci-lint options
+errorf
+gotest
 localmodule
+mgechev
 nolint
 nolintlint
-gotest
-unlabel
-errorf
 strconcat
-mgechev
+unlabel
 
 # awk
 RLENGTH
@@ -50,6 +50,7 @@ RSTART
 
 # bash
 neq
+nullglob
 
 # bats
 batslib
@@ -68,10 +69,11 @@ DSQLITE
 endef
 ifeq
 ifneq
+MAKEFLAGS
 notdir
 patsubst
 VTAB
 
 # check-spelling
-hunspell
 dic
+hunspell
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -49,7 +49,6 @@ RLENGTH
 RSTART
 
 # bash
-euo
 neq
 
 # bats
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -49,6 +49,7 @@ RLENGTH
 RSTART
 
 # bash
+euo
 neq
 
 # bats
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -156,7 +156,7 @@ jobs:
         systemctl status systemd-binfmt.service || true
       continue-on-error: true
 
-    - run: make -C bats
+    - run: make -C bats --keep-going
       shell: run-in-wsl {0}
       env:
         RDD_TRACE: true
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -164,7 +164,7 @@ jobs:
     - name: Collect RDD logs
       if: always()
       shell: run-in-wsl {0}
-      run: scripts/collect-logs.sh
+      run: scripts/collect-bats-logs.sh rdd-logs
 
     - name: Upload RDD service logs
       if: always()
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -161,14 +161,10 @@ jobs:
       env:
         RDD_TRACE: true
 
-    - name: Collect RDD service logs
+    - name: Collect RDD logs
       if: always()
       shell: run-in-wsl {0}
-      run: |
-        mkdir -p rdd-logs
-        for instance in $(make --no-print-directory -C bats bats-instances); do
-          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" || true
-        done
+      run: scripts/collect-logs.sh
 
     - name: Upload RDD service logs
       if: always()
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -67,23 +67,12 @@ bats-containers: check-bats build-rdd build-mock-controller
 	RDD_INSTANCE=bats-containers-controller $(BATS_CORE) tests/34-containers-controllers/
 .PHONY: bats-containers
 
-# Run all controller tests, continuing past failures.
-bats-controllers:
-	@rc=0; \
-	$(MAKE) bats-builtin || rc=1; \
-	$(MAKE) bats-rdd || rc=1; \
-	$(MAKE) bats-app || rc=1; \
-	$(MAKE) bats-lima || rc=1; \
-	$(MAKE) bats-containers || rc=1; \
-	exit $$rc
+# Run all controller tests.
+bats-controllers: bats-builtin bats-rdd bats-app bats-lima bats-containers
 .PHONY: bats-controllers
 
-# Run all BATS tests, continuing past failures so all suites produce logs.
-bats-all: bats-helpers bats-cli
-	@rc=0; \
-	$(MAKE) bats-service || rc=1; \
-	$(MAKE) bats-controllers || rc=1; \
-	exit $$rc
+# Run all BATS tests.
+bats-all: bats-helpers bats-cli bats-service bats-controllers
 .PHONY: bats-all
 
 # Run BATS tests with timeout to prevent hanging
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -29,7 +29,7 @@ export TMPDIR := $(shell wslpath -u $$(cmd.exe /c "echo %TEMP%"))
 endif
 
 # Run BATS tests for BATS helpers
-bats-helpers: check-bats
+bats-helpers: check-bats build-rdd
 	$(BATS_CORE) helpers/
 .PHONY: bats-helpers
 
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -67,12 +67,23 @@ bats-containers: check-bats build-rdd build-mock-controller
 	RDD_INSTANCE=bats-containers-controller $(BATS_CORE) tests/34-containers-controllers/
 .PHONY: bats-containers
 
-# Run all controller tests
-bats-controllers: bats-builtin bats-rdd bats-app bats-lima bats-containers
+# Run all controller tests, continuing past failures.
+bats-controllers:
+	@rc=0; \
+	$(MAKE) bats-builtin || rc=1; \
+	$(MAKE) bats-rdd || rc=1; \
+	$(MAKE) bats-app || rc=1; \
+	$(MAKE) bats-lima || rc=1; \
+	$(MAKE) bats-containers || rc=1; \
+	exit $$rc
 .PHONY: bats-controllers
 
-# Run all BATS tests
-bats-all: bats-helpers bats-cli bats-service bats-controllers
+# Run all BATS tests, continuing past failures so all suites produce logs.
+bats-all: bats-helpers bats-cli
+	@rc=0; \
+	$(MAKE) bats-service || rc=1; \
+	$(MAKE) bats-controllers || rc=1; \
+	exit $$rc
 .PHONY: bats-all
 
 # Run BATS tests with timeout to prevent hanging
--- a/bats/helpers/paths.bash
+++ b/bats/helpers/paths.bash
@@ -18,7 +18,7 @@ win_to_wsl_exports() {
 # Get instance paths from rdd (single source of truth).
 # shellcheck disable=SC1090
 if is_windows; then
-    source <("${PATH_REPO_ROOT}/bin/rdd.exe" svc paths --shell | win_to_wsl_exports)
+    source <("${PATH_REPO_ROOT}/bin/rdd.exe" svc paths --output=shell | win_to_wsl_exports)
 else
-    source <("${PATH_REPO_ROOT}/bin/rdd" svc paths --shell)
+    source <("${PATH_REPO_ROOT}/bin/rdd" svc paths --output=shell)
 fi
--- a/bats/helpers/paths.bash
+++ b/bats/helpers/paths.bash
@@ -6,44 +6,25 @@ inside_repo_clone() {
     [[ -d "${PATH_REPO_ROOT}/pkg/rancher-desktop-daemon" ]]
 }
 
-if is_macos; then
-    PATH_APP_HOME="${HOME}/Library/Application Support/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CONFIG="${HOME}/Library/Preferences/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CACHE="${HOME}/Library/Caches/rancher-desktop-${RDD_INSTANCE}"
-    PATH_LOGS="${PATH_APP_HOME}"
-fi
-
-if is_linux; then
-    PATH_APP_HOME="${HOME}/.local/share/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CONFIG="${HOME}/.config/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CACHE="${HOME}/.cache/rancher-desktop-${RDD_INSTANCE}"
-    PATH_LOGS="${PATH_APP_HOME}"
-fi
-
-# Get the WSL (Unix) path for a Windows shell folder id; for a list of valid ids,
-# see https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder
-wslpath_from_win32_folder_id() {
-    local folder_id=$1
-    local windows_path
-    windows_path=$(powershell.exe -Command "[System.Environment]::GetFolderPath(${folder_id})")
-    local unix_path
-    unix_path=$(wslpath -u "${windows_path}")
-    tr -d "\r" <<<"${unix_path}"
+# Convert 'export KEY="C:\win\path"' to 'export KEY="/mnt/c/wsl/path"'.
+win_to_wsl_exports() {
+    tr -d "\r" | while IFS= read -r line; do
+        if [[ "${line}" =~ ^(export\ [A-Z_]+=)\"(.+)\"$ ]]; then
+            printf '%s"%s"\n' "${BASH_REMATCH[1]}" "$(wslpath -u "${BASH_REMATCH[2]}")"
+        fi
+    done
 }
 
+# Get instance paths from rdd (single source of truth).
+# shellcheck disable=SC1090
 if is_windows; then
-    LOCALAPPDATA="$(wslpath_from_win32_folder_id 28)"
-    PROGRAMFILES="$(wslpath_from_win32_folder_id 38)"
-
-    PATH_APP_HOME="${LOCALAPPDATA}/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CONFIG="${LOCALAPPDATA}/rancher-desktop-${RDD_INSTANCE}"
-    PATH_CACHE="${PATH_APP_HOME}/cache"
-    PATH_LOGS="${PATH_APP_HOME}"
-    PATH_DISTRO="${PATH_APP_HOME}/distro"
-    PATH_DISTRO_DATA="${PATH_APP_HOME}/distro-data"
+    source <("${PATH_REPO_ROOT}/bin/rdd.exe" svc paths --shell | win_to_wsl_exports)
+else
+    source <("${PATH_REPO_ROOT}/bin/rdd" svc paths --shell)
 fi
 
-# PATH_RD is the "short directory" (e.g., ~/.rd2) as documented in docs/design/cmd_service.md.
-# LIMA_HOME uses PATH_RD instead of PATH_APP_HOME because of socket name length constraints.
-PATH_RD="${HOME}/.rd${RDD_INSTANCE}"
-LIMA_HOME="${PATH_RD}/lima"
+PATH_APP_HOME="${RDD_DIR}"
+PATH_ARGS_FILE="${RDD_ARGS_FILE}"
+PATH_LOGS="${RDD_LOG_DIR}"
+PATH_RD="${RDD_SHORT_DIR}"
+LIMA_HOME="${RDD_LIMA_HOME}"
--- a/bats/helpers/paths.bash
+++ b/bats/helpers/paths.bash
@@ -22,9 +22,3 @@ if is_windows; then
 else
     source <("${PATH_REPO_ROOT}/bin/rdd" svc paths --shell)
 fi
-
-PATH_APP_HOME="${RDD_DIR}"
-PATH_ARGS_FILE="${RDD_ARGS_FILE}"
-PATH_LOGS="${RDD_LOG_DIR}"
-PATH_RD="${RDD_SHORT_DIR}"
-LIMA_HOME="${RDD_LIMA_HOME}"
--- a/bats/tests/10-cli/5-paths.bats
+++ b/bats/tests/10-cli/5-paths.bats
@@ -1,0 +1,41 @@
+load '../../helpers/load'
+
+ALL_KEYS="dir log_dir short_dir lima_home tls_dir kubeconfig pid_file args_file"
+
+@test 'rdd svc paths prints all keys in table format' {
+    run -0 rdd svc paths
+    for key in ${ALL_KEYS}; do
+        assert_line --regexp "^${key} "
+    done
+}
+
+@test 'rdd svc paths --json produces valid JSON with all keys' {
+    run -0 rdd svc paths --json
+    local json="${output}"
+    for key in ${ALL_KEYS}; do
+        jq --exit-status --arg k "${key}" 'has($k)' <<<"${json}"
+    done
+}
+
+@test 'rdd svc paths --shell produces shell export statements' {
+    run -0 rdd svc paths --shell
+    for key in ${ALL_KEYS}; do
+        upper_key=$(echo "${key}" | tr '[:lower:]' '[:upper:]')
+        assert_line --regexp "^export RDD_${upper_key}="
+    done
+}
+
+@test 'rdd svc paths <key> prints only the value' {
+    run -0 rdd svc paths log_dir
+    # Output should be a single line with no key prefix
+    assert_output --regexp '^(/|[A-Z]:)'
+    refute_output --regexp '^log_dir'
+    [[ "${#lines[@]}" -eq 1 ]]
+}
+
+@test 'rdd svc paths with invalid key fails and lists valid keys' {
+    run -1 rdd svc paths no_such_key
+    assert_output --partial 'unknown key'
+    assert_output --partial 'no_such_key'
+    assert_output --partial 'valid keys'
+}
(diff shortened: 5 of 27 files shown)


pkg/util/logfile/logfile.go
```diff
@@ -0,0 +1,103 @@
...
+const retentionCount = 5
+
+type numberedFile struct {
+	n    int
+	name string
+}
+
+// Create opens the next numbered log file in dir and updates the symlink.
+//
+// File naming: {name}.{N}.log, with symlink {name}.log -> {name}.{N}.log.
```
mook-as: What does that mean on Windows? I'm aware that you're only building out macOS for now, but that still means we need to make APIs that make sense for later Windows implementation.

pkg/util/logfile/logfile.go
```diff
@@ -0,0 +1,103 @@
...
+	var numberedFiles []numberedFile
+	for _, entry := range entries {
+		matches := pattern.FindStringSubmatch(entry.Name())
+		if matches == nil {
+			continue
+		}
+		n, err := strconv.Atoi(matches[1])
+		if err != nil {
+			continue
+		}
```
mook-as, line 54: How about using `sscanf`?

```golang
var n int
_, err := fmt.Sscanf(entry.Name() + "\x00", name + ".%d.log\x00", &n)
if err == nil {
	numberedFiles = append(numberedFiles, numberedFile{n: n, name: entry.Name()})
	if n > maxN {
		maxN = n
	}
}
```

Down side here is that we need to do manual end-of-input handling.

pkg/util/logfile/logfile.go
```diff
@@ -0,0 +1,103 @@
...
+	filePath := filepath.Join(dir, fileName)
+
+	f, err := os.Create(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("create log file %s: %w", filePath, err)
+	}
+
+	if header != "" {
+		if _, err := f.WriteString(header); err != nil {
+			f.Close()
```
mook-as, line 80: Do we need to delete the file? Or is leaving the file (possibly with some contents) fine?

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	for i := 1; i <= 3; i++ {
+		f, err := Create(dir, "app", true, "")
+		assert.NilError(t, err, "Create #%d", i)
+		f.Close()
+	}
+
+	// All three files should exist
+	for i := 1; i <= 3; i++ {
+		name := filepath.Join(dir, "app."+itoa(i)+".log")
```
mook-as: Isn't using `fmt.Sprintf()` easier than having a wrapper `itoa`?

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,184 @@
...
+	target, err := os.Readlink(filepath.Join(dir, "app.log"))
+	assert.NilError(t, err)
+	assert.Equal(t, target, "app.3.log")
+}
+
+func TestPruning(t *testing.T) {
+	dir := t.TempDir()
+
+	// Create 7 files with pruning enabled
+	for i := range 7 {
```
mook-as: What do you feel about `i := range retentionCount + 2`, just to make it clear that things should be pruned?

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,184 @@
...
+		f, err := Create(dir, "prune", false, "")
+		assert.NilError(t, err, "Create #%d", i+1)
+		f.Close()
+	}
+
+	// Files 1 and 2 should be pruned (7 - 5 = 2)
+	for _, n := range []int{1, 2} {
```
mook-as: Isn't old-school `for i := 1; i <= 2; i++` clearer? Especially for the other loop…

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// Manually create files with gaps: 1, 3, 5
+	for _, n := range []int{1, 3, 5} {
+		f, err := os.Create(filepath.Join(dir, "gap."+itoa(n)+".log"))
+		assert.NilError(t, err, "create gap file")
+		f.Close()
+	}
+
+	// Next file should be 6 (max=5, next=6)
+	f, err := Create(dir, "gap", false, "")
```
mook-as: Should also test pruning behaviour here. Which brings up the awkward question: should gaps mean we prune fewer files?

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	f, err := Create(dir, "test", false, "")
+	assert.NilError(t, err)
+	f.Close()
+
+	_, err = os.Stat(filepath.Join(dir, "test.1.log"))
+	assert.NilError(t, err, "expected test.1.log in nested dir")
+}
+
+func TestSymlinkFollowsTransparently(t *testing.T) {
```
mook-as: I don't understand why this is needed given that we previously tested that the symlink exists, but 🤷

mook-as: Should we test when `test.1.log` exists _as a directory_, and the behaviour around that (and whether it gets pruned, or at least return an error)?

… Probably not worth it?

scripts/collect-logs.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+# Collect RDD service and Lima hostagent logs into a single directory.
```
mook-as:
```suggestion
# Collect RDD service and Lima hostagent BATS logs into a single directory.
```

Maybe call this file `collect-bats-logs.sh` to be clearer?

scripts/collect-logs.sh
```diff
@@ -0,0 +1,70 @@
...
+
+    # Service logs (rdd.stdout, rdd.stderr)
+    if log_dir=$(resolve_path log_dir); then
+        collect_logs "$log_dir" "$dest"
+    fi
+
+    # Lima hostagent logs (ha.stdout, ha.stderr per VM)
+    if lima_home=$(resolve_path lima_home); then
+        for vm_dir in "$lima_home"/*/; do
+            [ -d "$vm_dir" ] || continue
```
mook-as: Set `nullglob` at the top?

.github/actions/spelling/expect/tools.txt
```diff
@@ -49,6 +49,7 @@ RLENGTH
 RSTART
 
 # bash
+euo
```
mook-as: We should spell it `set -o errexit -o nounset -o pipefail` instead. `euo` seems like a thing that we might accidentally trip.

.github/workflows/bats.yaml
```diff
@@ -161,14 +161,10 @@ jobs:
...
         RDD_TRACE: true
 
-    - name: Collect RDD service logs
+    - name: Collect RDD logs
       if: always()
       shell: run-in-wsl {0}
-      run: |
-        mkdir -p rdd-logs
-        for instance in $(make --no-print-directory -C bats bats-instances); do
-          RDD_INSTANCE="$instance" bin/rdd svc log > "rdd-logs/${instance}.log" || true
```
mook-as, line 170: Not this PR, but maybe we should make `rdd svc log` print the relevant logs instead, and maybe make it do the copying in the future…

.github/workflows/bats.yaml
```diff
@@ -161,14 +161,10 @@ jobs:
...
-        done
+      run: scripts/collect-logs.sh
```
mook-as: It might be good to make the output directory explicit here.

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,97 @@
...
+
+func newServicePathsCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "paths [key]",
+		Short: "Print instance paths",
+		Args:  cobra.MaximumNArgs(1),
+		RunE:  servicePathsAction,
+	}
+	command.Flags().Bool("json", false, "Output as JSON")
+	command.Flags().Bool("shell", false, "Output as shell export statements")
```
mook-as: I think `--output=[json|shell]` would be better (so we can add a `=table` that's the default).  That also means no worries about mutually-incompatible options.

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,97 @@
...
+	if asJSON && asShell {
+		return errors.New("--json and --shell are mutually exclusive")
+	}
+
+	paths := instancePaths()
+	keys := slices.Sorted(maps.Keys(paths))
+
+	// Filter to a single key if specified.
+	if len(args) == 1 {
+		key := args[0]
```
mook-as, line 65: I _think_ `rdd svc paths --key=log_dir` might be better, but I'm not as sure as I am about `--output`.

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,97 @@
...
+	switch {
+	case asJSON:
+		m := make(map[string]string, len(keys))
+		for _, key := range keys {
+			m[key] = paths[key]
+		}
+		return json.NewEncoder(w).Encode(m)
+	case asShell:
+		for _, key := range keys {
+			if _, err := fmt.Fprintf(w, "export RDD_%s=%q\n", strings.ToUpper(key), paths[key]); err != nil {
```
mook-as, line 86: I hope we won't have to think about funny paths, but I guess we can deal with that when we get there.

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,97 @@
...
+		for _, key := range keys {
+			if len(key) > maxKey {
+				maxKey = len(key)
+			}
+		}
+		for _, key := range keys {
+			if _, err := fmt.Fprintf(w, "%-*s  %s\n", maxKey, key, paths[key]); err != nil {
+				return err
+			}
+		}
```
mook-as: Consider using [`tabwriter`](https://pkg.go.dev/text/tabwriter@go1.26.0#Writer) instead.

docs/design/environment.md
```diff
@@ -0,0 +1,50 @@
...
+
+```shell
+source <(rdd svc paths --shell)
+```
+
+| Variable | Description | Example (macOS, instance `2`) |
+| --- | --- | --- |
+| `RDD_ARGS_FILE` | Saved startup arguments | `~/Library/Application Support/rancher-desktop-2/args.json` |
+| `RDD_DIR` | Service instance directory | `~/Library/Application Support/rancher-desktop-2` |
+| `RDD_KUBECONFIG` | Kubeconfig file | `~/.rd2/kube.config` |
```
mook-as:
```suggestion
| `RDD_KUBECONFIG` | Kubeconfig file | `~/.rd2/kubeconfig` |
```

… I don't actually care, just being a bit too pedantic. It's actually in `Application Support` anyway.

bats/tests/10-cli/5-paths.bats
```diff
@@ -0,0 +1,41 @@
...
+    for key in ${ALL_KEYS}; do
+        assert_line --regexp "^${key} "
+    done
+}
+
+@test 'rdd svc paths --json produces valid JSON with all keys' {
+    run -0 rdd svc paths --json
+    local json="${output}"
+    for key in ${ALL_KEYS}; do
+        jq --exit-status --arg k "${key}" 'has($k)' <<<"${json}"
```
mook-as, line 16: Just poking, don't worry about this.
```suggestion
        jq --exit-status "has(\"${key}\")" <<<"${json}"
```
Yeah, I think your original is easier to read.

bats/tests/10-cli/5-paths.bats
```diff
@@ -0,0 +1,41 @@
...
+    done
+}
+
+@test 'rdd svc paths --shell produces shell export statements' {
+    run -0 rdd svc paths --shell
+    for key in ${ALL_KEYS}; do
+        upper_key=$(echo "${key}" | tr '[:lower:]' '[:upper:]')
```
mook-as, line 23: Too bad macOS `bash` is too old for `${key^^}`.

Given that we require WSL on Windows, I'm tempted to say you must install bash from hoembrew on mac. But that needs to be its own issue/PR though.

bats/tests/10-cli/5-paths.bats
```diff
@@ -0,0 +1,41 @@
...
+        assert_line --regexp "^export RDD_${upper_key}="
+    done
+}
+
+@test 'rdd svc paths <key> prints only the value' {
+    run -0 rdd svc paths log_dir
+    # Output should be a single line with no key prefix
+    assert_output --regexp '^(/|[A-Z]:)'
+    refute_output --regexp '^log_dir'
+    [[ "${#lines[@]}" -eq 1 ]]
```
mook-as:
```suggestion
    assert_equal "${#lines[@]}" 1
```
Mostly because doing it this way will show the actual count.

bats/Makefile
```diff
@@ -67,12 +67,23 @@ bats-containers: check-bats build-rdd build-mock-controller
...
-bats-controllers: bats-builtin bats-rdd bats-app bats-lima bats-containers
+# Run all controller tests, continuing past failures.
+bats-controllers:
+	@rc=0; \
+	$(MAKE) bats-builtin || rc=1; \
+	$(MAKE) bats-rdd || rc=1; \
+	$(MAKE) bats-app || rc=1; \
+	$(MAKE) bats-lima || rc=1; \
+	$(MAKE) bats-containers || rc=1; \
+	exit $$rc
```
mook-as: Please tell `make` about our actual dependencies; doing it this way means `make -j` will not do the right thing. (Fun fact: GNU make has `make -j -O` so the output isn't scrambled.)

pkg/service/cmd/service.go
```diff
@@ -213,20 +213,17 @@ func Start(ctx context.Context, args []string) error {
...
-	// TODO The log files should eventually move to the log directory
-	stdoutPath := filepath.Join(instance.Dir(), "rdd.stdout.log")
-	stderrPath := filepath.Join(instance.Dir(), "rdd.stderr.log")
-	if err := os.RemoveAll(stdoutPath); err != nil {
-		return err
-	}
-	if err := os.RemoveAll(stderrPath); err != nil {
-		return err
+	keepLogs := os.Getenv("RDD_KEEP_LOGS") != ""
+	title := os.Getenv("RDD_LOG_TITLE")
```
mook-as, line 217: These keep showing up; should we stuff them in a package somewhere?

mook-as (changes requested):
Main issue is that the log file symlink thing is totally incompatible with Windows; not making it work is fine, but doing stuff that definitely will not work doesn't seem good.

bats/tests/10-cli/5-paths.bats
jandubois: I copied this from the bats-core tests, but they don't have the assert_* functions available in that repo.

cmd/rdd/service_paths.go
jandubois: I typically optimize for usability, not for simplicity of implementation. So I prefer `--json` over `-o json`. But given that `kubectl` does the `--output` thing (because it has a gazillion of possible formats), I'm fine with doing this here too.

cmd/rdd/service_paths.go
jandubois: This is syntactic sugar for manual use. You can already do

```
rdd svc paths --json | jq -r .log_dir
```

So I'm not inclined to add `--key`. You can argue that this case is not common enough to be supported at all, and we can remove it, but I don't see the harm.

cmd/rdd/service_paths.go
jandubois: Not sure what funny paths would be a problem. We do quote the path, so anything reasonable should just work.

docs/design/environment.md
jandubois: It is actually `kubeconfig.yaml`. 😄 

The `~/.rd2/kube.config` is supposed to be the kubeconfig for the k3s cluster, in case you decide not to merge it with the one in `~/.kube/config`.

Maybe `RDD_KUBECONFIG` should be `RDD_CONFIG` to avoid confusion. And the filename just `config.yaml`. But that should happen in a separate PR. Let me know what you think before I create an issue!

pkg/service/cmd/service.go
jandubois: I don't know where to put them. And they are just 2 calls to `os.Getenv()`, so doesn't seem to be worth creating a package for.

pkg/util/logfile/logfile.go
jandubois: Symlinks should work fine if you have developer mode enabled, which seems like a reasonable requirement for a tool like Rancher Desktop.

If we don't have symlinks, then we need to rename the current log to the latest numbered version, when we create a new one.

I'm going to switch to that system because it is functionally almost identical (except the current log doesn't have a number).

We cannot change the log filename for hostagent; lima is accessing the file with the hard-coded name internally. That's why the active logs also cannot be moved to the log directory.

pkg/util/logfile/logfile.go
jandubois: We can delete it; but it doesn't really matter. It would just be another logfile (and most likely empty). I don't see how this is going to fail unless your filesystem is full or broken already.

bats/Makefile
jandubois: I don't know how to do that. The whole point of this change is to make sure `make` will continue after one of the dependencies failed (at least in CI I want all BATS failures, and not just the first group that fails). Running `make -j` will stop on first failure to build a dependency, afaik.

pkg/util/logfile/logfile.go
jandubois: I find the regex version clearer. The `\x00` stuff looks like it is some unsafe  code using CGO.

pkg/util/logfile/logfile_test.go
jandubois: If anything, we will prune more files (if the gap is during the retention period). I don't really care about adding more tests because gaps can't occur naturally; the user must have manually deleted files.

We can later refine this mechanism to work like logrotate or something; for now this is good enough.

pkg/util/logfile/logfile_test.go
jandubois: All the symlink code is gone now anyways.

jandubois: Don't really care. This can only happen if chaos monkey is installed on the machine...

jandubois (commented):
I've added a 👍 on all comments I believe I have addressed, and a 👎 on the ones I pushed back on.

Let me know if I missed any comments, or if you disagree with the ones I pushed back on!

(I have not yet pushed my feedback response commit; need to do some more testing)

docs/design/environment.md
mook-as: I don't know, it just says `Kubeconfig file`, so I assumed it's equivalent to `rdd service config` / `rdd ctl config view`.  In that sense, we should document that using `rdd service config` rather than reading the file directly is preferred, actually.  But yeah, `RDD_CONFIG` probably makes more sense as a path.

bats/Makefile
mook-as: If you want to run as much as possible instead of failing on the first error, you should use [`make --keep-going`](https://www.gnu.org/software/make/manual/html_node/Testing.html) instead.  And that should be a choice of the CI step that invokes `make`, not the Makefile.

jandubois: Thanks! I don't know how I missed `--keep-going` because that is exactly what I was looking for.

bats/helpers/paths.bash
```diff
@@ -6,44 +6,19 @@ inside_repo_clone() {
...
-    local folder_id=$1
-    local windows_path
-    windows_path=$(powershell.exe -Command "[System.Environment]::GetFolderPath(${folder_id})")
-    local unix_path
-    unix_path=$(wslpath -u "${windows_path}")
-    tr -d "\r" <<<"${unix_path}"
+# Convert 'export KEY="C:\win\path"' to 'export KEY="/mnt/c/wsl/path"'.
+win_to_wsl_exports() {
+    tr -d "\r" | while IFS= read -r line; do
+        if [[ "${line}" =~ ^(export\ [A-Z_]+=)\"(.+)\"$ ]]; then
```
mook-as, line 12: This eats all lines that _don't_ match the RE; is that expected?

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func newServicePathsCommand() *cobra.Command {
+	command := &cobra.Command{
+		Use:   "paths [key]",
+		Short: "Print instance paths",
+		Args:  cobra.MaximumNArgs(1),
+		RunE:  servicePathsAction,
+	}
+	command.Flags().StringP("output", "o", "table", "Output format: table, json, or shell")
```
mook-as: Consider using something like https://github.com/rancher-sandbox/rancher-desktop/blob/2f1623b54ee7ffd84d1708cafbbd15293f4c937d/src/go/rdctl/cmd/enum.go

… Actually, consider doing that in a _separate_ PR :)

cmd/rdd/service_paths.go
```diff
@@ -0,0 +1,90 @@
...
+	return command
+}
+
+func servicePathsAction(cmd *cobra.Command, args []string) error {
+	format, _ := cmd.Flags().GetString("output")
+	switch format {
+	case "table", "json", "shell":
```
mook-as: Seems like the format strings should be consts, to catch dumb typos (none of which I've spotted yet).

cmd/rdd/service_paths.go
mook-as: We quote it in a way suitable for round-tripping in golang, not shell. But yeah, let's just ignore this until we find issues.

docs/design/cmd_service.md
```diff
@@ -117,6 +117,42 @@ Show control plane logs.
 - `--stdout` (`-o`): Print stdout instead of stderr (default is stderr)
 - `--follow` (`-f`): Follow log output
 
+## `rdd service paths`
+
+Prints instance directory and file paths. Accepts an optional key argument to print a single path.
```
mook-as, line 122: None of the documentation (here, or `environment.md`) actually say that the key (`log_dir`) will be omitted in that case; it should be documented.

Might be good to show the outputs in the examples too.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -221,8 +222,20 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
...
+	// match what Lima expects for StopGracefully.
+	keepLogs := os.Getenv("RDD_KEEP_LOGS") != ""
+	title := os.Getenv("RDD_LOG_TITLE")
+	var header string
+	if title != "" {
+		// JSONL format: Lima's event watcher parses it as a zero-value Event
+		// and skips it; PropagateJSON logs it as a raw info line.
+		b, _ := json.Marshal(struct {
+			Title string `json:"title"`
+		}{title})
```
mook-as, line 235: Note to self: this is basically
```go
b, _ = json.Marshal(map[string]string{"title": title})
```

pkg/util/logfile/logfile.go
```diff
@@ -0,0 +1,103 @@
...
+	pattern := regexp.MustCompile(`^` + regexp.QuoteMeta(name) + `\.(\d+)\.log$`)
+
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, fmt.Errorf("read log directory %s: %w", dir, err)
+	}
+
+	// Find the highest existing sequence number.
+	maxN := 0
+	var numberedFiles []numberedFile
```
mook-as, line 45: Making this a `map[int]string` seems easier, but I guess this doesn't really hurt?

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,201 @@
...
+	dir := t.TempDir()
+
+	f, err := Create(dir, "test", false, "")
+	assert.NilError(t, err)
+	f.Close()
+
+	// Should create test.log directly (no numbered file on first call)
+	_, err = os.Stat(filepath.Join(dir, "test.log"))
+	assert.NilError(t, err, "expected test.log to exist")
+	_, err = os.Stat(filepath.Join(dir, "test.1.log"))
```
mook-as, line 28: Seems like you want to do a `Readdir` and check that only `test.log` exists, but this should be fine too.

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,201 @@
...
+	for i := 1; i <= count; i++ {
+		f, err := Create(dir, "prune", false, "")
+		assert.NilError(t, err, "Create #%d", i)
+		f.Close()
+	}
+
+	// File 1 should be pruned
+	name := filepath.Join(dir, "prune.1.log")
+	_, err := os.Stat(name)
+	assert.Assert(t, os.IsNotExist(err), "expected %s to be pruned", name)
```
mook-as: Would `assert.ErrorIs` be a better fit here?
In particular, the documentation for [`os.IsNotExist`](https://pkg.go.dev/os@go1.26.0#IsNotExist) says:
> This function predates [errors.Is](https://pkg.go.dev/errors#Is). It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,201 @@
...
+	// Active file should exist
+	_, err = os.Stat(filepath.Join(dir, "prune.log"))
+	assert.NilError(t, err, "expected prune.log to exist")
+}
+
+func TestKeepAll(t *testing.T) {
+	dir := t.TempDir()
+
+	// Create 7 files with keepAll=true
+	for i := 1; i <= 7; i++ {
```
mook-as: Naming the `7` as `retentionCount + 2` might be clearer.

mook-as (dismissed):
There are small improvements to be done, but if you feel like doing those as a follow-up that's fine too.

bats/helpers/paths.bash
jandubois: Yes, it is. It shouldn't really happen, but if it does, what should we do with those lines?

It is for BATS, not something end users run. So it won't get through CI if this is broken.

cmd/rdd/service_paths.go
jandubois: Please file an issue if you think we have enough usage for that pattern!

pkg/util/logfile/logfile.go
jandubois: Doesn't seem easier to me; you still need all the same loops unless I'm missing something.

jandubois:
I've marked all comments that I addressed with 👍 and the ones I think are fine as-is with 👎.

bats/helpers/paths.bash
mook-as: I'd expect them to be printed without modification, since that's the behaviour on non-Windows systems. Or, I suppose, fail immediately, but just hiding it doesn't seem right.

Basically, I'm wary of the drift between Windows and not-Windows.

pkg/util/logfile/logfile.go
mook-as: You would not need to declare the struct type, but yeah in use it's basically the same.

pkg/util/logfile/logfile_test.go
```diff
@@ -0,0 +1,204 @@
...
+		_, err := os.Stat(name)
+		assert.NilError(t, err, "expected %s to exist", name)
+	}
+
+	// No app.3.log (the third call created app.log, not a numbered file)
+	_, err = os.Stat(filepath.Join(dir, "app.3.log"))
+	assert.Assert(t, errors.Is(err, fs.ErrNotExist), "expected no app.3.log")
+}
+
+func TestPruning(t *testing.T) {
```
mook-as, line 57: I just realized we didn't test that `prune.99.log.not-really` does _not_ affect the pruning. Not sure if you'd feel that'd be too obscure.

(That's why I had the `\x00` in the `sscanf` implementation; it didn't have `$`.)

mook-as (approved):

PR #164, opened 2026-02-10
Warm CRD watch caches before deletion to prevent finalizer timeout

jandubois:
Warm the API server's watch cache for each custom resource type before deleting its CRD, preventing a race with the built-in CRD finalizer.

### Problem

On a fresh API server restart, watch caches are created on demand. When `cleanupUnusedCRDs` deletes a CRD, the CRD finalizer tries to list CR instances, which triggers lazy cache creation. `ResilientWatchCacheInitialization` (locked on since K8s v1.34) rejects that request with HTTP 429 instead of blocking. The finalizer enters exponential backoff and can exceed the 30-second cleanup timeout, causing startup failures.

### Fix

A single `List` request with retry on 429 before each CRD deletion forces cache initialization (typically <25ms). By the time the finalizer runs, the cache is warm and the 429 race cannot occur.

### DIFF
--- a/pkg/service/controllers/manager.go
+++ b/pkg/service/controllers/manager.go
@@ -18,13 +18,16 @@ import (
 
 	"github.com/go-logr/logr"
 
+	"k8s.io/apiextensions-apiserver/pkg/apihelpers"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


pkg/service/controllers/manager.go
```diff
@@ -623,6 +631,51 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
...
+	var version string
+	for _, v := range crd.Spec.Versions {
+		if v.Storage {
+			version = v.Name
+			break
+		}
+	}
+	if version == "" {
+		return
+	}
```
mook-as: This can be replaced with [`apihelpers.GetCRDStorageVersion`](https://pkg.go.dev/k8s.io/apiextensions-apiserver@v0.35.0/pkg/apihelpers#GetCRDStorageVersion).

pkg/service/controllers/manager.go
```diff
@@ -623,6 +631,51 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
...
+	err := wait.PollUntilContextTimeout(ctx, 10*time.Millisecond, 5*time.Second, true, func(ctx context.Context) (bool, error) {
+		_, listErr := dynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{Limit: 1})
+		if listErr == nil {
+			return true, nil
+		}
+		if apierrors.IsTooManyRequests(listErr) {
+			return false, nil
+		}
+		// Other errors (e.g., CRD not yet established) are not worth blocking on.
+		return true, nil
```
mook-as: Is that not just `return !apierrors.IsTooManyRequests(listErr), nil`? (With an updated comment, of course.)

mook-as (changes requested):
With the explanation, looks reasonable; just some local changes.

mook-as (approved):

PR #166, opened 2026-02-11
Distinguish starting Lima instance from broken state

jandubois:
Lima reports StatusBroken when the hostagent PID file exists but the socket is not yet responding. During startup this is normal: the hostagent creates its PID file before the socket. If the hostagent process is alive and the socket doesn't exist yet, treat the instance as starting rather than broken.

Also add a crash recovery test that kills the hostagent with SIGKILL and verifies the controller detects and restarts the VM. The existing broken-state test now simulates breakage by replacing the socket with a regular file instead of killing the hostagent, so it works on both VZ and QEMU backends.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -203,6 +203,9 @@ assert_limavm_not_exists() {
     # shellcheck disable=SC2030 # Persisted via save_var, not subshell
     OLD_HA_PID=$(get_hostagent_pid "${VM_NAME}")
     assert [ -n "${OLD_HA_PID}" ]
+    # Verify the PID is a running process. On Windows the hostagent runs as a
+    # Win32 process but BATS runs inside WSL, so kill can't reach it.
+    kill -0 "${OLD_HA_PID}"
     save_var OLD_HA_PID
 
     kill -9 "${OLD_HA_PID}"
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -98,6 +98,11 @@ get_instance_running_transition_time() {
         --output jsonpath='{.status.conditions[?(@.type=="InstanceRunning")].lastTransitionTime}'
 }
 
+get_hostagent_pid() {
+    local name=$1
+    cat "${LIMA_HOME}/${name}/ha.pid"
+}
+
 assert_recovery_completed() {
     local before_time=$1
     run -0 get_instance_running_transition_time "${VM_NAME}"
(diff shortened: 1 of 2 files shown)


bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -179,27 +184,69 @@ assert_limavm_not_exists() {
...
 
-    local pid
-    pid=$(cat "${pid_file}")
-    assert [ -n "${pid}" ]
+    # shellcheck disable=SC2030 # Persisted via save_var, not subshell
+    OLD_HA_PID=$(get_hostagent_pid "${VM_NAME}")
+    assert [ -n "${OLD_HA_PID}" ]
+    save_var OLD_HA_PID
+
+    kill -9 "${OLD_HA_PID}"
```
mook-as, line 211: Shouldn't we also assert that `OLD_HA_PID` refers to a valid PID?

I'm trying to figure out if I prefer `kill -KILL` for scripts… but that's kinda silly too :D

Also, side note: this does not work on Windows. At least add a comment about that? (Because, effectively, it's in the Win32 PID namespace.)

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -90,11 +90,24 @@ func (r *LimaVMReconciler) handleRunningState(ctx context.Context, limaVM *v1alp
...
+	// exists but the socket is not yet responding. During startup this is normal:
+	// the hostagent creates its PID file before the socket. If the hostagent
+	// process is alive and the socket doesn't exist yet, treat it as starting.
 	if inst.Status == limatype.StatusBroken {
-		logger.Info("Lima instance is in broken state, attempting force stop to recover", "errors", inst.Errors)
+		haSockPath := filepath.Join(inst.Dir, filenames.HostAgentSock)
+		if inst.HostAgentPID > 0 && !fileExists(haSockPath) {
+			logger.Info("Instance reports broken but hostagent is alive and socket not yet created, treating as starting", "pid", inst.HostAgentPID)
+			if !base.HasConditionWithReason(limaVM.Status.Conditions, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStarting) {
+				if err := r.updateCondition(ctx, limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonStarting, "Lima instance is starting"); err != nil {
```
mook-as, line 102: Not for this PR, but I wonder if we want something like

```
changed, err := r.setCondition(ctx, limaVM, ConditionInstanceRunning, metav1.ConditionFalse, ReasonRestarting, "…")
```

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -90,11 +90,24 @@ func (r *LimaVMReconciler) handleRunningState(ctx context.Context, limaVM *v1alp
...
+					logger.Error(err, "Failed to update starting condition")
+					return ctrl.Result{}, err
+				}
+			}
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
mook-as, line 108: The requeue is basically there to poll for the start to finish, right? (A comment might be useful, but it might just be because I got distracted and forgot the comment right above)

mook-as (dismissed):
Basically looks fine. Might want to check that `OLD_HA_PID` refers to a process.

mook-as (approved):

PR #167, opened 2026-02-12
`rdd service config`: add option to wait for changes

mook-as:
This will be used by the front end to monitor file changes (by waiting for the process to exit).

### DIFF
--- a/cmd/rdd/service.go
+++ b/cmd/rdd/service.go
@@ -59,6 +59,7 @@ func newServiceConfigCommand() *cobra.Command {
 		Args:  cobra.NoArgs,
 		RunE:  serviceConfigAction,
 	}
+	command.Flags().Bool("wait", false, "Wait until the kubeconfig changes before printing")
 	return command
 }
 
@@ -83,6 +84,13 @@ func serviceConfigAction(cmd *cobra.Command, _ []string) error {
 	if err := ensureServiceRunning(cmd.Context()); err != nil {
 		return err
 	}
+
+	if wait, err := cmd.Flags().GetBool("wait"); err == nil && wait {
+		if err := service.WatchKubeconfig(); err != nil {
+			return err
+		}
+	}
+
 	contents, err := service.GetKubeconfig()
 	if err != nil {
 		return err
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.25.0
 
 require (
 	github.com/MakeNowJust/heredoc v1.0.0
+	github.com/fsnotify/fsnotify v1.9.0
 	github.com/go-logr/logr v1.4.3
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/websocket v1.5.4-0.20250319132907-e064f32e3674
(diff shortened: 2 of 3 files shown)


cmd/rdd/service.go
```diff
@@ -59,6 +59,7 @@ func newServiceConfigCommand() *cobra.Command {
 		Args:  cobra.NoArgs,
 		RunE:  serviceConfigAction,
 	}
+	command.Flags().Bool("wait", false, "Wait until the kubeconfig changes before printing")
```
mook-as, line 62: Not sure if we want to call [`MarkHidden("wait")`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.MarkHidden); thoughts?

jandubois:
The `rdd svc config` code is calling `service.checkReadiness` (via `ensureServiceRunning`, which calls `service.Wait` etc).

I would consider it a bug if `checkReadiness` returns `true` while the config file does not yet exist, or hasn't been updated. So I think this should be fixed instead of adding a `--wait` option. Why would you ever call it without `--wait`?

mook-as:
`--wait` here is to wait for it to change; the goal is to catch when the kubeconfig has been regenerated (i.e. `rdd service delete + rdd service start`).  I guess this might be useful as a different (hidden) subcommand instead? I can do that, if that seems better.

The goal here is that I don't want the front end to know that the kubeconfig file actually exists on disk; it should just ask `rdd` to tell it when it needs to be reloaded.  I don't even need to output, since it can just do `rdd service config` again to fetch it.

jandubois:
> `--wait` here is to wait for it to change; the goal is to catch when the kubeconfig has been regenerated (i.e. `rdd service delete + rdd service start`). I guess this might be useful as a different (hidden) subcommand instead? I can do that, if that seems better.

This should still not be necessary. `rdd svc delete` will call `StopWithWait(true)` if the service exists. Then you can either use `rdd svc start --wait` to make sure `start` doesn't return until the service is ready, or you call `rdd svc config`, which does an implicit wait.

All the CLI commands have been designed that you can easily script them without having to retry etc. Internally the API calls will throw an error when you e.g. try to start a service that hasn't been created.

But you can just call `rdd svc config` and it will internally call `service.Create`, `service.Start` and `service.Wait` as necessary.

So I still think everything you need already exists.

jandubois:
> it should just ask `rdd` to tell it when it needs to be reloaded

I think this should be just part of regular error handling: when you make an API call and get an auth error, you try to refresh the config, which will restart the service if it has been stopped. If that fails too, then you have a permanent error that you need to bring to the user's attention.

mook-as:
> > it should just ask `rdd` to tell it when it needs to be reloaded
> 
> I think this should be just part of regular error handling: when you make an API call and get an auth error, you try to refresh the config, which will restart the service if it has been stopped.

That would be very confusing for the backend, because we have multiple connections (essentially, one per type for watches, plus whatever else we happen to do).  It's not a single service to talk to the backend.  This means that when the connection dies, we'd get half a dozen `rdd service config` calls at the same time, causing multiple _simultaneous_ startups, which isn't currently supported (and is unlikely to ever be).

jandubois:
Ok, let's discuss this in person; I don't understand how your proposal would help with all existing connections needing to be reestablished.

Note also that in general the daemon should not be shut down while the app is running. So maybe the correct thing to do is shut down the app when the daemon goes away.

jandubois:
> This means that when the connection dies, we'd get half a dozen `rdd service config` calls at the same time, causing multiple _simultaneous_ startups

This is a race condition that needs to be resolved anyways: there has to be a check in the service that prevents it from starting if there is already an rdd.pid for the current instance. So only the first `rdd svc config` would restart the service, all the others would just wait for the service to become ready.

mook-as:
Discussed this online; the config (from `rdd service config`) will never change, at least for the lifetime of the UI process.  I will make a PR to document this expclitly, but that means this PR is unnecessary.

PR #168, opened 2026-02-12
docs/cmd_service: update specs

mook-as:
From discussion with Jan, the kubeconfig will not change until reset (or, at some point in the future, when we need to update the `notAfter` date).

Replaces #167

### DIFF
--- a/docs/design/cmd_service.md
+++ b/docs/design/cmd_service.md
@@ -113,7 +113,9 @@ Prints a kube config with context and service account[^sa] setup to give access 
 
 [^sa]: A locked deployment profile will result in a service account with limited functionality to prevent bypassing the profile restrictions.
 
-Runs `rdd service start` to ensure the `apiserver` is ready to accept requests.
+This command will implicitly start the service, and will block until startup is complete.
+
+The returned config will be static for the lifetime of the service, but may change the next time the service starts.
 
 ## `rdd ctl`
 


docs/design/cmd_service.md
```diff
@@ -113,7 +113,9 @@ Prints a kube config with context and service account[^sa] setup to give access
 
 [^sa]: A locked deployment profile will result in a service account with limited functionality to prevent bypassing the profile restrictions.
 
-Runs `rdd service start` to ensure the `apiserver` is ready to accept requests.
+This command will implicitly start the service, and will block until startup is complete.
+
+The returned config will be static for the lifetime of the service, and will be the same across service restarts.
```
jandubois: As you noted in the PR summary, the config may change on service restarts. This can happen when the certs need to be refreshed, but could also happen when networking changes (maybe a hostname needs to be added to the cert).

There are potentially changes to the capabilities if a deployment profile changes the RBAC settings of the service account, but I don't think this will change the config.

I would say something like: "The returned config will remain valid while the service is running. But it should not be cached by the client; the client should request a fresh config every time the client restarts (which will implicitly restart the service if it is still stopped)."

mook-as: _should not be cached by the client_ is too broad; I intend to cache it in memory for the lifetime of the client, which would run afoul of that.  How about _should not be persisted…_ instead?

But again, you are explicitly saying that the config will change at some unpredictable point (_when the networking changes_, etc.); this is incompatible with saying that the frontend should assume that the kubeconfig does not change within the lifetime of the (frontend) process.

I'm basically confused as to how you expect me to have a working connection, if you're saying both that I should not detect the change, and also that it will change and I need to detect it.

I'm basically making this PR to document my expectations, so that if we need to change it we can document when that happens so we can update the frontend; your feedback is covering future cases, which will mean I need to implement the frontend to match _that_, which doesn't exist yet.

jandubois: > _should not be cached by the client_ is too broad; I intend to cache it in memory for the lifetime of the client, which would run afoul of that. How about _should not be persisted…_ instead?

Yes, that is fine. I thought that was implied by "the client should request a fresh config every time the client restarts".

> But again, you are explicitly saying that the config will change at some unpredictable point (_when the networking changes_, etc.); this is incompatible with saying that the frontend should assume that the kubeconfig does not change within the lifetime of the (frontend) process.

No, I'm only talking about caching the value across front-end restarts. The config should not change while the frontend remains running (assuming the daemon is not crashing or shut down from the CLI). If the frontend shuts down the daemon, then the frontend is expected to exit as well.

> I'm basically confused as to how you expect me to have a working connection, if you're saying both that I should not detect the change, and also that it will change and I need to detect it.

It will not change while the frontend runs. It may change when the frontend restarts.
 The daemon restarting without the frontend restarting is not a supported scenario.

> I'm basically making this PR to document my expectations, so that if we need to change it we can document when that happens so we can update the frontend; your feedback is covering future cases, which will mean I need to implement the frontend to match _that_, which doesn't exist yet.

It is good to tie down the spec so we are all on the same page. I hope my comments cleared the confusion. If you still think you need to detect a potential config change, then I'm not communicating clearly.

---

Maybe we can at some point have some logic to prevent the control plane from shutting down while there are open connections. Not sure if this is possible. We also must be careful not to block the user logging out, or shutting down.

jandubois (approved):
Thanks, LGTM

jandubois:
I just realized one other potential reason for our miscommunication:

I think of the lifetime of the service as the time until the service gets deleted (until factory-reset), and I think of the lifetime of the daemon as the time until the process is stopped.

Maybe we need to define these terms in our docs, so we all use them the same way.

mook-as:
… well, I'm literally looking at the documentation for `rdd service start`, so I think of the service as the thing I'm starting. That is, the process, not the state.

jandubois:
> … well, I'm literally looking at the documentation for `rdd service start`, so I think of the service as the thing I'm starting. That is, the process, not the state.

That's why we need to define how we want to use these terms.

So the lifetime of a LimaVM instance will also be only until it is stopped, and not until it is deleted? How can you start something again that no longer exists? And how do you delete something that doesn't exist? This usage does not really work for me.

But what would be the names for services and instances as persistent state and not just their ephemeral process?

PR #172, opened 2026-02-18
Update the App spec

jandubois:
Update the spec to match the conventions we use for LimaVM resources, e.g. `spec.running: true` instead of `spec.status: running`, and `status.Conditions` instead of `status.status`.

### DIFF
--- a/docs/design/api_app.md
+++ b/docs/design/api_app.md
@@ -54,33 +54,52 @@ kind: App
 metadata:
   name: rancher-desktop
   namespace: rancher-desktop
+
 spec:
   containerEngine:
     name: moby
   kubernetes:
     version: 1.30.0
-  status: running
-  dryRun: false
+  running: true
+
 status:
-  appVersion: 2.0.0
-  onlineStatus: true
-  status: starting
+  containerEngine:
+    name: moby
+  kubernetes:
+    version: 1.30.0
+
   progress: "Downloading Kubernetes 1.30.0"
+  version: 2.0.0
+  onlineStatus: true
+
+  conditions:
+  - type: AppCreated
+    status: "True"
+    reason: Created
+    message: Rancher Desktop App created successfully
+  - type: AppRunning
+    status: "True"
+    reason: Started
+    message: Rancher Desktop App started successfully
 ```
 
-### `spec`
+- **spec**: Has all the usual fields that are in `settings.json` in "Rancher Desktop 1.x".
 
-The `spec` has all the usual fields that are in `settings.json` in "Rancher Desktop 1.x".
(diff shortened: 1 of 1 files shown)


docs/design/api_app.md
```diff
@@ -54,38 +54,57 @@ kind: App
...
   namespace: rancher-desktop
+
 spec:
   containerEngine:
     name: moby
   kubernetes:
     version: 1.30.0
-  status: running
-  dryRun: false
+  running: true
```
mook-as, line 63: I think that was so we can set it to the desired state, but off the top of my head I can't think of any desired states other than "created, but stopped instead of running" and "running", so this seems fine.

docs/design/api_app.md
```diff
@@ -54,38 +54,57 @@ kind: App
...
-#### `status`
+- **spec.running**: Set to `true` when the LimaVM should be running, set to `false` when it should be stopped.
 
-The `status` field is set to the desired state of the VM: `running` or `stopped`. The controller will update the `spec.status` field of the `LimaVM` object to match.
+- **status.conditions**: Standard Kubernetes conditions tracking the App state.
 
-#### `dryRun`
+  | Type         | Status      | Reason         | Description                                                  |
+  |--------------|-------------|----------------|--------------------------------------------------------------|
+  | `AppCreated` | `Unknown`   | `Pending`      | Reconciler has seen the resource; creation not yet attempted |
```
mook-as: Do all the conditions need the `App` prefix? This is already on the `App` resource, so it's rather redundant.

docs/design/api_app.md
```diff
@@ -54,38 +54,57 @@ kind: App
...
   name: rancher-desktop
   namespace: rancher-desktop
+
 spec:
   containerEngine:
     name: moby
   kubernetes:
     version: 1.30.0
-  status: running
-  dryRun: false
```
mook-as, line 63: We will need something equivalent to the `dryRun` functionality for the UI (so it can validate settings before committing), but that can be done via server-side dry run I guess?

mook-as (dismissed):

docs/design/api_app.md
jandubois: I added this to the spec before I figured out that Kubernetes has a native mechanism for that:

```
kubectl apply ... --dry-run=server
```

When using the API you have to append a `DryRun=All` query parameter to the `PUT` or `PATCH` URL.

docs/design/api_app.md
jandubois: I agree. I think we should also remove the `Instance` prefix from the LimaVM conditions.

mook-as (approved):

PR #174, opened 2026-02-18
Add container namespaces

mook-as:
This adds a `Namespace` kind to manage container engine namespaces.  The name is a bit confusing (because Kubernetes has its own namespaces); feel free to suggest alternatives if you prefer.


jandubois:
> feel free to suggest alternatives if you prefer.

I definitely prefer `ContainerNamespace` to avoid confusion with the builtin `namespace` resource type.

I'm also not sure why this PR needs support for applyconfigurations. I'm not saying we don't want to use them, but they seem mostly unrelated to having a namespace resource type.

Haven't looked at the actual code yet.

pkg/apis/containers/v1alpha1/applyconfiguration/containers/v1alpha1/namespace.go
jandubois, line 1: Why does this file exist? Is it a leftover that should have been deleted as part of the renaming?

pkg/controllers/containers/containernamespace/container_namespace_webhooks.go
```diff
@@ -0,0 +1,37 @@
...
+	"errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlwebhookadmission "sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/containers/v1alpha1"
+)
+
+type ContainerNamespaceDeleteValidator struct {
+	Client client.Client
```
jandubois: This seems unused. `setupWebhookWithRuntimeConfig` uses

```go
Validator: &ContainerNamespaceDeleteValidator{},
```

pkg/controllers/containers/containernamespace/container_namespace_webhooks.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// ValidateCreate implements [ctrlwebhookadmission.Validator].
+func (c *ContainerNamespaceDeleteValidator) ValidateCreate(context.Context, *v1alpha1.ContainerNamespace) (ctrlwebhookadmission.Warnings, error) {
+	return nil, errors.New("webhook does not implement create")
+}
+
+// ValidateDelete implements [ctrlwebhookadmission.Validator].
+func (c *ContainerNamespaceDeleteValidator) ValidateDelete(context.Context, *v1alpha1.ContainerNamespace) (ctrlwebhookadmission.Warnings, error) {
+	// TODO: We should validate that:
```
jandubois, line 25: Please create a GitHub issue for this!

jandubois (dismissed):
Thanks, LGTM

I think there is one unused file and one unused struct field, but otherwise looks fine.

pkg/controllers/containers/containernamespace/container_namespace_webhooks.go
mook-as: Thanks, created #184

jandubois (approved):
Thanks, LGTM

PR #175, opened 2026-02-19
Generate apply configurations

mook-as:
This generates apply configurations for use with server-side apply.  There are no uses of it yet; that will be in follow-up PRs.


jandubois (approved):
Thanks, LGTM

PR #176, opened 2026-02-19
Detect template ConfigMap changes and restart running instances

### LINKED ISSUE #149: Watch the `templateRef` ConfigMap and update the Lima instance to match template changes
This will typically require stopping and restarting the instance.

jandubois:
#### Summary

- Add `observedTemplateResourceVersion` to LimaVM `status` to track the last template ConfigMap version written to disk
- When the template ConfigMap changes, compare its data with the on-disk `lima.yaml`; if they differ, write the new template and stop a running instance so the existing state machine restarts it
- A stopped instance gets its `lima.yaml` updated without starting

#### Design

The `Owns(&corev1.ConfigMap{})` watch already triggers reconciliation when the managed template ConfigMap changes. This PR adds the logic to act on it:

1. The reconciler compares `templateConfigMap.ResourceVersion` against `status.observedTemplateResourceVersion`
2. On mismatch, it reads the on-disk `lima.yaml` and compares with the ConfigMap data
3. If the template data differs, it writes the new file and calls `stopInstance()` for a running VM
4. The next reconcile sees `spec.running=true` + stopped instance and starts it normally
5. If only the `resourceVersion` changed (e.g. a label update), it records the new version without restarting

Closes #149

### DIFF
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -10,6 +10,7 @@ mkfifo
 mktemp
 moby
 netcat
+printenv
 pulumi
 Trivy
 
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -391,6 +391,108 @@ assert_restart_annotation_absent() {
     refute_output "true"
 }
 
+# Template change detection tests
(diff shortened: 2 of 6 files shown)


jandubois:
Superseded by #206, which reimplements the changes based on the event-watcher instead of polling.

PR #178, opened 2026-02-20
Add limavm edit command

### LINKED ISSUE #73: Create `rdd limavm edit` command
It would work similar to `rdd ctl edit cm`, except it would extract the `template` key from the ConfigMap data, and invoke the editor just for that. So it unwraps the YAML from being inside a string.

If the user deletes all content of the template, it means they want to abort the edit command without making any changes.

If the modified template does not pass ConfigMap validation, the command should offer the user a chance to fix the modified template instead of simply discarding the changes. This can be done later as a separate task.

Nino-K:
The rdd limavm edit command extracts the template key from the ConfigMap data and opens it in the editor. It performs preliminary validation on the user input to ensure the file is neither empty nor unchanged.

This change fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/73

### DIFF
--- a/cmd/rdd/limavm.go
+++ b/cmd/rdd/limavm.go
@@ -6,9 +6,11 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -21,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
 	corev1scheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/kubectl/pkg/cmd/util/editor"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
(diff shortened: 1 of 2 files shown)


cmd/rdd/limavm.go
```diff
@@ -44,10 +47,75 @@ func newLimaVMCommand() *cobra.Command {
...
+	}
+
+	if updatedContent == templateData {
+		logrus.Info("No changes made to template, skipping update")
+		return nil
+	}
+
+	configMap.Data[limav1alpha1.TemplateConfigMapKey] = updatedContent
+	if err := c.Update(ctx, configMap); err != nil {
+		return fmt.Errorf("failed to update template ConfigMap: %w", err)
```
jandubois, line 113: Issue #73 says:

> If the modified template does not pass ConfigMap validation, the command should offer the user a chance to fix the modified template instead of simply discarding the changes. This can be done later as a separate task.

I guess you have decided that that should happen in a separate PR?

cmd/rdd/limavm.go
```diff
@@ -347,3 +415,12 @@ func newLimaVMLogsCommand() *cobra.Command {
 	command.Flags().BoolP("follow", "f", false, "Follow log output")
 	return command
 }
+
+// editorEnvs returns an ordered list of env vars to check for editor preferences.
+func editorEnvs() []string {
+	return []string{
+		"KUBE_EDITOR",
```
jandubois, line 422: We could also add an `RDD_EDITOR` option at the top, but I don't think that is really needed.

jandubois (dismissed):
Thanks, LGTM

bats/tests/33-lima-controllers/limavm-lifecycle.bats
```diff
@@ -212,3 +212,83 @@ EOF
...
+}
+
+@test "limavm edit fails gracefully for nonexistent VM" {
+    run -1 rdd limavm edit "nonexistent-vm"
+    assert_output --partial 'LimaVM \"nonexistent-vm\" not found in any namespace'
+}
+
+@test "limavm edit rejects invalid template" {
+    skip
+    # TODO: Implement test for invalid template edit once validation is wired in.
```
jandubois: Validation is actually working already; we tested this together yesterday.

Nino-K: I think the confusion may be due to the wording in the TODO comment. I was referring specifically to the case where the limavm template validation fails, this test covers that scenario. It is also directly related to your first comment above regarding issue #73.

jandubois: Ok, this may be difficult to test in BATS. You can only ask the user if they want to try again if stdin/stdout is a terminal. If it isn't, then you display the error and exit, just like now.

Nino-K: Yeah, I thought I would keep the placeholder to think about it when we tackle #182

jandubois: It is better to put reminders into a GitHub issue.
We don't really want to have dead/non-functional code in the tree.

jandubois (approved):

PR #179, opened 2026-02-20
mock controller: use apply config

mook-as:
#175 generated apply configurations.  This uses them in the existing mock reconcilers.

The image reconciler is a bit annoying: we can't use apply configurations with `GenerateName` (because it can't figure out whether it needs to create a new object, or update one of the existing objects), so we need to keep some of the old logic to be able to create the image if it doesn't exist, then update it.  Since the relevant field selectors select upon the status (which is a sub-resource), we can't do that atomically anyway.

### DIFF
--- a/pkg/controllers/mock/container_reconciler.go
+++ b/pkg/controllers/mock/container_reconciler.go
@@ -16,9 +16,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	metav1apply "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


jandubois:
AI Code Review

---

## Bug: volume reconciler skips the status subresource

The container and image reconcilers apply the status subresource
separately via `SubResource("status").Apply`. The volume reconciler
instead passes `WithStatus(statusApplyConfig)` in the main apply call
(`volume_reconciler.go:90`). If the Volume CRD defines a status
subresource, the API server ignores status fields in the main apply, so
status data would be silently lost.

Also, `errs = append(errs, err)` on line 93 appends unconditionally,
including nil on success. `errors.Join` filters nils, so this causes no
runtime failure, but it is inconsistent with the other two reconcilers,
which guard with `if err != nil`.

## Bug: shared `statusApplyConfig` mutated across loop iterations

`image_reconciler.go:117` calls `statusApplyConfig.WithRepoTag(tag)`,
which mutates the receiver in place. The `statusApplyConfig` is built
once before the tag loop (line 91), so each iteration overwrites the
previous tag on the same object. This works today because the loop runs
sequentially and each apply sends the current state, but:

1. It is a latent bug if the code is reordered or parallelized.
2. The same `statusApplyConfig` serves the `else` branch (dangling
   images, line 125), which would carry a stale tag from a previous
   inspect if both paths ever ran. The branches are exclusive per
   inspect today, but the coupling is fragile.

Building a fresh `statusApplyConfig` per tag, or at least documenting
the mutation, would make the intent clear.

## Inconsistency: volume reconciler builds ownerReference inline

The container and image reconcilers build `ownerReference` once before
the loop. The volume reconciler constructs it inline inside the loop
body (`volume_reconciler.go:83-89`). Not a bug, but inconsistent with
the other two reconcilers in this PR.

## Minor: unrelated change in `sanitizeKubernetesObjectName`

The image commit adds `"_", "-"` to the replacer
(`image_reconciler.go:49`). This changes name sanitization for all
callers. If intentional, it deserves a mention in the commit message; it
is easy to miss buried in the apply-config conversion.

## Nit: `imageName` is vestigial

`image_reconciler.go:86-89` computes `imageName` but uses it only in
the time-parse error log on line 101. The old code used it throughout
`upsertImage` error messages. Not wrong, just a near-dead variable.

mook-as:
I believe I have addressed all the AIness at this point.

jandubois (approved):
Thanks, LGTM

PR #186, opened 2026-02-23
apis: container: move namespace to .status

mook-as:
Because `.metadata.labels` is a `map[string]...`, we can't use it as part of a field selector.  Move the `namespace` (and where appropriate, `name`) to be part of status instead.  This affects containers, images, and volumes.


jandubois (approved):
Thanks, LGTM

PR #190, opened 2026-02-24
Update existing CRDs on startup instead of skipping them

jandubois:
`installControllerCRDs` skipped CRDs that already existed, so new schema fields (e.g. `status.observedTemplateResourceVersion`) were never added after an upgrade. Controllers writing to these fields got silent "unknown field" warnings and lost the data.

Replace the skip with an update: fetch the existing CRD, overwrite its spec with the embedded version, and persist it. This ensures the schema stays current across binary upgrades.

Also includes a fix for readiness check hanging when `--controllers=""` is a session override.

### DIFF
--- a/pkg/service/controllers/manager.go
+++ b/pkg/service/controllers/manager.go
@@ -20,6 +20,7 @@ import (
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


pkg/service/controllers/manager.go
```diff
@@ -385,10 +386,18 @@ func (scm *SharedControllerManager) installControllerCRDs(ctx context.Context) e
...
 			}
 
-			// Check if CRD already exists
-			_, err = crdClient.Get(ctx, crd.Name, metav1.GetOptions{})
+			// Update existing CRDs so new schema fields take effect on upgrade.
+			existing, err := crdClient.Get(ctx, crd.Name, metav1.GetOptions{})
 			if err == nil {
-				klog.Infof("%s CRD already exists", controllerName)
+				if !apiequality.Semantic.DeepEqual(existing.Spec, crd.Spec) {
+					existing.Spec = crd.Spec
```
Nino-K, line 393: Is it intentional to overwrite the full CRD here and not just the validation and schema bits?

Nino-K (approved):

PR #192, opened 2026-02-24
Makefile: use fork of ltag for better excludes

mook-as:
This changes ltag to a fork, to support:
- Reading ignores from `.gitignore`
- Exclude files by path, to avoid being too generic

Given that https://github.com/containerd/ltag/pull/16 already exists, I'm not sure trying to upstream this would be useful.


jandubois (approved):
Thanks, LGTM

PR #194, opened 2026-02-24
Replace try-polling with rdd ctl wait in BATS tests

jandubois:
Replace manual `try`-loops with native `kubectl wait` semantics wherever possible: `--for=condition` for status conditions, `--for=jsonpath` for field values, `--for=create` for resource existence, and `--for=delete` for resource removal. Remove helper functions that existed only to support the polling pattern.

Remaining `try`-loops poll non-Kubernetes state (filesystem, processes, SSH readiness, log content) or need output validation from `rdd ctl get`.

### DIFF
--- a/bats/tests/31-rdd-controllers/notary-external.bats
+++ b/bats/tests/31-rdd-controllers/notary-external.bats
@@ -37,7 +37,8 @@ assert_process_exited() {
     echo "$!" >"${BATS_FILE_TMPDIR}/controller_pid"
 
     # Wait for external controller to register in discovery system
-    try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system --allow-missing-template-keys=false --output jsonpath='{.data.rdd}'
+    rdd ctl wait --for=jsonpath='{.data.rdd}' \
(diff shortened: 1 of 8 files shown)


bats/tests/31-rdd-controllers/notary-external.bats
```diff
@@ -37,7 +37,8 @@ assert_process_exited() {
     echo "$!" >"${BATS_FILE_TMPDIR}/controller_pid"
 
     # Wait for external controller to register in discovery system
-    try --max 20 --delay 1 -- rdd ctl get configmap rdd-controller-manager --namespace rdd-system --allow-missing-template-keys=false --output jsonpath='{.data.rdd}'
+    rdd ctl wait --for=jsonpath='{.data.rdd}' \
```
Nino-K, line 40: The old command failed when the field was absent, the new version does not. I assume the `wait` implementation is doing the magic here?

jandubois: Yes, it waits until the jsonpath expression returns a non-empty value

Nino-K (approved):

PR #195, opened 2026-02-25
Decouple restart request from execution via status.restartNeeded

jandubois:
Add a restart mechanism that separates "request restart" from "execute restart." A new `status.restartNeeded` field tracks pending restarts. The reconciler sets this field (from the `restartRequested` annotation) and `handleRunningState` acts on it — stopping only when the VM is fully running, never during startup.

The reconciler translates the annotation to status in two cycles (status first, then annotation removal) for crash safety. `handleRestartNeeded` stops running instances, waits for starting ones, and clears the flag for stopped ones.

Also adds:
- `status.restartCount` to track `Running` state transitions
- 30-second timeout on `StopGracefully` before falling back to force stop
- `rdd limavm restart` command (sets annotation + `spec.running=true`)
- BATS tests for restart of running/stopped VMs and CLI annotation

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-cli.bats
+++ b/bats/tests/33-lima-controllers/limavm-cli.bats
@@ -163,6 +163,31 @@ assert_created() {
     assert_fatal "${not_found}"
 }
 
+@test "lima restart sets annotation and running=true" {
+    # Create a template ConfigMap and VM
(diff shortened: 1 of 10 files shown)


jandubois:
This PR, which is still based on the store.Inspect() polling mechanism, has been folded into #199 because it needs to be integrated with the hostagent events watcher.

PR #197, opened 2026-02-25
CI: BATS: Run in parallel

mook-as:
Since we use an instance per BATS suite, we can run them in parallel to make CI (hopefully) a bit faster.  We already have `--keep-going` to avoid stopping on the first failure; with `--jobs --output-sync=target` we can do that in parallel but still print logs from each suite as a group.

I'm not sure this actually _is_ faster, but the previous commit (56e68e9dc33b02fb1161a7b542bbeeb2e4c18305) seemed to take extra long (~30m), so this is at least faster than that? A different comparison build (4dee57c1fe06444edd85d807c652aca53e2d8e4e) took ~13m macOS, ~12m Linux, ~16m Windows (BATS step only). So this is still much faster, slightly faster, and slightly faster.

### DIFF
--- a/.github/actions/spelling/expect/tools.txt
+++ b/.github/actions/spelling/expect/tools.txt
@@ -1,8 +1,11 @@
 # Words from tools.
 buildkit
 buildvcs
+coreutils
 cpanm
 cpanminus
+gnubin
+libexec
 ltag
 maxdepth
 mindepth
@@ -10,6 +13,7 @@ mkfifo
 mktemp
 moby
 netcat
+nproc
 pulumi
 Trivy
 
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -75,12 +75,15 @@ jobs:
       if: runner.os == 'macOS'
       working-directory: ${{ runner.temp }}
       run: |
-        brew install bash
+        brew install bash coreutils make
 
         # Set up run-in-wsl helper script (as a symlink to bash)
         mkdir -p bin
         ln -s $(brew --prefix bash)/bin/bash bin/run-in-wsl
         readlink -f bin >> "$GITHUB_PATH"
+
+        # Use GNU make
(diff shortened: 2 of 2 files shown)


jandubois:
> A different comparison build ([4dee57c](https://github.com/rancher-sandbox/rancher-desktop-daemon/commit/4dee57c1fe06444edd85d807c652aca53e2d8e4e)) took ~13m macOS, ~12m Linux, ~16m Windows (BATS step only). So this is still much faster, slightly faster, and slightly faster.

CleanShot 2026-02-25 at 18 09 15@2x

I see the numbers for this PR as 8m/10m/18m, so agree with the first 2, but Windows is actually slightly slower.

I see a slight advantage in that this might show BATS failures on Unix faster, but does not allow us to merge any faster because we still have to wait for the slowest step.

jandubois (approved):
Thanks, LGTM

I have slight concerns that `--output-sync=target` will swallow output when the GitHub runner disconnects prematurely and only partial output is available. But that happens only rarely and is a sign of a bigger issue anyways.

I do wonder what `nproc` returns on a Windows runner? If it just returns `1`, then that would explain why the Windows timing doesn't seem to change beyond statistical fluctuation.

We could split BATS into separate jobs that would run in parallel on different runners (duplicating the setup requirements), but I don't think we should spend time on that yet.

jandubois:
> I have slight concerns that `--output-sync=target` will swallow output when the GitHub runner disconnects prematurely and only partial output is available. But that happens only rarely and is a sign of a bigger issue anyways.

Looks like I ran into this earlier than expected: https://github.com/rancher-sandbox/rancher-desktop-daemon/actions/runs/22425591114/job/64932878710?pr=199

I think one test is simply hanging because `rdd svc stop` never finishes. But because it hangs, we don't get to see where it hangs; the whole output for `bats-lima` is missing.

I think the right fix would be to run each `make bats-xxx` step with a timeout, maybe 30min or something generous, but don't leave it hanging for 6 hours until GitHub kills the action.

PR #198, opened 2026-02-25
mock controller: Add a mock logs endpoint

mook-as:
For working on the UI, we need a mock logs endpoint.  Add it so we can use it for the front end later.

This needs to be looked at after #188 is merged.

### DIFF
--- a/bats/helpers/commands.bash
+++ b/bats/helpers/commands.bash
@@ -23,6 +23,22 @@ ctrctl() {
 curl() {
     command "curl${EXE}" "$@"
 }
+
+# Check if curl supports WebSockets; some tests may require it.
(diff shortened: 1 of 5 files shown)


bats/helpers/commands.bash
```diff
@@ -23,6 +23,22 @@ ctrctl() {
 curl() {
     command "curl${EXE}" "$@"
 }
+
+# Check if curl supports WebSockets; some tests may require it.
+curl_has_websocket_support() {
```
jandubois, line 28: This could be `assert_curl_has_websocket_support` and include the `skip` logic because that is the only way it is being called.

jandubois (approved):
Thanks, LGTM

PR #199, opened 2026-02-25
Replace hostagent polling with event-driven watcher

### LINKED ISSUE #149: Watch the `templateRef` ConfigMap and update the Lima instance to match template changes

### LINKED ISSUE #150: Watch the hostagent log file
When an instance is started we need to run a goroutine watching the hostagent log file. That way we can properly detect when the instance is started and functional. It also allows us to detect when the instance has shut down.

This can be extended to keep a timer when signalling the instance to shut down, so the goroutine can initiate a forceful shutdown if the graceful one doesn't complete.

### LINKED ISSUE #196: `rdd lima start/stop/delete` commands all should have a `--wait=false` option
Just like `rdd svc start/stop/delete`.

By default all CLI commands wait until the desired state has been reached, to make scripting simple.

Maybe we should have a `--timeout=10s` option. In `kubectl` `--timeout=0s` means to wait indefinitely, so we can't use that as an alternative for `--wait=false`; we need both, just like e.g. `kubectl delete ... --wait=false`.

jandubois:
Replace the 5-second `store.Inspect()` polling loop with a push-based watcher that tails Lima hostagent event streams. The watcher goroutine tracks lifecycle phases (`Starting`, `Running`, `Stopped`) and triggers reconciles via `source.Channel` as transitions occur.

Add restart support: a `restartRequested` annotation triggers a two-phase annotation-to-status translation (set `status.restartNeeded`, then remove annotation). The reconciler stops the instance when `restartNeeded` is set on a running VM and lets normal start logic bring it back. A monotonic `restartCount` replaces `lastTransitionTime` for detecting recovery completion: its next value is predictable, so callers can use `kubectl wait --for=jsonpath` to block until a restart finishes.

Add `--wait` (default `true`) and `--timeout` flags to the `start`, `stop`, `restart`, and `delete` CLI commands. When `--wait` is true, the command blocks until the desired state is reached. A `--timeout` of `0` (the default) waits indefinitely; a positive duration caps the wait.

`hostagent_watcher.go` (new): Watcher goroutine races `haCmd.Wait()` against `events.Watch()`. Process exit triggers `phase=Stopped` even without an `Exiting` event (e.g., SIGKILL). A `procExited` channel signals when the process is dead, distinct from the watcher goroutine finishing.

`limavm_controller.go`: Add `instanceStates` map and `reconcileChan`. Register a shutdown runnable that sends SIGINT to all hostagents, waits up to 10 seconds, then falls back to SIGKILL. Use `Status().Patch()` instead of `Status().Update()` for condition writes to avoid resource version conflicts during slow operations. Check `restartRequested` annotation before dispatching to `handleRunningState`.

`limavm_lifecycle.go`: Split `handleRunningState` into `handleWatchedState` (watcher active, phase-based dispatch) and `handleUnwatchedState` (no watcher, `store.Inspect` fallback for orphaned hostagents after controller restart). Remove the broken-state heuristic and polling `RequeueAfter`. Use `exec.Command` instead of `exec.CommandContext` since the hostagent lifecycle is managed explicitly through `shutdownAllHostagents` and `stopInstance`. Record the `Starting` condition before `waitForPIDFile` so the status transition is visible even if the object changes during startup. Force-stop orphaned VM drivers (e.g., QEMU) when the hostagent exits but the driver survives in its own process group. Add `handleRestartAnnotation` (two-phase annotation-to-status translation) and `handleRestartNeeded` (stop-then-start adapted to watcher phases). Increment `restartCount` atomically with the `Running=True/Started` condition in `handleWatchedState`.

`cmd/rdd/limavm.go`: Add `restart` command that sets the `restartRequested` annotation and `spec.running=true` in a single patch. Add `--wait` and `--timeout` flags to `start`, `stop`, `restart`, and `delete`. Add `waitForRunningCondition` and `waitForRestartCount` poll helpers.

`service.go`: Wait for the controller manager goroutine to finish before `Run` returns. Without this, the service process exits before `shutdownAllHostagents` completes, and hostagents survive shutdown.

`api_lima.md`: Add `restartNeeded` and `restartCount` to status documentation. Add lifecycle sequence diagrams for normal start/stop, crash recovery, orphan recovery, restart, and graceful shutdown.

`cmd_lima.md`: Document `--wait` and `--timeout` flags on `start`, `stop`, `restart`, and `delete` commands.

Tests: Add BATS tests for orphaned hostagent recovery (SIGKILL service, verify hostagent survives, restart, verify controller kills orphan and starts fresh), graceful shutdown (verify service stop terminates hostagents), restart (running VM restart, stopped VM restart, annotation on stopped VM), and `--timeout` failure (non-functional template cannot boot within deadline). Remove broken-state recovery tests. Replace `lastTransitionTime`-based recovery detection with `restartCount`. Add `RDD_VM_TYPE` support to reproduce QEMU-specific behavior on macOS.

Closes #149 
Closes #150
Closes #196

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-cli.bats
+++ b/bats/tests/33-lima-controllers/limavm-cli.bats
@@ -64,12 +64,13 @@ assert_created() {
     assert_running "false"
 
     # Start the VM (cross-namespace lookup: no --namespace flag needed)
-    run -0 rdd limavm start "test-vm"
+    # --wait=false because the template is non-functional (no real VM boots)
+    run -0 rdd limavm start --wait=false "test-vm"
     assert_output --partial 'started'
     assert_running "true"
 
     # Stop the VM
-    run -0 rdd limavm stop "test-vm"
+    run -0 rdd limavm stop --wait=false "test-vm"
     assert_output --partial 'stopped'
     assert_running "false"
 
@@ -78,6 +79,19 @@ assert_created() {
     assert_output --partial 'deleted'
     run -1 rdd ctl get limavm "test-vm" --namespace "${LIMA_TEST_NS}"
     assert_output --partial "not found"
+}
+
+@test "lima create --start sets running=true" {
+    rdd ctl create configmap "start-vm" --namespace "${LIMA_TEST_NS}" --from-literal="template=${TEMPLATE}"
+
+    # --wait=false because the template is non-functional (no real VM boots)
+    run_e -0 rdd limavm create "start-vm" "start-vm" --namespace "${LIMA_TEST_NS}" --start --wait=false
+    assert_created "start-vm" "${LIMA_TEST_NS}" "start-vm"
+
+    run -0 rdd ctl get limavm "start-vm" --namespace "${LIMA_TEST_NS}" -o jsonpath='{.spec.running}'
+    assert_output "true"
+
+    run -0 rdd limavm delete --wait "start-vm"
 }
 
 @test "lima create with file template" {
@@ -163,6 +177,37 @@ assert_created() {
     assert_fatal "${not_found}"
 }
 
+@test "lima restart sets running=true" {
+    # Create a template ConfigMap and VM
+    rdd ctl create configmap "restart-vm" --namespace "${LIMA_TEST_NS}" --from-literal="template=${TEMPLATE}"
+    run_e -0 rdd limavm create "restart-vm" "restart-vm" --namespace "${LIMA_TEST_NS}"
+    assert_created "restart-vm" "${LIMA_TEST_NS}" "restart-vm"
+
+    # --wait=false because the template is non-functional (no real VM boots)
+    run -0 rdd limavm restart --wait=false "restart-vm"
+    assert_output --partial "restart requested"
+
+    # Verify spec.running is true
+    run -0 rdd ctl get limavm "restart-vm" --namespace "${LIMA_TEST_NS}" \
+        --output jsonpath='{.spec.running}'
+    assert_output "true"
+
+    # Delete the VM
+    run -0 rdd limavm delete "restart-vm"
+    rdd ctl wait --for=delete "limavm/restart-vm" --namespace "${LIMA_TEST_NS}" --timeout="30s"
+}
+
+@test "lima start --timeout fails when VM cannot reach desired state" {
+    rdd ctl create configmap "timeout-vm" --namespace "${LIMA_TEST_NS}" \
+        --from-literal="template=${TEMPLATE}"
+    run_e -0 rdd limavm create "timeout-vm" "timeout-vm" --namespace "${LIMA_TEST_NS}"
+    assert_created "timeout-vm" "${LIMA_TEST_NS}" "timeout-vm"
+
+    # Non-functional template cannot boot, so --wait --timeout should fail
+    run -1 rdd limavm start --wait --timeout=3s "timeout-vm"
+    assert_output --partial 'context deadline exceeded'
+}
+
 @test "lima help text is displayed" {
     # lima is an alias of the limavm command
     run -0 rdd lima --help
@@ -170,6 +215,7 @@ assert_created() {
     assert_output --partial "Create a new LimaVM"
     assert_output --partial "Start a LimaVM"
     assert_output --partial "Stop a LimaVM"
+    assert_output --partial "Restart a LimaVM"
     assert_output --partial "Delete a LimaVM"
     assert_output --partial "Show LimaVM logs"
     assert_output --partial "Execute shell in Lima VM"
--- a/cmd/rdd/limavm.go
+++ b/cmd/rdd/limavm.go
@@ -18,11 +18,14 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/validation"
+	k8swatch "k8s.io/apimachinery/pkg/watch"
 	corev1scheme "k8s.io/client-go/kubernetes/scheme"
+	toolswatch "k8s.io/client-go/tools/watch"
 	"k8s.io/kubectl/pkg/cmd/util/editor"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrl "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -44,6 +47,7 @@ func newLimaVMCommand() *cobra.Command {
 		newLimaVMCreateCommand(),
 		newLimaVMStartCommand(),
 		newLimaVMStopCommand(),
+		newLimaVMRestartCommand(),
 		newLimaVMDeleteCommand(),
 		newLimaVMLogsCommand(),
 		newLimaVMShellCommand(),
@@ -119,6 +123,9 @@ func limaVMEditAction(ctx context.Context, name string) error {
 func newLimaVMCreateCommand() *cobra.Command {
 	var namespace string
 	var dryRun bool
+	var start bool
+	var wait bool
+	var timeout time.Duration
 	command := &cobra.Command{
 		Use:   "create NAME TEMPLATE",
 		Short: "Create a new LimaVM resource",
@@ -129,54 +136,124 @@ TEMPLATE can be one of:
 - A file path (if it's not a valid ConfigMap name) - creates a ConfigMap with the VM name`,
 		Args: cobra.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return limaVMCreateAction(cmd.Context(), args[0], args[1], namespace, dryRun)
+			return limaVMCreateAction(cmd.Context(), args[0], args[1], namespace, dryRun, start, wait, timeout)
 		},
 	}
 	command.Flags().StringVarP(&namespace, "namespace", "n", metav1.NamespaceDefault, "Namespace for the LimaVM resource")
 	command.Flags().BoolVar(&dryRun, "dry-run", false, "If set, do not commit any changes to the cluster")
+	command.Flags().BoolVar(&start, "start", false, "Start the VM after creation")
+	command.Flags().BoolVar(&wait, "wait", true, "Wait for the operation to complete (only applies with --start)")
+	command.Flags().DurationVar(&timeout, "timeout", 0, "Timeout for --wait (0 means wait indefinitely)")
 	return command
 }
 
 func newLimaVMStartCommand() *cobra.Command {
+	var wait bool
+	var timeout time.Duration
 	command := &cobra.Command{
 		Use:   "start NAME",
 		Short: "Start a LimaVM by setting running=true",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return limaVMSetRunningAction(cmd.Context(), args[0], true)
+			return limaVMSetRunningAction(cmd.Context(), args[0], true, wait, timeout)
 		},
 	}
+	command.Flags().BoolVar(&wait, "wait", true, "Wait for the operation to complete")
+	command.Flags().DurationVar(&timeout, "timeout", 0, "Timeout for --wait (0 means wait indefinitely)")
 	return command
 }
 
 func newLimaVMStopCommand() *cobra.Command {
+	var wait bool
+	var timeout time.Duration
 	command := &cobra.Command{
 		Use:   "stop NAME",
 		Short: "Stop a LimaVM by setting running=false",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return limaVMSetRunningAction(cmd.Context(), args[0], false)
+			return limaVMSetRunningAction(cmd.Context(), args[0], false, wait, timeout)
 		},
 	}
+	command.Flags().BoolVar(&wait, "wait", true, "Wait for the operation to complete")
+	command.Flags().DurationVar(&timeout, "timeout", 0, "Timeout for --wait (0 means wait indefinitely)")
 	return command
+}
+
+func newLimaVMRestartCommand() *cobra.Command {
+	var wait bool
+	var timeout time.Duration
+	command := &cobra.Command{
+		Use:   "restart NAME",
+		Short: "Restart a LimaVM instance",
+		Long:  "Set the restartRequested annotation and spec.running=true to restart the instance.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return limaVMRestartAction(cmd.Context(), args[0], wait, timeout)
+		},
+	}
+	command.Flags().BoolVar(&wait, "wait", true, "Wait for the operation to complete")
+	command.Flags().DurationVar(&timeout, "timeout", 0, "Timeout for --wait (0 means wait indefinitely)")
+	return command
+}
+
+func limaVMRestartAction(ctx context.Context, name string, wait bool, timeout time.Duration) error {
+	c, err := getKubeClient(ctx)
+	if err != nil {
+		return err
+	}
+	limaVM, err := findLimaVM(ctx, c, name)
+	if err != nil {
+		return err
+	}
+
+	beforeCount := limaVM.Status.RestartCount
+
+	patch := client.MergeFrom(limaVM.DeepCopy())
+	limaVM.Spec.Running = true
+	if limaVM.Annotations == nil {
+		limaVM.Annotations = make(map[string]string)
+	}
+	limaVM.Annotations[limav1alpha1.AnnotationRestartRequested] = time.Now().UTC().Format(time.RFC3339)
+
+	if err := c.Patch(ctx, limaVM, patch); err != nil {
+		return fmt.Errorf("failed to restart LimaVM: %w", err)
+	}
+
+	logrus.Infof("LimaVM %q restart requested in namespace %q", name, limaVM.Namespace)
+
+	if wait {
+		waitCtx, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout)
+		defer cancel()
+		key := client.ObjectKeyFromObject(limaVM)
+		if err := watchUntil(waitCtx, c, key, func(vm *limav1alpha1.LimaVM) bool {
+			return vm.Status.RestartCount > beforeCount
+		}); err != nil {
+			return fmt.Errorf("failed waiting for restart to complete: %w", err)
+		}
+		logrus.Infof("LimaVM %q restarted in namespace %q", name, limaVM.Namespace)
+	}
+	return nil
 }
 
 func newLimaVMDeleteCommand() *cobra.Command {
 	var wait bool
+	var timeout time.Duration
 	command := &cobra.Command{
 		Use:   "delete NAME",
 		Short: "Delete a LimaVM resource",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return limaVMDeleteAction(cmd.Context(), args[0], wait)
+			return limaVMDeleteAction(cmd.Context(), args[0], wait, timeout)
 		},
 	}
 	command.Flags().BoolVar(&wait, "wait", false, "Wait for the VM to be deleted before returning")
+	command.Flags().DurationVar(&timeout, "timeout", 0, "Timeout for --wait (0 means wait indefinitely)")
 	return command
 }
 
 // getKubeClient returns a controller-runtime client configured for the RDD control plane.
-func getKubeClient(ctx context.Context) (client.Client, error) {
+// The returned client supports Watch for event-driven waiting on resource changes.
+func getKubeClient(ctx context.Context) (client.WithWatch, error) {
 	if err := ensureServiceRunning(ctx); err != nil {
 		return nil, err
 	}
@@ -191,7 +268,7 @@ func getKubeClient(ctx context.Context) (client.Client, error) {
 	if err := limav1alpha1.AddToScheme(runtimeScheme); err != nil {
 		return nil, fmt.Errorf("failed to add LimaVM types to scheme: %w", err)
 	}
-	c, err := client.New(config, client.Options{Scheme: runtimeScheme})
+	c, err := client.NewWithWatch(config, client.Options{Scheme: runtimeScheme})
 	if err != nil {
 		return nil, fmt.Errorf("failed to create client: %w", err)
 	}
@@ -263,7 +340,7 @@ func takeOwnership(ctx context.Context, c client.Client, limaVM *limav1alpha1.Li
 	return nil
 }
 
-func limaVMCreateAction(ctx context.Context, name, template, namespace string, dryRun bool) error {
+func limaVMCreateAction(ctx context.Context, name, template, namespace string, dryRun, start, wait bool, timeout time.Duration) error {
 	c, err := getKubeClient(ctx)
 	if err != nil {
 		return err
@@ -292,7 +369,6 @@ func limaVMCreateAction(ctx context.Context, name, template, namespace string, d
 	}()
 
 	// Create the LimaVM resource with the template reference
-	running := false
 	limaVM := &limav1alpha1.LimaVM{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -302,7 +378,7 @@ func limaVMCreateAction(ctx context.Context, name, template, namespace string, d
 			TemplateRef: limav1alpha1.TemplateReference{
 				Name: configMapName,
 			},
-			Running: running,
+			Running: start,
 		},
 	}
 
@@ -324,10 +400,22 @@ func limaVMCreateAction(ctx context.Context, name, template, namespace string, d
 			createdConfigMap = nil
 		}
 	}
+
+	if start && wait && !dryRun {
+		waitCtx, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout)
+		defer cancel()
+		key := client.ObjectKeyFromObject(limaVM)
+		if err := watchUntil(waitCtx, c, key, func(vm *limav1alpha1.LimaVM) bool {
+			return apimeta.IsStatusConditionPresentAndEqual(vm.Status.Conditions, "Running", metav1.ConditionTrue)
+		}); err != nil {
+			return fmt.Errorf("failed waiting for LimaVM to start: %w", err)
+		}
+		logrus.Infof("LimaVM %q started in namespace %q", name, namespace)
+	}
 	return nil
 }
 
-func limaVMSetRunningAction(ctx context.Context, name string, running bool) error {
+func limaVMSetRunningAction(ctx context.Context, name string, running, wait bool, timeout time.Duration) error {
 	c, err := getKubeClient(ctx)
 	if err != nil {
 		return err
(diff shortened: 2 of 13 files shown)


cmd/rdd/limavm.go
```diff
@@ -346,14 +412,72 @@ func limaVMSetRunningAction(ctx context.Context, name string, running bool) erro
...
+		default:
+		}
+		var vm limav1alpha1.LimaVM
+		if err := c.Get(ctx, key, &vm); err != nil {
+			return err
+		}
+		if apimeta.IsStatusConditionPresentAndEqual(vm.Status.Conditions, "Running", status) {
+			return nil
+		}
+		time.Sleep(time.Second)
```
mook-as: This is just polling, right? `kubectl wait` seems to be using [`watch.UntilWithSync`](https://github.com/kubernetes/kubernetes/blob/330950ca52bfc99297b73f4a70199f82c0705342/staging/src/k8s.io/client-go/tools/watch/until.go#L128); would it make sense to do that?

cmd/rdd/limavm.go
```diff
@@ -346,14 +412,72 @@ func limaVMSetRunningAction(ctx context.Context, name string, running bool) erro
...
+			return fmt.Errorf("failed waiting for LimaVM to be %s: %w", action, err)
+		}
+	}
 	return nil
 }
 
-func limaVMDeleteAction(ctx context.Context, name string, wait bool) error {
+// contextWithOptionalTimeout returns a context with a deadline if timeout > 0,
+// or the original context if timeout is 0 (wait indefinitely).
+func contextWithOptionalTimeout(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
```
mook-as: This looks pretty close to [`watch.ContextWithOptionalTimeout`](https://pkg.go.dev/k8s.io/client-go/tools/watch#ContextWithOptionalTimeout); would it make sense to just use that?

cmd/rdd/limavm.go
```diff
@@ -346,14 +412,72 @@ func limaVMSetRunningAction(ctx context.Context, name string, running bool) erro
...
+		}
+		if apimeta.IsStatusConditionPresentAndEqual(vm.Status.Conditions, "Running", status) {
+			return nil
+		}
+		time.Sleep(time.Second)
+	}
+}
+
+// waitForRestartCount polls until restartCount exceeds beforeCount.
+func waitForRestartCount(ctx context.Context, c client.Client, key client.ObjectKey, beforeCount int32) error {
```
mook-as: Similarly, `UntilWithSync` is probably fine here too.

docs/design/api_lima.md
```diff
@@ -41,7 +42,7 @@ status:
     message: Lima instance started successfully
 ```
 
-- **metadata.annotations**: Can be used to request "actions" from the reconciler, like resetting (deleting and recreating with the same settings), or restarting the instance. The `status.conditions` can be used to store the state machine information.
+- **metadata.annotations**: Request actions from the reconciler, such as resetting (deleting and recreating with the same settings) or restarting the instance. The `restartRequested` annotation (set to any value) tells the reconciler to restart the instance. The reconciler translates it to `status.restartNeeded` and removes the annotation. The `status.conditions` store the state machine information.
```
mook-as: Would putting the annotation list in a table be easier to read? (We should also note that they all start with lima.rancherdesktop.io/ somewhere)

docs/design/api_lima.md
```diff
@@ -59,13 +60,18 @@ status:
 
 - **status.templateConfigMap**: Name of the ConfigMap containing the validated template. The reconciler creates this ConfigMap after copying and validating the template from `spec.templateRef`. This ConfigMap is owned by the LimaVM and deleted automatically when the LimaVM is deleted.
 
+- **status.restartNeeded**: Indicates a restart has been requested but not yet executed. The reconciler sets this field when it processes a `restartRequested` annotation or detects a template change on a running instance. When the instance is running, the reconciler stops it (clearing `restartNeeded` in the same write); the next reconcile starts it via normal logic. When the instance is already stopped, the reconciler clears the flag immediately. When the instance is starting, the reconciler waits for boot to finish before acting.
+
+- **status.restartCount**: Tracks how many times the instance has reached the Running state. The reconciler increments this counter each time it sets `Running=True/Started`. Useful for detecting recovery completion without relying on `lastTransitionTime` comparisons.
```
mook-as: That sounds like it should be `startCount` instead, depending on how you want to consider non-restart starts. Presumably we just care that it's monotonically increasing.

docs/design/api_lima.md
```diff
@@ -59,13 +60,18 @@ status:
...
+
 - **status.conditions**: Standard Kubernetes conditions tracking the LimaVM state.
 
     | Type      | Status    | Reason         | Description                                                  |
     |-----------|-----------|----------------|--------------------------------------------------------------|
     | `Created` | `Unknown` | `Pending`      | Reconciler has seen the resource; creation not yet attempted |
     | `Created` | `True`    | `Created`      | Lima instance exists on disk and is ready                    |
     | `Created` | `False`   | `CreateFailed` | Instance creation or preparation failed                      |
+    | `Running` | `False`   | `Starting`     | Hostagent launched, VM booting                               |
```
mook-as, line 80: Do we also need a `Running=False/Stopping` state?

docs/design/api_lima.md
```diff
@@ -59,13 +60,18 @@ status:
...
 
+- **status.restartNeeded**: Indicates a restart has been requested but not yet executed. The reconciler sets this field when it processes a `restartRequested` annotation or detects a template change on a running instance. When the instance is running, the reconciler stops it (clearing `restartNeeded` in the same write); the next reconcile starts it via normal logic. When the instance is already stopped, the reconciler clears the flag immediately. When the instance is starting, the reconciler waits for boot to finish before acting.
+
+- **status.restartCount**: Tracks how many times the instance has reached the Running state. The reconciler increments this counter each time it sets `Running=True/Started`. Useful for detecting recovery completion without relying on `lastTransitionTime` comparisons.
+
 - **status.conditions**: Standard Kubernetes conditions tracking the LimaVM state.
 
     | Type      | Status    | Reason         | Description                                                  |
```
mook-as, line 75: For my own understanding; presumably not worth putting it in the actual documentation:

```mermaid
stateDiagram-v2
  [*] --> Created=Unknown/Pending: Resource created
  Created=Unknown/Pending --> Created=True/Created: Lima instance created
  Created=Unknown/Pending --> Created=False/CreateFailed: Instance creation failed
  Created=True/Created --> Running=False/Starting: Host agent launched
  Running=False/Starting --> Running=True/Started: Lima instance started
  Running=False/Starting --> Running=False/StartFailed: Instance failed to start
  Running=True/Started --> Running=False/Stopped: Lima instance stopped
  Running=True/Started --> Running=False/StopFailed: Instance failed to stop
```

docs/design/api_lima.md
```diff
@@ -145,7 +151,149 @@ The mutating webhook creates the ConfigMap during admission, but cannot set an o
...
+    R->>HA: Start new hostagent
+    R->>W: Start new watcher
+    HA-->>W: Event: Running=true
+    W->>R: Enqueue reconcile (phase=Running)
+    R->>R: Increment restartCount + Set Running=True/Started
+```
+
+#### Orphan Recovery (Control Plane Restart)
+
+The control plane crashes. Hostagents survive (own process group). On restart, the reconciler has no watcher state. It detects orphaned hostagents via `store.Inspect()`, kills them, and starts fresh with a watcher.
```
mook-as: The host agent out living the control plane (even accidentally) seems like it can be an issue, if the user just never starts it again? But maybe that's fine…

docs/design/api_lima.md
```diff
@@ -145,7 +151,149 @@ The mutating webhook creates the ConfigMap during admission, but cannot set an o
...
+
+```mermaid
```
mook-as, line 244: For this diagram, Reconciler and Watcher are in the same process right? And the other two are independent processes? Well, I guess `store.Inspect` isn't really a process…

docs/design/api_lima.md
```diff
@@ -145,7 +151,149 @@ The mutating webhook creates the ConfigMap during admission, but cannot set an o
...
+    participant R as Reconciler
+    participant W as Watcher
+    participant HA as Hostagent
+    participant S as store.Inspect
+
+    Note over R: Control plane restarts (no watcher state)
+    R->>S: Inspect instance
+    S-->>R: Status=Running (orphaned hostagent)
+    R->>HA: Kill orphaned hostagent (SIGTERM→SIGKILL)
+    R->>R: Requeue
```
mook-as, line 255: Do we also need a part for when the instance is stopped (`.preparing` exists)? Probably not worth it…

docs/design/environment.md
```diff
@@ -21,6 +21,7 @@ These variables configure the BATS test framework. They have no effect on `rdd`
 | --- | --- | --- |
 | `RDD_TRACE` | Enables verbose trace output in BATS tests. | `false` |
 | `RDD_NAMESPACE` | Default Kubernetes namespace for BATS controller tests. | `default` |
+| `RDD_VM_TYPE` | Lima VM type for tests that boot a VM (`qemu` or `vz`). Useful for reproducing QEMU-specific failures on macOS. | Lima's default (`vz` on macOS, `qemu` on Linux) |
```
mook-as, line 24: Presumably, `wsl2` on Windows… we _do_ need to support that :p

pkg/apis/lima/v1alpha1/limavm_types.go
```diff
@@ -11,6 +11,10 @@ import (
 // TemplateConfigMapKey is the key used to store the template text in the templateConfigMap.
 const TemplateConfigMapKey = "template"
 
+// AnnotationRestartRequested triggers a restart when set on a LimaVM resource.
+// The reconciler translates it to status.restartNeeded and removes the annotation.
+const AnnotationRestartRequested = "lima.rancherdesktop.io/restartRequested"
```
mook-as: We should perhaps extract `SchemeGroupVersion.Group` into a package-level constant (doesn't have to be exported) and then use it to construct the annotation name, so we can avoid issues with typos.

pkg/apis/lima/v1alpha1/limavm_types.go
```diff
@@ -59,6 +63,18 @@ type LimaVMStatus struct {
 	// This field is informational - internal code should use GetTemplateConfigMapName() instead.
 	// +optional
 	TemplateConfigMap string `json:"templateConfigMap,omitempty"`
+
+	// restartNeeded indicates a restart has been requested but not yet executed.
+	// Set by the reconciler when it processes a restartRequested annotation or
+	// detects a template change on a running instance. Cleared when the restart
```
mook-as: It's cleared when the instance stops, before the restart completes. (Per the diagrams in the markdown.)

pkg/apis/lima/v1alpha1/limavm_types.go
```diff
@@ -59,6 +63,18 @@ type LimaVMStatus struct {
...
+	// completes or the instance is already stopped.
+	// +optional
+	RestartNeeded bool `json:"restartNeeded,omitempty"`
+
+	// restartCount tracks how many times the instance has reached the Running state.
+	// Incremented each time the controller sets Running=True/Started.
+	// +optional
+	RestartCount int32 `json:"restartCount,omitempty"`
```
mook-as, line 77: I guess this is technically `+kubebuilder:default=0`, but that's the zero-value anyway so we don't need a defaulter?

bats/Makefile
```diff
@@ -5,6 +5,11 @@
 # BATS integration testing targets
 BATS_CORE := ./lib/bats-core/bin/bats
 
+# Per-target timeout (seconds) to prevent hanging CI jobs.
+# Override with BATS_TIMEOUT=0 to disable.
+BATS_TIMEOUT ?= 1800
```
mook-as: Given that the PR is so big, the timeout stuff should be split into a separate PR.

pkg/service/cmd/service.go
```diff
@@ -670,9 +671,14 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
...
 
-	// Start shared controller manager if any controllers are enabled
+	// Start shared controller manager if any controllers are enabled.
+	// The WaitGroup ensures the controller manager completes shutdown
+	// (including terminating hostagent processes) before Run returns.
+	var mgrWg sync.WaitGroup
 	if len(enabledControllers) > 0 {
+		mgrWg.Add(1)
 		go func() {
+			defer mgrWg.Done()
```
mook-as, line 682: I thought one of the linters complains that you should use `mgrWg.Go(func(){…})` instead…

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
 	}
 }
 
-// updateCondition sets a condition and updates the status in one call.
+// updateCondition sets a condition and patches the status subresource. Using
+// Patch (not Update) avoids resource version conflicts when the object is
+// modified externally during long-running operations like waitForPIDFile or
+// StopGracefully.
```
mook-as: Isn't the fact that this uses `Patch` an implementation detail, and that comment should go _inside_ the function?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
 func (r *LimaVMReconciler) updateCondition(ctx context.Context, limaVM *v1alpha1.LimaVM, conditionType string, status metav1.ConditionStatus, reason, message string) error {
+	patch := client.MergeFrom(limaVM.DeepCopy())
```
mook-as, line 340: I wish we don't need this copy, but I couldn't think of anything better.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+	return r.Status().Patch(ctx, limaVM, patch)
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *LimaVMReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.instanceStates = make(map[string]*instanceState)
+	r.reconcileChan = make(chan event.TypedGenericEvent[*v1alpha1.LimaVM])
+
+	// Stop all hostagents when the manager shuts down.
+	if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
```
mook-as: This is getting hard to read. Maybe define the function first, and then use it?

pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
```diff
@@ -0,0 +1,183 @@
...
+	r.instanceStates[name] = state
+	r.instanceStatesMu.Unlock()
+
+	go r.runWatcher(watchCtx, name, namespace, haCmd, instDir, begin, state)
+}
+
+// runWatcher is the watcher goroutine. It races haCmd.Wait() against
+// events.Watch() so that process exit is detected even without an Exiting event
+// (e.g. SIGKILL).
+func (r *LimaVMReconciler) runWatcher(ctx context.Context, name, namespace string, haCmd *exec.Cmd, instDir string, begin time.Time, state *instanceState) {
```
mook-as: Isn't `haCmd` available via `state`? Wish we could pass in fewer things, but the only thing possible seems to be to pass it in a struct, which isn't really that much better other than formatting…

pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
```diff
@@ -0,0 +1,183 @@
...
+		state.mu.Lock()
+		defer state.mu.Unlock()
+
+		if ev.Status.Exiting {
+			logger.Info("Hostagent exiting")
+			state.phase = phaseStopped
+			r.enqueueReconcile(name, namespace)
+			return true // stop watching
+		}
+		if ev.Status.Running {
```
mook-as, line 112: Do we need to worry about `Degraded`? I dunno anything, I'm just reading the struct.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -102,6 +106,10 @@ type LimaVMReconciler struct {
 	Scheme   *runtime.Scheme
 	Manager  ctrl.Manager
 	Recorder events.EventRecorder
+
+	instanceStatesMu sync.RWMutex
+	instanceStates   map[string]*instanceState
```
mook-as, line 116: Now that I think about it, why do we even allow multiple VMs in the initial implementations anyway… that seems like overkill.

pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
```diff
@@ -0,0 +1,183 @@
...
+	state, ok := r.instanceStates[name]
+	r.instanceStatesMu.RUnlock()
+	if !ok {
+		return
+	}
+
+	state.cancel()
+	<-state.done
+
+	r.instanceStatesMu.Lock()
```
mook-as, line 162: There's potentially a race here (if a new VM with the same name is created), but that seems unlikely for now.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+
+// shutdownAllHostagents terminates all running hostagents during graceful shutdown.
+// It sends SIGINT to each hostagent for graceful shutdown, waits for them to exit,
+// and falls back to SIGKILL after a timeout.
+func (r *LimaVMReconciler) shutdownAllHostagents() {
+	r.instanceStatesMu.RLock()
+	states := make(map[string]*instanceState, len(r.instanceStates))
+	for name, state := range r.instanceStates {
+		states[name] = state
+	}
```
mook-as: No need to accept this, the loop is so easy that it's probably clearer; just wanted to code golf:

```suggestion
	states := maps.Clone(r.instanceStates)
```

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+	r.instanceStatesMu.RUnlock()
+
+	if len(states) == 0 {
+		return
+	}
+
+	// Send SIGINT to all hostagents in parallel for graceful shutdown.
```
mook-as, line 384: Technically this is done serially…

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+	for _, state := range states {
+		if state.cmd != nil && state.cmd.Process != nil {
+			_ = state.cmd.Process.Signal(syscall.SIGINT)
```
mook-as, line 387: Note that this is not implementable on Windows, as far as I know? [`os.Signal`](https://pkg.go.dev/os#Signal) documentation says:
> On Windows, sending os.Interrupt to a process with os.Process.Signal is not implemented; it will return an error instead of sending a signal. 

So we'll need to figure out something else for graceful shutdown on Windows (and assume that sometimes graceful shutdown is unavailable).

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +327,77 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+		}
+	}
+
+	// Wait for each hostagent process to exit. The watcher goroutine may finish
+	// before the process (events.Watch returns on context cancel), so we wait on
+	// procExited (closed by haCmd.Wait) rather than done (closed by the watcher).
+	for name, state := range states {
+		select {
+		case <-state.procExited:
+		case <-time.After(10 * time.Second):
```
mook-as: Should this be a `const` so it's easier to find?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -76,119 +78,241 @@ func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.
...
+func (r *LimaVMReconciler) handleRestartAnnotation(ctx context.Context, limaVM *v1alpha1.LimaVM) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
-	// Get current instance status
-	inst, err := store.Inspect(ctx, limaVM.Name)
-	if err != nil {
-		logger.Error(err, "Failed to inspect Lima instance")
+	if !limaVM.Status.RestartNeeded {
+		logger.Info("Restart requested via annotation, setting status.restartNeeded")
+		patch := client.MergeFrom(limaVM.DeepCopy())
```
mook-as, line 92: Surely this can use applyconfig?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -76,119 +78,241 @@ func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.
...
+//   - Starting: return and let the watcher trigger the next reconcile.
+//   - Stopped/Unknown: clear restartNeeded and fall through to normal logic.
+func (r *LimaVMReconciler) handleRestartNeeded(ctx context.Context, limaVM *v1alpha1.LimaVM, phase instancePhase) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	switch phase {
+	case phaseRunning:
+		logger.Info("Restart needed: stopping running instance")
+		inst, err := store.Inspect(ctx, limaVM.Name)
+		if err != nil {
```
mook-as: I can't tell if these need events (for debugging)…

mook-as (commented):
I haven't looked at everything due to time constraints; that also means I don't have the big picture yet, so I'm probably missing things. In particular I haven't thought through the implications of concurrent events happening.

I feel like we're baking in more things that ignore Windows completely, and it will make porting things to Windows take even longer after. (In this case, more uses of `kill` in the BATS tests plus `shutdownAllHostagents`.) Ah well, we'll figure stuff out later.

docs/design/api_lima.md
jandubois: I do agree with you that `startCount` is "more correct", but I still prefer `restartCount` because that's what it is being used for, even though the very first start is not a restart. Every other one after the first is a restart though.

docs/design/api_lima.md
jandubois: This would be purely for logging; I haven't found a reason in the code that would require this state. We can always add it when we need it.

docs/design/api_lima.md
jandubois: I'm copying the diagram into the docs for now; it looks very clear to me. We can always delete it later, if we feel it is redundant.

docs/design/api_lima.md
jandubois: Not going to worry about for now. Host agents will be deleted by reboot at the latest.

We can do something later by providing e.g. a LaunchAgent or equivalent that will monitor the control plane, and either restart on failure, or cleanup after. I suspect we won't bother with it for quite a while.

docs/design/api_lima.md
jandubois: Everything is in the same process, except for the host agent, and potentially the VM driver subprocess (QEMU).

docs/design/environment.md
jandubois: Will deal with Windows later (maybe even next week).

pkg/service/cmd/service.go
jandubois: Are you talking about `errgroup.Go()`? We don't care about error propagation, I think the code is fine. I don't see any linter warnings.

pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
jandubois: `Degraded` doesn't seem to be used anywhere. I'll add a comment that are ignoring it intentionally.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Seems better to make the logic robust for multiple instances instead of having to revisit this again later.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yes, but we send the signal to all instances without waiting in between.

However, we are not waiting for the timeouts in parallel. I have a note about this, will deal with it later. This PR is already too big.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Will have to figure it out somehow. Maybe use IPC if signals are not working.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Server-side apply is actually more complicated for deletion: we then have to start worrying about field owners etc. So it seems more complicated for no benefit.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Everything is written to the log file. We should reserve events for stuff users would use to understand how the current `Conditions` came to be.

pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
jandubois: This is not possible. The finalizer doesn't delete the old resource until the watcher is stopped, and the admission controller doesn't allow a new one to be created with a name that already exists.

jandubois:
I have added a 👍 to each comment that resulted in a change, and 👎 to each one that didn't (doesn't mean I disagree, just that I don't want to do anything in this PR).

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -102,6 +111,10 @@ type LimaVMReconciler struct {
 	Scheme   *runtime.Scheme
 	Manager  ctrl.Manager
 	Recorder events.EventRecorder
+
+	instanceStatesMu sync.RWMutex
+	instanceStates   map[string]*instanceState
```
Nino-K, line 116: Does the key have to be name + namespace?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -7,8 +7,12 @@ package controllers
 import (
```
Nino-K, line 7: Don't we need the standard file headers? 

```
// SPDX-License-Identifier: Apache-2.0
// SPDX-FileCopyrightText: SUSE LLC
// SPDX-FileCopyrightText: The Rancher Desktop Authors
```

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -316,16 +332,79 @@ func (r *LimaVMReconciler) setCondition(limaVM *v1alpha1.LimaVM, conditionType s
...
+	patch := client.MergeFrom(limaVM.DeepCopy())
 	r.setCondition(limaVM, conditionType, status, reason, message)
-	return r.Status().Update(ctx, limaVM)
+	return r.Status().Patch(ctx, limaVM, patch)
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *LimaVMReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.instanceStates = make(map[string]*instanceState)
+	r.reconcileChan = make(chan event.TypedGenericEvent[*v1alpha1.LimaVM])
```
Nino-K: Could the following scenario occur? Since the reconcile channel is unbuffered, if the controller isn’t actively receiving at that exact moment (let's say it drops), could that be dangerous? Could it miss transitions and get “stuck?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Not really because the name is guaranteed to be unique in the cluster. The admission controller does not allow you to create a second instance with a name that already exists.

We have this rule because the resource name is also being used as the instance directory name, and you can only have one instance in a directory.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: The standard header are already there (maybe you missed them because the `diff` view starts at line 7, not line 1?

bats/tests/33-lima-controllers/limavm-cli.bats
```diff
@@ -163,13 +177,45 @@ assert_created() {
...
+    run -0 rdd ctl get limavm "restart-vm" --namespace "${LIMA_TEST_NS}" \
+        --output jsonpath='{.spec.running}'
+    assert_output "true"
+
+    # Delete the VM
+    run -0 rdd limavm delete "restart-vm"
+    rdd ctl wait --for=delete "limavm/restart-vm" --namespace "${LIMA_TEST_NS}" --timeout="30s"
+}
+
+@test "lima start --timeout fails when VM cannot reach desired state" {
```
Nino-K, line 200: Should this test clean up after itself?  If the run fails, the VM is left in the namespace. Should it have a teardown, or is there one that I'm missing?

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: I think that is quite unlikely, but the safe thing to do is to make the channel buffered:

https://github.com/rancher-sandbox/rancher-desktop-daemon/compare/1e674dc52bc61b920933d91751ecbc340aec053b..776b8f770af0c7f65be604e74c47e862524724a4

pkg/controllers/lima/limavm/controllers/limavm_controller.go
Nino-K: Yes, you are right, I didn't catch that.

bats/tests/33-lima-controllers/limavm-cli.bats
jandubois: We generally don't do teardown, but make sure everything is clean during startup.

The reason for skipping teardown is that it makes it easier to debug failures when the processes still exist.

Nino-K: I just realized I forgot about the startup, which makes sense.

Nino-K (approved):

PR #202, opened 2026-02-26
mock controller: use deterministic name for image

### LINKED ISSUE #200: Race condition in mock-controller
I saw this in an unrelated PR:

```
1..3
/Users/runner/Library/Logs/rancher-desktop-bats-containers-controller/mock-controller.log
# (2026-02-26T02:51:29Z: try): 0/30 tries: rdd ctl get namespace rdd-mocks -o name
# (2026-02-26T02:51:32Z: try): 1/30 tries: rdd ctl get container 1a0da50021d66e1e037d188d3c1b130d84d1f02cdeb2a9102b1a87a79ab8fb13 -o name
ok 1 containers are created
# (2026-02-26T02:51:33Z: try): 0/30 tries: rdd ctl get namespace rdd-mocks -o name
# (2026-02-26T02:51:34Z: try): 0/30 tries: rdd ctl get image --field-selector status.repoTag=ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest --output jsonpath={.items[*].status.repoTag}
# (2026-02-26T02:51:35Z: try): 0/30 tries: rdd ctl get image --field-selector status.repoTag=registry.opensuse.org/opensuse/leap:latest --output jsonpath={.items[*].status.repoTag}
# (2026-02-26T02:51:35Z: try): 0/30 tries: rdd ctl get image --field-selector status.repoTag=registry.opensuse.org/opensuse/leap:15.6 --output jsonpath={.items[*].status.repoTag}
not ok 2 images are created
# (from function `__assert_line' in file lib/bats-assert/src/assert_line.bash, line 295,
#  from function `assert_line' in file lib/bats-assert/src/assert_line.bash, line 131,
#  in test file tests/34-containers-controllers/containers-mock.bats, line 45)
#   `assert_line "${image}"' failed
# namespace/rdd-mocks
# ghcr.io/rancher-sandbox/rancher-desktop/rdx-proxy:latest
# registry.opensuse.org/opensuse/leap:latest
# registry.opensuse.org/opensuse/leap:15.6 registry.opensuse.org/opensuse/leap:15.6
#
# -- output does not contain line --
# line   : registry.opensuse.org/opensuse/leap:15.6
# output : registry.opensuse.org/opensuse/leap:15.6 registry.opensuse.org/opensuse/leap:15.6
# --
#
# (2026-02-26T02:51:37Z: try): 0/30 tries: rdd ctl get namespace rdd-mocks -o name
# (2026-02-26T02:51:37Z: try): 0/30 tries: rdd ctl get volume data -o name
ok 3 volumes are created
```

Got AI to analyze possible reasons for the flakiness:

⏺ The root cause is a race condition in the mock image reconciler, not a flake.

  What happens:

  1. The reconciler watches both the rdd-mocks Namespace and all Image objects (Watches(&containersv1alpha1.Image{}) in SetupWithManager)
  2. When the Namespace is created, the reconciler runs and creates Image objects with GenerateName
  3. Each Image creation triggers the watcher, which re-enqueues the Namespace
  4. The reconciler runs again, calls findOrCreateImage, which searches by .status.repoTag — but the status hasn't been applied yet (it's a separate SubResource("status").Apply call)
  5. The search misses the just-created image, so the reconciler creates a duplicate
  6. Both end up with the same .status.repoTag
  7. The test queries by field selector and gets registry.opensuse.org/opensuse/leap:15.6
  registry.opensuse.org/opensuse/leap:15.6 on one line instead of one per line

  Why it showed up now: Likely timing-dependent. The CI machine's load or scheduling happened to widen the gap between image creation and status update just enough.

  Fix options (in order of simplicity):
  1. Use a deterministic image name (derived from the repoTag) instead of GenerateName — then the second reconcile updates instead of creating
  2. Set the status in the same reconcile pass before returning, so the next run's findOrCreateImage finds it
  3. Remove the Image watcher from SetupWithManager since the reconciler creates images, it doesn't need to re-reconcile when they change

mook-as:
We're seeing multiple images with the same tag in some CI runs.  The hypothesis is that:
- `findOrCreateImage` doesn't see anything, a new image is created.
- `updateImage` updates its status; however, this doesn't land in the client cache.
- The image creation triggers a new reconcile run.
- `findOrCreateImage` looks at the cache and didn't see the updated status on the newly created image.
- It creates a new image instead (and we update the status).

Avoid the issue by creating deterministic image names (using the image ID plus the hash of the repo tag), so we avoid the whole issue with trying to find the existing image.

Fixes #200

### DIFF
--- a/pkg/controllers/mock/image_reconciler.go
+++ b/pkg/controllers/mock/image_reconciler.go
@@ -6,6 +6,7 @@ package mock
 
 import (
 	"context"
+	"crypto/sha256"
 	_ "embed"
 	"encoding/json"
 	"errors"
@@ -17,9 +18,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	metav1apply "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/tools/events"
@@ -102,20 +101,21 @@ func (r *imageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 		if len(inspect.RepoTags) > 0 {
 			for _, tag := range inspect.RepoTags {
-				image, err := r.findOrCreateImage(ctx, inspect.ID, tag)
-				if err != nil {
-					errs = append(errs, fmt.Errorf("failed to find or create image %s: %w", tag, err))
-					continue
-				}
+				// Deterministically generate a name for the Image based on the
+				// tag and the image ID, to avoid having to search for an
+				// existing image since we cannot atomically update the status
+				// at the same time as creating the object.
+				name := fmt.Sprintf("%s-%32x",
+					sanitizeKubernetesObjectName(inspect.ID),
+					sha256.Sum256([]byte(tag)))
 				statusApplyCopy := *statusApplyConfig
 				errs = append(errs, r.updateImage(ctx,
-					containersv1alpha1apply.Image(image.GetName(), image.GetNamespace()).
+					containersv1alpha1apply.Image(name, metav1.NamespaceDefault).
 						WithOwnerReferences(ownerReference),
 					statusApplyCopy.
 						WithRepoTag(tag).
 						WithNamespace(containerNamespace),
-				)...,
-				)
+				)...)
 			}
 		} else {
 			// No tags; create a single dangling image.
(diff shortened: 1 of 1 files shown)


pkg/controllers/mock/image_reconciler.go
```diff
@@ -102,20 +101,21 @@ func (r *imageReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
-				image, err := r.findOrCreateImage(ctx, inspect.ID, tag)
-				if err != nil {
-					errs = append(errs, fmt.Errorf("failed to find or create image %s: %w", tag, err))
-					continue
-				}
+				// Deterministically generate a name for the Image based on the
+				// tag and the image ID, to avoid having to search for an
+				// existing image since we cannot atomically update the status
+				// at the same time as creating the object.
+				name := fmt.Sprintf("%s-%32x",
```
jandubois: I assume `%32x` was meant to truncate the hash, but that should be `%.32x`. Except specifying a precision only works on integer values, not byte arrays. You have to pass a shorter slice if you want a shorter string.

But it doesn't really matter if we use 32 characters or 64, we are still below the max length limit. And nobody is going to type the name manually either way.

```suggestion
				name := fmt.Sprintf("%s-%x",
```

jandubois (dismissed):
Thanks, LGTM

jandubois (approved):

PR #203, opened 2026-02-26
docs: Add some stuff on required setup

mook-as:


jandubois (approved):
Thanks, LGTM

I am kind of hoping that we can build on msys2 and don't have to cross-compile, but on the other hand being able to cross-compile will make it easier to build an arm64 version.

Currently cross-compilation may be blocked by the need for CGO to link to the native sqlite library. (Also by CGO for VZ for macOS binaries).

mook-as:
Yep, once we get things to work in msys2 we can update the document.  We'll always need WSL installed anyway (for RDD), so using that instead of msys2 seemed easier at this point.

PR #204, opened 2026-02-26
CI: Set timeout per BATS suite

mook-as:
This lets us provide a command to set a timeout before running each BATS suite (individually), and use that in CI to make sure we don't end up getting stuck in any particular test.  There is a CI-level timeout, but that would cause the logs to not be printed (because we use `--output-sync` which buffers the output).

Also fixes an issue with the `bats-timeout` target, which ends up always running with no tests because it's not recursive.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -163,6 +163,7 @@ jobs:
       shell: run-in-wsl {0}
       env:
         RDD_TRACE: true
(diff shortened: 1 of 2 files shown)


jandubois:
This is mostly already included in https://github.com/rancher-sandbox/rancher-desktop-daemon/pull/199/files

mook-as:
Great, then we don't need this one. Looking forward to that being un-drafted.

PR #205, opened 2026-02-28
Add per-target timeout to BATS Makefile

jandubois:


bats/Makefile
```diff
@@ -5,6 +5,11 @@
 # BATS integration testing targets
 BATS_CORE := ./lib/bats-core/bin/bats
 
+# Per-target timeout (seconds) to prevent hanging CI jobs.
+# Override with BATS_TIMEOUT=0 to disable.
+BATS_TIMEOUT ?= 1800
+BATS := $(if $(filter-out 0,$(BATS_TIMEOUT)),timeout $(BATS_TIMEOUT)) $(BATS_CORE)
```
Nino-K, line 11: Is timeout available natively on MacOs? (I know I did `brew install coreutils` to have it in the past)

Nino-K (commented):
I would like to know about the one question regarding the timeout on macOS; otherwise, LGTM.

bats/Makefile
jandubois: No, you must have GNU `coreutils` installed. This is documented in `docs/development.md`.

Nino-K (approved):

PR #206, opened 2026-02-28
Detect template ConfigMap changes and restart running instances

jandubois:
When the template ConfigMap's `resourceVersion` differs from the last observed version, compare the on-disk `lima.yaml` with the ConfigMap data. If they differ, write the new template and set `restartNeeded` so the existing restart machinery handles the stop/start cycle.

For running or starting instances, the reconciler sets `status.restartNeeded=true` and lets `handleRestartNeeded` stop the instance and restart it. For stopped instances, the template is written to disk without starting the VM.

Side effects are ordered for crash safety: persist `restartNeeded` before writing to disk, then update `observedTemplateResourceVersion` last. If any step fails, the stale observed version causes the next reconcile to re-enter `handleTemplateUpdate` and retry.

Set `observedTemplateResourceVersion` during initial instance creation and when discovering an existing instance on disk.

Test that patching the managed template ConfigMap restarts a running instance and updates a stopped instance's on-disk `lima.yaml` without starting it.

Use Lima's env: field so we can SSH into the guest after restart and confirm the variable appears in `/etc/environment` — proving the VM rebooted with the updated template, not just that the file was written to disk.


pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -323,6 +347,82 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+
+	diskTemplate, err := readInstanceTemplate(limaVM.Name)
+	if err != nil {
+		logger.Error(err, "Failed to read on-disk template")
+		return ctrl.Result{}, err
+	}
+
+	newTemplate := templateConfigMap.Data[v1alpha1.TemplateConfigMapKey]
+
+	if string(diskTemplate) == newTemplate {
```
Nino-K, line 374: nit; it looks like the file comparison is done byte for byte, which means even a new trailing line could trigger a restart.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -323,6 +347,82 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+func (r *LimaVMReconciler) handleTemplateUpdate(ctx context.Context, limaVM *v1alpha1.LimaVM, templateConfigMap *corev1.ConfigMap) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	diskTemplate, err := readInstanceTemplate(limaVM.Name)
+	if err != nil {
+		logger.Error(err, "Failed to read on-disk template")
+		return ctrl.Result{}, err
+	}
+
+	newTemplate := templateConfigMap.Data[v1alpha1.TemplateConfigMapKey]
```
Nino-K, line 372: Does this need to check for the key? 

```
newTemplate, ok := templateConfigMap.Data[v1alpha1.TemplateConfigMapKey]
if !ok {
    return ...
}
```

jandubois: I don't think so. The ConfigMap validator returns an error when the key is missing, so the ConfigMap can neither be created nor be updated with this key.

But even if it could, this would just return an empty string as the template, which would then not compare as equal to the new template, so the key would be added. But as I wrote above, this can't actually happen.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yes, that is the way it is going to be for now. We can always revisit later.

Nino-K (approved):

PR #208, opened 2026-03-03
Re-warm CRD watch caches during cleanup to prevent finalizer timeout

jandubois:
CRD deletions and finalizations trigger API group handler rebuilds that tear down watch caches for sibling CRDs. The CRD finalizer then hits 429s ("storage is (re)initializing") and enters exponential backoff that can exceed the 30-second cleanup timeout.

Move cache warming from before-delete (where it was immediately invalidated) into the wait loop, re-warming on every poll iteration for CRDs that still exist.

### DIFF
--- a/pkg/service/controllers/manager.go
+++ b/pkg/service/controllers/manager.go
@@ -562,7 +562,7 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
 
 	crdClient := apiextensionsClient.ApiextensionsV1().CustomResourceDefinitions()
 
-	var deletedCRDNames []string
+	var deletedCRDs []*apiextensionsv1.CustomResourceDefinition
 
 	for _, controller := range controllersToCleanup {
 		crdData := controller.GetCRDData()
@@ -596,8 +596,6 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
 				continue
 			}
 
-			warmCRDCache(ctx, dynamicClient, existingCRD)
-
 			klog.InfoS("Deleting unused CRD", "crd", crd.Name, "controller", controller.GetName())
 			if err := crdClient.Delete(ctx, crd.Name, metav1.DeleteOptions{}); err != nil {
 				if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


pkg/service/controllers/manager.go
```diff
@@ -596,52 +596,58 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
...
-	if len(deletedCRDNames) == 0 {
+	if len(deletedCRDs) == 0 {
 		return nil
 	}
 
-	// Wait for all deleted CRDs to be fully removed. CRD finalization is
-	// event-driven and completes in milliseconds when no instances exist,
-	// but the API server must process the finalizer before the CRD disappears.
-	klog.InfoS("Waiting for CRD deletions to complete", "count", len(deletedCRDNames))
+	// Wait for all deleted CRDs to be fully removed. On each iteration,
```
Nino-K, line 614: Right now, you only start warming after we finish deleting all CRDs and enter `PollUntilContextTimeout`.

Let's say if you delete, say, 10 CRDs, the first CRDs’ may start immediately and hit 429 before the first poll iteration warms anything (especially if it takes time to delete). Is this something to consider?

pkg/service/controllers/manager.go
```diff
@@ -596,52 +596,58 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
...
+	// re-warm the watch cache for CRDs that still exist. CRD deletions and
```
Nino-K, line 615: What I'm wondering is if it makes sense to warm immediately after each successful delete?

pkg/service/controllers/manager.go
jandubois: Yes, but I didn't see any way to prevent this. But the finalizer is doing exponential backoff, so we need to warm up the cache before the retry period gets too long.

Time to delete should be very quick; they are just https DELETE requests that will be queued and processed asynchronously. So we should be able to warm up caches before the backoff period goes beyond 1 or 2 seconds.

pkg/service/controllers/manager.go
jandubois: We would have to group deletes by API group because each following delete request for a CRD from the same API group will flush the cache again.

pkg/service/controllers/manager.go
Nino-K: I guess if the time to delete is quick, that should be fine, my concern was only if it takes time to delete.

Nino-K (approved):

PR #209, opened 2026-03-03
Serialize CRD deletion per API group to prevent cleanup timeout

jandubois:
Each CRD deletion triggers an API group handler rebuild that tears down watch caches for all sibling CRDs in that group. Deleting all CRDs at once creates a thundering herd: the CRD finalizer hits 429 "storage is (re)initializing" errors, retries conflict with handler rebuilds, and work items get permanently dropped on 409 Conflict.

Delete CRDs sequentially within each API group, waiting for full finalization before deleting the next. Different API groups are independent, so they run in parallel. Warm each CRD's watch cache before deletion so the finalizer finds a ready cache.

### DIFF
--- a/cmd/rdd/main.go
+++ b/cmd/rdd/main.go
@@ -68,11 +68,14 @@ func parseInstanceFlag() {
 	}
 }
 
-// setLogLevel sets the log level from command flags.
+// setLogLevel sets the log level from command flags or RDD_LOG_LEVEL.
 func setLogLevel(cmd *cobra.Command, _ []string) error {
 	logLevel, err := cmd.Root().Flags().GetString("log-level")
 	if err != nil {
 		return err
+	}
+	if logLevel == "" {
+		logLevel = os.Getenv("RDD_LOG_LEVEL")
 	}
 	if logLevel == "" {
 		// Default log level: warn for regular mode, debug for developer mode
--- a/cmd/rdd/main.go
+++ b/cmd/rdd/main.go
@@ -75,7 +75,7 @@ func setLogLevel(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 	if logLevel == "" {
-		logLevel = os.Getenv("RDD_LOG_LEVEL")
+		logLevel = strings.TrimSpace(os.Getenv("RDD_LOG_LEVEL"))
 	}
 	if logLevel == "" {
 		// Default log level: warn for regular mode, debug for developer mode
--- a/pkg/service/controllers/manager.go
+++ b/pkg/service/controllers/manager.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apiextensions-apiserver/pkg/apihelpers"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


pkg/service/controllers/manager.go
```diff
@@ -596,58 +606,101 @@ func (scm *SharedControllerManager) cleanupUnusedCRDs(ctx context.Context, contr
...
+	defer cleanupCancel()
+
+	var wg sync.WaitGroup
+	errCh := make(chan error, len(crdsByGroup))
+
+	for group, crds := range crdsByGroup {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if err := deleteAndWaitForCRDs(cleanupCtx, crdClient, dynamicClient, group, crds); err != nil {
```
Nino-K, line 635: It looks like we are capturing the same `group` and `crd` variables as the loop iterates more. We might endup deleting the wrong CRD group.

```
for group, crds := range crdsByGroup {
    wg.Add(1)

    go func(groupName string, groupCRDs []*apiextensionsv1.CustomResourceDefinition) {
        defer wg.Done()

        if err := deleteAndWaitForCRDs(cleanupCtx, crdClient, dynamicClient, groupName, groupCRDs); err != nil {
            errCh <- err
        }
    }(group, crds)
}
```

pkg/service/controllers/manager.go
```diff
@@ -660,16 +713,13 @@ func warmCRDCache(ctx context.Context, dynamicClient dynamic.Interface, crd *api
...
-	// The cache typically initializes in under 25ms, so a few short retries suffice.
-	// Only retry on 429 (storage initializing); any other result means the cache is ready
-	// or the error is unrelated and not worth blocking on.
-	err = wait.PollUntilContextTimeout(ctx, 10*time.Millisecond, 5*time.Second, true, func(ctx context.Context) (bool, error) {
-		_, listErr := dynamicClient.Resource(gvr).List(ctx, metav1.ListOptions{Limit: 1})
-		return !apierrors.IsTooManyRequests(listErr), nil
+	// Retry until the List succeeds, meaning the watch cache is initialized.
+	// A 429 ("storage is (re)initializing") means cache creation is still in
+	// progress; keep retrying so the CRD finalizer finds a warm cache.
+	_ = wait.PollUntilContextTimeout(ctx, 500*time.Millisecond, 30*time.Second, true, func(ctx context.Context) (bool, error) {
```
Nino-K: What happens if it fails for other reasons, like `RBAC`, API Server unavailable, network, etc? It looks like it spins until the timeout?? just wanting to make sure that's what we want.

Nino-K: It looks like we are ignoring the `err` from `PollUntilContextTimeout`. Does it make sense to log this also?

cmd/rdd/main.go
```diff
@@ -68,12 +68,15 @@ func parseInstanceFlag() {
...
 
-// setLogLevel sets the log level from command flags.
+// setLogLevel sets the log level from command flags or RDD_LOG_LEVEL.
 func setLogLevel(cmd *cobra.Command, _ []string) error {
 	logLevel, err := cmd.Root().Flags().GetString("log-level")
 	if err != nil {
 		return err
 	}
+	if logLevel == "" {
+		logLevel = os.Getenv("RDD_LOG_LEVEL")
```
Nino-K: We should consider `strings.Trim` and `strings.ToLower` to avoid weirdness like `TRACE` or `" trace "`.

pkg/service/controllers/manager.go
jandubois: Since Go 1.22 range loop variables are per iteration and not shared, so each closure captures the right variable. We do not support Go 1.21 or older, so the code is correct as-is.

cmd/rdd/main.go
jandubois: Will call `strings.Trim`. `logrus` already does the `strings.ToLower` itself, so we don't need to repeat it.

Nino-K (approved):

PR #210, opened 2026-03-05
Fix stale Running condition after controller restart

jandubois:
- Add `rdd ctl await`, a condition-aware wait command with `--reason` and `--since` filters that `kubectl wait` lacks
- Reset the `Running` condition to `Unknown` on controller restart so consumers never act on stale status
- Replace `try`-polling helpers in BATS tests with `rdd ctl await`

### Problem

After a service restart, the `Running` condition persisted in kine reflects the previous run's state. Tests polling with `kubectl get` + jsonpath could read the stale condition before the controller reconciled, causing flaky assertions.

### Approach

**`rdd ctl await`** watches a resource via List+Watch (with automatic 410 Gone recovery) and supports `--reason` and `--since` flags. `--since=startup` resolves to the discovery ConfigMap's `creationTimestamp`, which the serve command recreates on every startup.

**`ensureServiceRunning`** waits for the discovery ConfigMap after both the "I started the service" and "service is already running" paths. The former also checks that the ConfigMap's `creationTimestamp` postdates the current startup, so `--since=startup` never reads a stale timestamp from a previous run.

**Controller reset** sets `Running` to `Unknown`/`Reconciling` on first reconcile when no watcher exists, preventing consumers from seeing a stale `True`/`False`.

### DIFF
--- a/cmd/rdd/service.go
+++ b/cmd/rdd/service.go
@@ -6,16 +6,23 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/developer"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
 	service "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/service/cmd"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/service/controllers"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/util/tail"
 )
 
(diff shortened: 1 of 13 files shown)


cmd/rdd/service.go
```diff
@@ -69,15 +75,61 @@ func ensureServiceRunning(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create kubernetes client: %w", err)
+	}
+
+	return wait.PollUntilContextCancel(ctx, 500*time.Millisecond, true, func(ctx context.Context) (bool, error) {
+		cm, err := client.CoreV1().ConfigMaps(controllers.RDDSystemNamespace).Get(
+			ctx, controllers.ControllerManagerConfigMapName, metav1.GetOptions{},
+		)
+		if err != nil {
+			return false, nil // Retry on transient errors.
```
Nino-K: That comment doesn't seem to be true, this relies on all errors. Maybe we should explicitly check the error, just in case you come across a non-retriable error.

Nino-K (commented):
From what I can see from the changes, it looks like the PR fixes the race for the code path that starts the service via `startAndWaitForReady()`, but not for callers that arrive during an already in progress startup. Is this something to consider?

Would it make sense to make `ensureServiceRunning()` also wait for a fresh discovery of ConfigMap, not only the `I am the one starting service` path?

Nino-K:
Also, I got a little confused by the PR's description says `--since=startup` but the implementation actually uses the ConfigMap’s `creationTimestamp`.

Nino-K (approved):

PR #211, opened 2026-03-05
Fix kine startup and shutdown ordering to reduce gRPC warnings

jandubois:
Start kine before `BuildGenericConfig`, which opens gRPC connections to the etcd backend immediately. Use a child context cancelled only after the controller manager finishes shutting down, so etcd client connections drain before the socket disappears.

Remove the `Datastore` field from the service config structs — nothing reads it now that kine starts directly in `Run()`.

### DIFF
--- a/pkg/service/cmd/service.go
+++ b/pkg/service/cmd/service.go
@@ -603,6 +603,25 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
 	klog.Infof("Version: %+v", version.Get())
 	klog.InfoS("Golang settings", "GOGC", os.Getenv("GOGC"), "GOMAXPROCS", os.Getenv("GOMAXPROCS"), "GOTRACEBACK", os.Getenv("GOTRACEBACK"))
 
+	// Start kine before NewConfig because BuildGenericConfig opens gRPC
+	// connections to the etcd backend immediately. Kine uses a child context
+	// cancelled after the controller manager shuts down, so the etcd client
+	// connections drain before the socket disappears.
+	var kineCancel context.CancelFunc
+	if opts.Datastore.Enabled {
+		dsConfig, err := datastore.NewConfig(opts.Datastore)
+		if err != nil {
+			return err
+		}
+		var kineCtx context.Context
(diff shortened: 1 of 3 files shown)


pkg/service/cmd/service.go
```diff
@@ -603,6 +603,25 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
...
+	// so kine outlives the kube-apiserver during shutdown — otherwise the
+	// etcd client channels produce gRPC warnings when the socket disappears.
+	var kineCancel context.CancelFunc
+	if opts.Datastore.Enabled {
+		dsConfig, err := datastore.NewConfig(opts.Datastore)
+		if err != nil {
+			return err
+		}
+		var kineCtx context.Context
+		kineCtx, kineCancel = context.WithCancel(context.Background())
```
Nino-K: Instead of using a separate context here, have you considered using the parent `ctx` but delaying the cancellation after `mgrWg.Wait()`?

pkg/service/cmd/service.go
```diff
@@ -603,6 +603,25 @@ func Run(ctx context.Context, opts options.CompletedOptions) error {
...
+		defer kineCancel()
+		klog.Infof("Starting kine/sqlite server listening on %s", opts.Datastore.EndpointConfig.Listener)
+		if err := datastore.NewServer(dsConfig.Complete()).Run(kineCtx); err != nil {
```
Nino-K, line 620: If I'm reading this correctly, it looks like we are building the datastore config twice? Once earlier for the early kine startup, and then we rebuild the entire service config afterwards.
Suggestion:
```
	config, err := options.NewConfig(opts)
	if err != nil {
		return err
	}
	completed, err := config.Complete()
	if err != nil {
		return err
	}
```

Then we can use completed; 

```
	if completed.Datastore.Config != nil {
        ...
		if err := datastore.NewServer(completed.Datastore).Run(kineCtx); err != nil {
			return fmt.Errorf("failed to start kine server: %w", err)
		}
	}
```

Nino-K (approved):

PR #214, opened 2026-03-06
Use instance-offset ports and resolve them at bind time

jandubois:
External controllers (`rdd-controller`, `mock-controller`) defaulted to ports `8080`/`8081` without instance offsets, causing collisions when parallel CI test suites started controllers for different instances on the same machine.

Replace `GetAvailablePort` with a two-part fix:

1. All port defaults now include instance offsets. Each instance reserves 4 consecutive ports starting at `8080` (external metrics, external health, embedded metrics, embedded health), plus per-instance offsets for webhook, passthrough, and apiserver ports.

2. Ports are resolved in `SharedControllerManager.Start()`, immediately before `controller-runtime` binds them. The old approach checked ports in `NewSharedControllerManager` or earlier, leaving a multi-second window (during CRD installation) for another process to claim the port.

`resolvePort` tries the desired port first and falls back to a system-assigned port if it is occupied.

### DIFF
--- a/cmd/rdd/kubectl.go
+++ b/cmd/rdd/kubectl.go
@@ -33,7 +33,7 @@ func newCtlCommand() *cobra.Command {
 
 func ctlAction(cmd *cobra.Command, args []string) error {
 	if !service.Exists() {
-		if err := service.Create(cmd.Context(), nil); err != nil {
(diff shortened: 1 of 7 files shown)


Nino-K (approved):
LGTM

PR #216, opened 2026-03-06
Document the controller manager discovery ConfigMap

jandubois:
Add `docs/design/discovery.md` describing the `rdd-controller-manager` ConfigMap schema and how it is used for service readiness, passthrough routing, external controller startup, and health gating.


docs/design/discovery.md
```diff
@@ -0,0 +1,49 @@
+# Controller Manager Discovery
+
+Controller managers register themselves in a shared ConfigMap so that other components can find them. The control plane uses this to determine which controllers are running, to route passthrough requests, and to detect when an external controller manager has gone away.
+
+## ConfigMap
```
Nino-K, line 5: nit; I would call this `ConfigMap Schema` instead of `ConfigMap`, but totally optional.

Nino-K (approved):

PR #217, opened 2026-03-06
Add app controller

Nino-K:
This PR adds the initial scaffolding for the app controller. Most of the changes are derived from the demo controller.


pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/events"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	AppName      = "rancher-desktop"
+	AppNamespace = "rancher-desktop"
```
jandubois: AppName should be `app`. AppNamespace should not be needed for a cluster-scoped resource.

The `App` resource has a `spec.Namespace` field, but that specifies where the LimaVM should be created. It is not where the `App` resource itself lives because it isn't namespaced.

pkg/controllers/app/app/crd.yaml
```diff
@@ -0,0 +1,121 @@
...
+    controller-gen.kubebuilder.io/version: v0.19.0
+  name: apps.app.rancherdesktop.io
+spec:
+  group: app.rancherdesktop.io
+  names:
+    kind: App
+    listKind: AppList
+    plural: apps
+    singular: app
+  scope: Namespaced
```
jandubois: This seems out-of-sync with the type definition. You need to run `make generate`. I've created a PR to check that generated files are up-to-date in our linting CI.

pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -0,0 +1,66 @@
...
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AppSpec defines the desired state of App.
+type AppSpec struct {
+    // running specifies whether the VM should be running.
+    // +kubebuilder:default=false
+    Running bool `json:"running"`
+	// Namespace is the namespace where this cluster-scoped App resource
+	// creates and manages its owned namespaced resources (e.g., LimaVMs).
```
jandubois:
```suggestion
	// creates and manages its owned namespaced resources (e.g., rancher-desktop).
```

jandubois (commented):
Just some early comments. Also plenty of linter feedback...

pkg/controllers/app/app/controller.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func newController() base.Controller {
+	return &controller{}
+}
+
+// Verify that controller implements base.Controller interface.
+var (
+	_ base.Controller            = &controller{}
+	_ base.PassthroughController = &controller{}
```
jandubois: Not sure why you implement the PassthroughController interface when you don't have any passthrough endpoints. I think it is premature to add this right now.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -0,0 +1,86 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
+)
+
+const (
+	appName = "rancher-desktop"
```
jandubois: The only allowed app name for the singleton is `app`.

```suggestion
	appName = "app"
```

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+	ConditionCreated = "Created"
```
jandubois, line 26: This is not being used. I assume this will happen in the next PR when you actually create a LimaVM resource.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var app v1alpha1.App
+	if err := r.Get(ctx, client.ObjectKey{Name: appName}, &app); err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			log.Error(err, "Unable to fetch App singleton")
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Handle deletion with finalizer
```
jandubois: There is no finalizer yet. But you will need one once you own resources, to clean them up, because we don't have a garbage collector.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -0,0 +1,86 @@
...
+	if base.IsBeingDeleted(&app) {
+		log.Info("App resource is being deleted, performing cleanup")
+		return ctrl.Result{}, nil
+	}
+
+	if apimeta.IsStatusConditionTrue(app.Status.Conditions, ConditionRunning) {
+		log.Info("App is running")
```
jandubois, line 55: It is not clear to me how we want to define `ConditionRunning`. Does it mean the LimaVM is running? If not, then how is `ConditionCreated` different from `ConditionRunning`?

Anyways, this doesn't really matter until the App owns a LimaVM resource.

pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -0,0 +1,68 @@
...
+func init() {
+	SchemeBuilder.Register(&App{}, &AppList{})
+}
+
+// GetResourceNamespace implements the base.ResourceNamespace interface.
+// It returns the namespace where this cluster-scoped App resource creates
+// and manages its owned namespaced resources.
+func (d *App) GetResourceNamespace() string {
+	if d.Spec.Namespace != "" {
+		return d.Spec.Namespace
```
jandubois: Should we use `a` instead of `d` for the receiver?

```suggestion
func (a *App) GetResourceNamespace() string {
	if a.Spec.Namespace != "" {
		return a.Spec.Namespace
```

jandubois (changes requested):
Looks good. Most comments are about cosmetics, but the AppName thing seems like a bug.

pkg/controllers/app/app/controller.go
```diff
@@ -0,0 +1,81 @@
...
+
+// APIGroup is the API group this controller belongs to.
+const APIGroup = "app"
+
+//go:embed crd.yaml
+var appCRD string
+
+// controller implements the base.Controller interface for app.
+type controller struct {
+	passthroughs map[string]http.Handler
```
jandubois: I think this can be removed.

```suggestion
```

pkg/controllers/app/app/controller.go
```diff
@@ -0,0 +1,81 @@
...
+	return appCRD
+}
+
+func (c *controller) GetPassthroughEndpoints() []string {
+	return slices.Collect(maps.Keys(c.passthroughs))
+}
+
+func (c *controller) GetPassthroughHandler(endpoint string) http.Handler {
+	return c.passthroughs[endpoint]
+}
```
jandubois: These methods are only needed for the PassthroughController interface.

jandubois (changes requested):
Some more cleanup because of the removal of the PassthroughController interface.

jandubois (approved):
Thanks, LGTM

PR #218, opened 2026-03-07
Verify generated files in CI

jandubois:
Add a verify-generate job to the lint workflow that runs `make generate` and fails if the working tree is dirty afterward. This catches stale CRDs and deepcopy files before they merge.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -31,6 +31,19 @@ jobs:
         --jq '.[0].number // 0')
       env:
         GH_TOKEN: ${{ github.token }}
+  verify-generate:
+    name: Verify generated files
+    needs: get-pr-number
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/lint.yaml
```diff
@@ -31,6 +31,19 @@ jobs:
         --jq '.[0].number // 0')
       env:
         GH_TOKEN: ${{ github.token }}
+  verify-generate:
+    name: Verify generated files
+    needs: get-pr-number
+    runs-on: ubuntu-latest
+    if: ${{ ! needs.get-pr-number.outputs.number }}
```
Nino-K, line 38: Does it make sense to make the condition explicit, because right now it's confusing if we are dealing with a boolean, a string, etc? 

```
if: ${{ needs.get-pr-number.outputs.number == '0' }}
```

Nino-K (approved):

PR #221, opened 2026-03-09
Get PR numbers with more explicit condition

### LINKED ISSUE #220: Change the get PR numbers condition more explicit
We may want to consider making the condition more explicit to avoid confusion.

Currently, the workflow uses:
```
if: ${{ ! needs.get-pr-number.outputs.number }}
```
Since number is a string output, relying on truthiness here can be a bit unclear and harder to reason about for future readers of the workflow.

Instead, we could use an explicit comparison such as:
```
if: ${{ needs.get-pr-number.outputs.number == '0' }}
```
This makes the intent clearer and avoids relying on implicit expression coercion.

Nino-K:
Change the `get-pr-number` condition to be more explicit to avoid confusion.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/220

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -35,7 +35,7 @@ jobs:
   bats:
     name: BATS
     needs: get-pr-number
-    if: ${{ ! needs.get-pr-number.outputs.number }}
+    if: ${{ needs.get-pr-number.outputs.number == '0' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jandubois:
Please check https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/220#issuecomment-4026900958 and following comments. I think this PR by itself is not very useful. If we touch this, then we should review the whole part that determines the PR number.

@mook-as is there a reason we need the complex step to determine the PR number when there is already a context variable with it?

mook-as:
IIRC, because the workflows could be triggered from non-PR contexts (e.g. nightly BATS runs); I do not recall if we looked harder to see if we can do different fallback though.

jandubois:
> because the workflows could be triggered from non-PR contexts

Doesn't that just mean the PR number variable would be empty, or not exist, and the condition to check that it is not false would just fail the way it does now?

mook-as:
Maybe? We should test that, and maybe we can drop the thing. I _thought_ we needed it, but I am no longer sure.

jandubois:
It looks like we don't need this.

PR #222, opened 2026-03-10
Add `rdd yq` subcommand

### LINKED ISSUE #70: Create `rdd yq` command
Tests are getting more complicated, and having `jq` or `yq` always available would simplify them.

We include virtually all of `yqlib` when we reference Lima functionality, so we can add an `rdd yq` subcommand basically for free, just like `limactl yq` (also just like `rdd kubectl`).

jandubois:
Delegate to mikefarah/yq, which is already a transitive dependency via Lima. Adapted from Lima's limactl yq implementation.

### DIFF
--- a/cmd/rdd/yq.go
+++ b/cmd/rdd/yq.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-FileCopyrightText: Copyright (c) 2017 Mike Farah
+
+// This file has been adapted from https://github.com/lima-vm/lima/blob/master/cmd/yq/yq.go
+
+package main
+
+import (
+	"os"
+
+	command "github.com/mikefarah/yq/v4/cmd"
+	"github.com/spf13/cobra"
+)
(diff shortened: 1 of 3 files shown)


cmd/rdd/yq.go
```diff
@@ -0,0 +1,43 @@
...
+		Use:                "yq",
+		Short:              "Evaluate and filter YAML documents",
+		Args:               cobra.ArbitraryArgs,
+		DisableFlagParsing: true,
+		RunE:               yqAction,
+	}
+}
+
+func yqAction(*cobra.Command, []string) error {
+	os.Args = os.Args[1:]
```
Nino-K, line 29: This is bypassing the abstraction that Cobra already provides. We should be using `args` that Cobra already provides. All we have to do is to change the args of the `yaAction` to `func yqAction(_ *cobra.Command, args []string)`.

```
func yqAction(_ *cobra.Command, args []string) error {
    cmd := command.New()

    if len(args) > 0 {
        _, _, err := cmd.Find(args)
        if err != nil && args[0] != "__complete" {
            args = append([]string{"eval"}, args...)
        }
    }

    cmd.SetArgs(args)
    if err := cmd.Execute(); err != nil {
        os.Exit(1)
    }
    return nil
}
```
Also, I think that way we can avoid mutating the entire process argv by doing 
```
os.Args = os.Args[1:]
```

Nino-K (approved):

PR #223, opened 2026-03-10
Read templateRef ConfigMap directly from the API server

jandubois:
The defaulter webhook looked up the templateRef ConfigMap through the informer cache. When the CLI creates a ConfigMap and immediately creates a LimaVM, the informer may not have synced the ConfigMap yet, causing a "not found" error. Use the API reader to bypass the cache.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-cli.bats
+++ b/bats/tests/33-lima-controllers/limavm-cli.bats
@@ -43,7 +43,7 @@ assert_created() {
     # Create VM with ConfigMap template
     run_e -1 rdd limavm create "test-missing-vm" "test-missing-template" --namespace "${LIMA_TEST_NS}"
     assert_stderr_line --partial 'denied the request'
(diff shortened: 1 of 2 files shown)


Nino-K (commented):
The BATS changes do not reflect the and verify the race we are trying to fix. It looks like it just relaxes the error assertion. 

```
- assert_stderr_line --partial 'ConfigMap \"test-missing-template\" not found'
+ assert_stderr_line --partial '\"test-missing-template\" not found'
```

Nino-K (approved):

Nino-K:
> The BATS changes do not reflect the and verify the race we are trying to fix. It looks like it just relaxes the error assertion.
> 
> ```
> - assert_stderr_line --partial 'ConfigMap \"test-missing-template\" not found'
> + assert_stderr_line --partial '\"test-missing-template\" not found'
> ```

Talked to Jan offline, and these changes are doing the right thing.

PR #224, opened 2026-03-10
Stop deleting unused CRDs on startup

jandubois:
CRD cleanup during startup caused flaky timeouts: deleting a CRD triggers an API group handler rebuild that tears down watch caches for sibling CRDs. The CRD finalizer then hits "storage is (re)initializing" errors and enters exponential backoff, exceeding the cleanup timeout.

Removing unused CRDs is also unsafe when instances still exist in kine from a previous run: the instances may have finalizers that only their controller can process, so neither the CRD finalizer nor manual deletion can clean them up without the controller running.

Unused CRDs are inert. The API endpoints remain, but without a controller, nothing reconciles them. If the controller is later re-enabled, installControllerCRDs updates the CRD schema in place.

Update the BATS start-parameters tests to verify active controllers via the discovery ConfigMap instead of CRD presence.

### DIFF
--- a/bats/tests/20-service/4-start-parameters.bats
+++ b/bats/tests/20-service/4-start-parameters.bats
@@ -1,59 +1,25 @@
 load '../../helpers/load'
 
-# Controller testing using CRD presence validation
+# Verify that --controllers selects which controllers run.
+# The discovery ConfigMap tracks enabled controllers; CRDs persist
+# across restarts and cannot indicate which controllers are active.
 
-CRD_GROUP=customresourcedefinition.apiextensions.k8s.io # spellchecker:ignore
-
-get_controllers() {
-    run -0 rdd ctl get CustomResourceDefinition --output=name
-    # If `$output` is empty, `refute_lines` will fail because `$lines` is unset.
-    if [[ -z "${output}" ]]; then
-        run -0 echo '<no output>'
-    fi
-}
-
-check_rdd_controllers() {
-    local assert=$1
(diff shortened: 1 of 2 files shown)


bats/tests/20-service/4-start-parameters.bats
```diff
@@ -1,59 +1,25 @@
...
-    # Essential APIs should work but no controllers
-    run -0 rdd ctl get namespaces
-    get_controllers
-    check_rdd_controllers refute
-    check_app_controllers refute
+# Verify that --controllers selects which controllers run.
+# The discovery ConfigMap tracks enabled controllers; CRDs persist
+# across restarts and cannot indicate which controllers are active.
+
+get_enabled_controllers() {
```
Nino-K, line 7: It looks as though this is assuming the ConfigMap is already present and propagated by the time this test checks it. If discovery registration is slower and lags slightly behind `rdd svc start`, this could still cause flakiness. 

Before, the tests were flaky because of the CRD installation timing, now they might be flaky because of the discovery publication timing. Am I understanding this correctly?

jandubois: Yes, that is correct. Which is why #210 is now waiting for the discovery configmap to exist before declaring the cluster ready.

So unless you run `rdd svc start --wait=false` you can be assured that the configmap exists and `rdd ctl await --since=startup` will do the right thing.

Nino-K (approved):

PR #229, opened 2026-03-11
Add support for running BATS tests under MSYS2

jandubois:
The test harness assumed WSL2 as the only Windows environment. MSYS2 (both MSYS and MINGW shells) requires different path conversion and environment handling.

Changes:
- Detect MSYS2 via uname (MSYS* and MINGW* patterns)
- Add is_msys() helper to distinguish MSYS2 from WSL
- Export MSYS_NO_PATHCONV=1 to prevent argument path mangling
- Convert drive-letter paths (/c/...) via cygpath in rdd() wrapper
- Rename win_to_wsl_exports to win_to_posix_exports
- Resolve TMPDIR via cygpath on MSYS2, wslpath on WSL

### DIFF
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -30,7 +30,13 @@ ifneq ($(shell sh -c 'command -v winver.exe'),)
 # On Windows, make sure we pass the environment variables back to RDD
 export WSLENV := $(WSLENV):RDD_INSTANCE:RDD_DEVELOPER_MODE
 # Use the Windows temporary directory
+ifeq ($(shell sh -c 'command -v wslpath'),)
+# MSYS2: convert Windows TEMP path using cygpath
+export TMPDIR := $(shell cygpath -u '$(TEMP)')
+else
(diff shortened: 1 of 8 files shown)


bats/Makefile
```diff
@@ -30,8 +30,14 @@ ifneq ($(shell sh -c 'command -v winver.exe'),)
 # On Windows, make sure we pass the environment variables back to RDD
 export WSLENV := $(WSLENV):RDD_INSTANCE:RDD_DEVELOPER_MODE
 # Use the Windows temporary directory
+ifeq ($(shell sh -c 'command -v wslpath'),)
```
Nino-K, line 33: Optional; It might be more robust to use `uname` instead of `wslpath`. The output would be something like: `MSYS_*`,`MINGW_*`, `CYGWIN_*` or for wsl `*-WSL2`

Nino-K (approved):
LGTM

PR #233, opened 2026-03-12
Add design doc for embedded resource scheme

jandubois:
Controllers can embed binary resources in the rdd binary and expose them to other controllers via an `embed://` URL scheme. An in-process registry provides zero-overhead access when provider and consumer share a process; the API server passthrough proxy handles cross-process access.

### DIFF
--- a/docs/design/embed.md
+++ b/docs/design/embed.md
@@ -1,0 +1,88 @@
+# Embedded Resources
+
+Controllers can embed binary resources (disk images, data files) in the `rdd` binary using Go's `embed` package. Other controllers access these resources through an `embed://` URL scheme, regardless of whether they run in the same process or on a different machine.
+
+The goal is to ship `rdd` as a single binary with all resources embedded. Go's `embed.FS` memory-maps the data, so embedded files do not consume heap memory.
+
+## URL Format
+
+```
+embed://<controller>/<path>
+```
+
+The `<controller>` is the controller name (e.g., `demo`, `limavm`). The `<path>` identifies the file within the provider's filesystem.
+
+Example: `embed://demo/opensuse.iso`
+
+## Provider Side
+
+A controller embeds resources by calling `embed.Register()` once during `init()`:
+
+```go
+//go:embed data
+var dataFS embed.FS
+
+func init() {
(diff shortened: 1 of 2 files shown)


docs/design/embed.md
```diff
@@ -0,0 +1,88 @@
+# Embedded Resources
+
+Controllers can embed binary resources (disk images, data files) in the `rdd` binary using Go's `embed` package. Other controllers access these resources through an `embed://` URL scheme, regardless of whether they run in the same process or on a different machine.
+
+The goal is to ship `rdd` as a single binary with all resources embedded. Go's `embed.FS` memory-maps the data, so embedded files do not consume heap memory.
+
+## URL Format
+
+```
+embed://<controller>/<path>
```
mook-as, line 10: Should this be `x-embed://` or `rdd-embed://` to avoid conflicts? It probably doesn't matter, feel free to ignore this comment.

docs/design/embed.md
```diff
@@ -0,0 +1,88 @@
...
+## Provider Side
+
+A controller embeds resources by calling `embed.Register()` once during `init()`:
+
+```go
+//go:embed data
+var dataFS embed.FS
+
+func init() {
+    embed.Register("demo", dataFS)
```
mook-as, line 26: That is not an existing function in [package `embed`](https://pkg.go.dev/embed@go1.26.1#pkg-index); presumably that should be `rddembed.Register` or something along those lines, especially since we use `embed.FS` just a bit earlier…

jandubois: This was just an early brain-dump that needs more fleshing out. The idea is that the `App` controller can embed the `distro.tgz` and use `embed:/app/distro.tgz` in a template file, and the `LimaVM` controller would be able to fetch the image. Same thing for the `lima-guestagent` or any other bundled files.

The passthrough mechanism is useful because the distro image belongs to the `App` controller, but the `LimaVM` controller needs to be able to access it. And they can potentially run in different processes.

Of course this is a bit of overengineering for the limited use case we have so far, so we may defer this unless it ends up being easy to implement.

jandubois:
Doesn't look like we need this anytime soon, so closing it as "deferred".

PR #234, opened 2026-03-12
Preserve Lima serial logs across VM restarts

### LINKED ISSUE #247: Preserve `ha.stderr.log` files in CI
There is already code implementing log rotation for them, and `RDD_KEEP_LOGS` prevents `rdd svc delete` from deleting the Lima logs, but the controller still calls `instance.Delete()` from Lima itself, which will delete the instance directory anyways, so CI never gets to collect any logs.

jandubois:
Serial logs (`serial.log`, `serialp.log`, `serialv.log`) are overwritten by the VM driver on each start. Add `logfile.Rotate()` to rename the active log to a numbered backup without creating a new file, and call it for serial logs in `startInstance()` before the hostagent launches.

Extract shared rotation logic from `Create()` into a `rotate()` helper so both `Create` and `Rotate` use the same numbering and pruning code.

Fixes #247

### DIFF
--- a/pkg/instance/logs.go
+++ b/pkg/instance/logs.go
@@ -27,9 +27,11 @@ func PreserveLogs(srcDir, instanceName string) (int, error) {
 
 	// Filter to .log files before creating the destination directory,
 	// so we don't leave empty directories when no logs exist.
+	// Symlinked log files are not expected and not supported; they
+	// would become dangling after the source directory is removed.
 	var logFiles []os.DirEntry
 	for _, entry := range entries {
-		if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".log") {
+		if entry.Type().IsRegular() && strings.HasSuffix(entry.Name(), ".log") {
 			logFiles = append(logFiles, entry)
 		}
 	}
@@ -47,20 +49,22 @@ func PreserveLogs(srcDir, instanceName string) (int, error) {
 		return 0, err
 	}
 
+	var errs []error
 	count := 0
 	for _, entry := range logFiles {
 		src := filepath.Join(srcDir, entry.Name())
 		dst := filepath.Join(destDir, entry.Name())
 		if err := os.Rename(src, dst); err != nil {
-			if count == 0 {
-				os.Remove(destDir)
-			}
-			return count, fmt.Errorf("preserve %s: %w", entry.Name(), err)
+			errs = append(errs, fmt.Errorf("preserve %s: %w", entry.Name(), err))
+			continue
 		}
 		count++
 	}
+	if count == 0 {
+		os.Remove(destDir)
+	}
 
-	return count, nil
+	return count, errors.Join(errs...)
 }
 
 // nextAvailableDir creates a directory named {name} under parent. If it
--- a/pkg/instance/logs.go
+++ b/pkg/instance/logs.go
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package instance
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// PreserveLogs moves .log files from srcDir into a uniquely named
+// subdirectory of LogDir(). The subdirectory uses instanceName as a
+// base, with a numeric suffix to avoid collisions (.2, .3, etc.).
+//
+// Returns the number of files moved. Skips srcDir entries that are
+// directories or do not end in ".log".
+func PreserveLogs(srcDir, instanceName string) (int, error) {
+	entries, err := os.ReadDir(srcDir)
+	if err != nil {
+		return 0, fmt.Errorf("read instance directory %s: %w", srcDir, err)
+	}
+
+	// Filter to .log files before creating the destination directory,
+	// so we don't leave empty directories when no logs exist.
+	var logFiles []os.DirEntry
+	for _, entry := range entries {
+		if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".log") {
+			logFiles = append(logFiles, entry)
+		}
+	}
+	if len(logFiles) == 0 {
+		return 0, nil
+	}
+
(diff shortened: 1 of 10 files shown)


jandubois:
Waiting for #242 to be merged. The I will rebase on top and address any Windows-specific issues that might arise.

jandubois:
Latest AI report: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260325-232213-pr-234.html

I don't want to address any of the remaining findings (and the verdict is a rare "merge as-is" anyways).

pkg/instance/logs.go
```diff
@@ -0,0 +1,84 @@
...
+	entries, err := os.ReadDir(srcDir)
+	if err != nil {
+		return 0, fmt.Errorf("read instance directory %s: %w", srcDir, err)
+	}
+
+	// Filter to .log files before creating the destination directory,
+	// so we don't leave empty directories when no logs exist.
+	var logFiles []os.DirEntry
+	for _, entry := range entries {
+		if !entry.IsDir() && strings.HasSuffix(entry.Name(), ".log") {
```
mook-as: … Oh I guess we're not using `entry.Type().IsRegular()` because we _do_ want to copy over symlinks. It might be useful to note that in one of the comments.

pkg/instance/logs.go
```diff
@@ -0,0 +1,84 @@
...
+	destDir, err := nextAvailableDir(logDir, instanceName)
+	if err != nil {
+		return 0, err
+	}
+
+	count := 0
+	for _, entry := range logFiles {
+		src := filepath.Join(srcDir, entry.Name())
+		dst := filepath.Join(destDir, entry.Name())
+		if err := os.Rename(src, dst); err != nil {
```
mook-as, line 57: … Or not? This seems to rename the symlink (instead of dereferencing it), so maybe I'm wrong and it _should_ be a `IsRegular` check above? Again, noting symlink behaviour in a comment would be useful.

pkg/instance/logs.go
```diff
@@ -0,0 +1,84 @@
...
+			if count == 0 {
+				os.Remove(destDir)
+			}
+			return count, fmt.Errorf("preserve %s: %w", entry.Name(), err)
```
mook-as: This returns an error on the first failure, but only delete the output directory if it also happens on the first file. So if we had `a.log`, `b.log`; failure on `a.log` results in a missing directory, and failure on `b.log` results in a directory with just `a.log`. That seems inconsistent; maybe we should instead iterate through all the files, collect all the errors, and only unlink the directory if _none_ of the renames succeeded?

pkg/instance/logs_test.go
```diff
@@ -0,0 +1,105 @@
...
+	assert.NilError(t, err)
+	assert.Equal(t, filepath.Base(dir3), "instance.3")
+}
+
+func TestNextAvailableDir_ParentNotExist(t *testing.T) {
+	_, err := nextAvailableDir("/nonexistent/path", "instance")
+	assert.Assert(t, err != nil, "expected error for nonexistent parent")
+}
+
+func TestPreserveLogs_MovesLogFiles(t *testing.T) {
```
mook-as, line 48: There should be a test with one of the log files being a symlink, I think.

pkg/instance/logs_test.go
```diff
@@ -0,0 +1,105 @@
...
+	assert.NilError(t, err)
+	assert.Equal(t, filepath.Base(dir2), "instance.2")
+
+	dir3, err := nextAvailableDir(parent, "instance")
+	assert.NilError(t, err)
+	assert.Equal(t, filepath.Base(dir3), "instance.3")
+}
+
+func TestNextAvailableDir_ParentNotExist(t *testing.T) {
+	_, err := nextAvailableDir("/nonexistent/path", "instance")
```
mook-as: For sanity, I think the nonexistent path should be `filepath.Join(t.TempDir(), "nonexistent")` (in case some idiot actually creates that path… I'm probably that idiot.)

Same with `TestPreserveLogs_SrcDirNotExist`.

pkg/instance/logs_test.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func TestPreserveLogs_MovesLogFiles(t *testing.T) {
+	origLogDir := LogDir
+	logDir := t.TempDir()
+	LogDir = func() string { return logDir }
+	defer func() { LogDir = origLogDir }()
+
+	srcDir := t.TempDir()
+	assert.NilError(t, os.WriteFile(filepath.Join(srcDir, "serial.log"), []byte("serial output"), 0o644))
```
mook-as, line 55: I wonder if using a [`fstest.MapFS`](https://pkg.go.dev/testing/fstest@go1.26.1#MapFS) would be useful here. Probably not…

pkg/service/cmd/service.go
```diff
@@ -420,12 +421,50 @@ func Delete() error {
...
+			continue
+		}
+		instDir := filepath.Join(instance.LimaHome(), entry.Name())
+		count, err := instance.PreserveLogs(instDir, entry.Name())
+		if err != nil {
+			logrus.WithError(err).WithField("instance", entry.Name()).Warn("Failed to preserve instance logs")
+			continue
+		}
+		if count > 0 {
+			logrus.WithFields(logrus.Fields{"instance": entry.Name(), "count": count}).Info("Preserved instance logs")
```
mook-as: Should this be at higher verbosity, since that's normally not relevant?

Looking at that made me realize we're using `logrus` instead of `klog` here; as a future item, I think we should just use `klog` everywhere (even in the command line bits).

mook-as (dismissed):
I'd like to see more handling of symlinks (though I don't think we actually end up having any here), and there's some minor improvements possible, but generally I wouldn't complain too hard if this gets merged as-is.

pkg/instance/logs.go
jandubois: I changed to `IsRegular()` because we don't support symlinks, and don't expect symlinks.

pkg/instance/logs.go
jandubois: Added a comment. There should be no symlinks.

mook-as (approved):
Thanks; explicitly not supporting symlinks is just fine.

PR #236, opened 2026-03-13
Implement App controller lifecycle for LimaVM

Nino-K:
Introduce the core reconciliation loop for the App controller singleton. The controller creates and manages the LimaVM instance from the temporary embedded template, propagates `spec.running`, mirrors VM status conditions onto the App resource, and watches the owned LimaVM for status updates.

Deletion is handled by a finalizer to ensure owned resources are cleaned up. The input template ConfigMap is created only for bootstrapping and then removed after the LimaVM webhook uses the managed template.

Addresses: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/227

### DIFF
--- a/.github/actions/spelling/expect/rd.txt
+++ b/.github/actions/spelling/expect/rd.txt
@@ -28,5 +28,6 @@ guestagent
 limactl
 limavm
 pidfile
+rosetta
 syncer
 workdir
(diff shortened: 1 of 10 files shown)


jandubois:
Latest AI Deep Review: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260324-190453-pr-236.html

jandubois (approved):
Thanks, LGTM

Final AI review at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260324-145835-pr-236.html still shows some minor issues. Most can be ignored; a few can be addressed in later updates.

PR #242, opened 2026-03-19
Run Lima BATS tests on Windows via MSYS2

jandubois:
### Summary

- WSL2's `binfmt_misc` interop (which transparently executes `.exe` files from Linux) breaks mid-run on GitHub Actions runners, causing flaky test failures that are impossible to fix from our side. MSYS2 avoids this entirely because `rdd.exe` runs natively without interop translation. WSL2 is still installed for Lima (which needs it as a VM backend), but BATS itself no longer depends on a WSL distro.
- Fix Windows process signaling so stopping or restarting a VM no longer kills the RDD service (hostagents now run in their own process group)
- Fix `stopInstanceForcibly` to terminate the WSL2 distro and clean up stale PID/socket files; always terminate before unregistering to prevent `wslservice.exe` deadlocks
- Add Windows-specific Lima VM template (`finch` rootfs) and SSH key pre-generation to avoid MSYS2 path conversion issues
- Switch CI from WSL2-based BATS to MSYS2-based BATS, eliminating the dependency on a WSL distro for the test runner itself
- Honor `RDD_LOG_LEVEL` in externally spawned controllers so debug logging propagates correctly

### Key changes

Process isolation (`process_windows.go`, `limavm_lifecycle.go`): Each hostagent subprocess gets `CREATE_NEW_PROCESS_GROUP`. New `process.Interrupt()` sends `CTRL_BREAK_EVENT` targeted at the hostagent's group only. `stopInstanceForcibly` uses `TerminateProcess` instead of console-wide break events.

WSL2 deadlock prevention (`limavm_lifecycle.go`): `stopInstanceForcibly` now runs unconditionally before `wsl --unregister`, even for stopped instances. A "stopped" WSL2 distro can retain kernel-level state that deadlocks `wslservice.exe`, bricking all subsequent `wsl.exe` calls.

BATS hardening (`limavm-running.bats`): The cleanup test wraps `rdd ctl wait --for=delete` with an outer timeout and kills `wslservice.exe` if the wait hangs, recovering from WSL deadlocks instead of blocking the entire suite.

CI workflow (`bats.yaml`): Windows BATS tests run under MSYS2 (UCRT64) instead of inside a WSL2 distro. The Lima controller tests use the `finch` rootfs on Windows and the `opensuse` image on macOS.

### DIFF
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -6,5 +6,6 @@ routable #https://en.wiktionary.org/wiki/routable#English
 standalone #https://en.wiktionary.org/wiki/standalone#English
 startable #https://en.wiktionary.org/wiki/startable
 unmarshal #https://en.wiktionary.org/wiki/unmarshal#English
+wslservice
 versioned #https://en.wiktionary.org/wiki/versioned#English
 whitespaces #https://en.wiktionary.org/wiki/whitespaces#English
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,17 +1,9 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
-keypair
-msystem
-nohup
-taskkill
-ucrt
 passthrough #https://en.wiktionary.org/wiki/passthrough
 passthroughs #https://en.wiktionary.org/wiki/passthrough
-pgid
-rootfs #https://en.wiktionary.org/wiki/rootfs
 routable #https://en.wiktionary.org/wiki/routable#English
 standalone #https://en.wiktionary.org/wiki/standalone#English
 startable #https://en.wiktionary.org/wiki/startable
 unmarshal #https://en.wiktionary.org/wiki/unmarshal#English
 versioned #https://en.wiktionary.org/wiki/versioned#English
 whitespaces #https://en.wiktionary.org/wiki/whitespaces#English
-wslservice
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,4 +1,5 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
+keypair
 nohup
 taskkill
 passthrough #https://en.wiktionary.org/wiki/passthrough
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,6 +1,7 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
 passthrough #https://en.wiktionary.org/wiki/passthrough
 passthroughs #https://en.wiktionary.org/wiki/passthrough
+pgid
 routable #https://en.wiktionary.org/wiki/routable#English
 standalone #https://en.wiktionary.org/wiki/standalone#English
 startable #https://en.wiktionary.org/wiki/startable
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,4 +1,6 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
+nohup
+taskkill
 passthrough #https://en.wiktionary.org/wiki/passthrough
 passthroughs #https://en.wiktionary.org/wiki/passthrough
 pgid
@@ -6,6 +8,6 @@ routable #https://en.wiktionary.org/wiki/routable#English
 standalone #https://en.wiktionary.org/wiki/standalone#English
 startable #https://en.wiktionary.org/wiki/startable
 unmarshal #https://en.wiktionary.org/wiki/unmarshal#English
-wslservice
 versioned #https://en.wiktionary.org/wiki/versioned#English
 whitespaces #https://en.wiktionary.org/wiki/whitespaces#English
+wslservice
--- a/.github/actions/spelling/allow.txt
+++ b/.github/actions/spelling/allow.txt
@@ -1,10 +1,13 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
 keypair
+msystem
 nohup
 taskkill
+ucrt
 passthrough #https://en.wiktionary.org/wiki/passthrough
 passthroughs #https://en.wiktionary.org/wiki/passthrough
 pgid
+rootfs #https://en.wiktionary.org/wiki/rootfs
 routable #https://en.wiktionary.org/wiki/routable#English
 standalone #https://en.wiktionary.org/wiki/standalone#English
 startable #https://en.wiktionary.org/wiki/startable
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -108,7 +108,7 @@ jobs:
         # openssh is required because Lima constructs SSH paths using cygpath
         # (/c/Users/...) which only MSYS2's ssh understands; Windows OpenSSH
         # needs native paths and fails with "No such file or directory".
-        C:\msys64\usr\bin\pacman.exe --noconfirm -S gzip jq make openbsd-netcat openssh
+        C:\msys64\usr\bin\pacman.exe --noconfirm --sync gzip jq make openbsd-netcat openssh
 
         # Set up MSYS2 wrapper script
         New-Item -ItemType Directory -Name bin -Force
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -57,14 +57,14 @@ jobs:
         if test $(cat /proc/sys/net/ipv4/ip_unprivileged_port_start) -gt 80; then
           sudo sh -c 'echo 80 > /proc/sys/net/ipv4/ip_unprivileged_port_start'
         fi
-        
+
         # Install QEMU for Lima VM operations
         sudo apt-get update
         sudo apt-get install --yes qemu-system-x86 qemu-utils
 
-        # Set up run-in-wsl helper script (as a symlink to bash)
+        # Set up run-in-shell helper script (as a symlink to bash)
         mkdir -p bin
-        ln -s $(which bash) bin/run-in-wsl
+        ln -s $(which bash) bin/run-in-shell
         readlink -f bin >> "$GITHUB_PATH"
 
         # Install a newer version of jq
@@ -77,9 +77,9 @@ jobs:
       run: |
         brew install bash coreutils make
 
-        # Set up run-in-wsl helper script (as a symlink to bash)
+        # Set up run-in-shell helper script (as a symlink to bash)
         mkdir -p bin
-        ln -s $(brew --prefix bash)/bin/bash bin/run-in-wsl
+        ln -s $(brew --prefix bash)/bin/bash bin/run-in-shell
         readlink -f bin >> "$GITHUB_PATH"
 
         # Use GNU make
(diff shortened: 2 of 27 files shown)


.github/actions/spelling/allow.txt
```diff
@@ -1,9 +1,17 @@
 cooldown #https://en.wiktionary.org/wiki/cooldown#English
+keypair
```
mook-as: I believe non-English words should go in `expect` instead of `allow`.

.github/workflows/bats.yaml
```diff
@@ -100,43 +100,36 @@ jobs:
...
-        wsl --distribution openSUSE --exec sed -i '/^command =/d' /etc/wsl-distribution.conf
 
-        # Make sure WSL interop is enabled
-        wsl --distribution openSUSE --exec sh -c 'echo ":WSLInterop-custom:M::MZ::/init:PF" | tee /etc/binfmt.d/zzz-WSLInterop-custom.conf' # spellchecker:ignore
-        wsl --shutdown
+        # Install MSYS2 packages for BATS.
+        # openssh is required because Lima constructs SSH paths using cygpath
+        # (/c/Users/...) which only MSYS2's ssh understands; Windows OpenSSH
+        # needs native paths and fails with "No such file or directory".
+        C:\msys64\usr\bin\pacman.exe --noconfirm -S gzip jq make openbsd-netcat openssh
```
mook-as:
```suggestion
        C:\msys64\usr\bin\pacman.exe --noconfirm --sync gzip jq make openbsd-netcat openssh
```

.github/workflows/bats.yaml
```diff
@@ -100,43 +100,36 @@ jobs:
...
         wsl --install --no-launch --version 2 --name openSUSE openSUSE-Leap-16.0
-        # Disable first-boot script
-        wsl --distribution openSUSE --exec sed -i '/^command =/d' /etc/wsl-distribution.conf
 
-        # Make sure WSL interop is enabled
-        wsl --distribution openSUSE --exec sh -c 'echo ":WSLInterop-custom:M::MZ::/init:PF" | tee /etc/binfmt.d/zzz-WSLInterop-custom.conf' # spellchecker:ignore
-        wsl --shutdown
+        # Install MSYS2 packages for BATS.
+        # openssh is required because Lima constructs SSH paths using cygpath
+        # (/c/Users/...) which only MSYS2's ssh understands; Windows OpenSSH
```
mook-as, line 109: Does this mean we'll need to always ship cygwin/msys ssh with RDD? Or is this something that we can fix down the road?

pkg/util/process/process_windows.go
```diff
@@ -33,10 +49,39 @@ func Kill(pid int) error {
...
+//
+// Platform asymmetry: if the target process is already dead, taskkill /T
+// returns exit code 128 (treated as success), but surviving children (e.g.,
+// SSH port forwarders) are not killed because taskkill cannot traverse the
+// tree from a dead parent. On Unix, kill(-pgid) still reaches all group
+// members. This is acceptable: orphaned port forwarders cannot rebind their
+// ports and are harmless. Windows Job Objects would fix this if needed.
+//
+// Returns nil if the process no longer exists (taskkill exit code 128).
+func KillTree(ctx context.Context, pid int) error {
```
mook-as, line 82: This sounds like something that [jobs](https://learn.microsoft.com/en-us/windows/win32/procthread/job-objects) might be relevant for.

However, I could not spot any callers to double check if that is the case.

mook-as (commented):
I think it would be useful copy the reasoning about the move to msys2 in the commit message to the PR summary, so it would be easier to understand in the future (but I do appreciate that it's in the commit message).

Does this mean we currently support BATS on Windows in _both_ WSL2 and MSYS2?

I have not reviewed the changes in detail.

.github/actions/spelling/allow.txt
jandubois: Ok, will move all the new terms to the `expect/*` files. Although technically `keypair` is an English word (https://en.wiktionary.org/wiki/keypair#English), but will move that one too.

.github/workflows/bats.yaml
jandubois: We will use `wsl.exe` instead of `ssh.exe` for shell access.

The cygwin/msys dependency is currently hard-coded in Lima:

```go
func identityFileEntry(ctx context.Context, privateKeyPath string) (string, error) {
	if runtime.GOOS == "windows" {
		privateKeyPath, err := ioutilx.WindowsSubsystemPath(ctx, privateKeyPath)
		if err != nil {
			return "", err
		}
		return fmt.Sprintf(`IdentityFile='%s'`, privateKeyPath), nil
	}
	return fmt.Sprintf(`IdentityFile="%s"`, privateKeyPath), nil
}
```

and

```go
func WindowsSubsystemPath(ctx context.Context, orig string) (string, error) {
	out, err := exec.CommandContext(ctx, "cygpath", filepath.ToSlash(orig)).CombinedOutput()
	if err != nil {
		logrus.WithError(err).Errorf("failed to convert path to mingw, maybe not using Git ssh?")
		return "", err
	}
	return strings.TrimSpace(string(out)), nil
}
```

I think the problem is that Windows OpenSSH does not support the ControlPath logic that Lima uses. That is just a performance optimization, but this will be a larger change than just dealing with the path to `ssh.exe`.

pkg/util/process/process_windows.go
jandubois: Yes, jobs would be better. Line 75 already mentions that. I think we don't need it yet, but I could be wrong. It would definitely be cleaner.

pkg/service/cmd/service.go
```diff
@@ -338,8 +338,21 @@ func StopWithWait(wait bool) error {
 	_ = cleanupDiscoveryConfigMap() // Clean up discovery configmap to prevent stale controller info
 
 	pid := PID()
```
mook-as, line 340: Re: the AI review about PID reuse: Ideally that would be fixed via using pidfd (equivalently, handles on Windows). However, I think we can punt that to a follow-up PR.

jandubois: Yes, I'm coming around to the conclusion that using a Job object for each hostagent is the right thing to do on Windows. That should get rid of the reliance on PIDs completely and also guarantee that all hostagents and their children will be terminated when the RDD service exits.

This is a bigger change that should go into a new PR.

jandubois:
Latest AI review is at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260324-181620-pr-242.html

I think I want to address any further issues in follow-up PRs and get this merged instead, so undrafting this PR.

bats/helpers/vm_template.bash
```diff
@@ -0,0 +1,70 @@
+# Lima VM template selection for BATS tests.
+#
+# On Windows (WSL2), set RDD_WSL_DISTRO to choose between distros:
+#   - "finch"    (default): Finch rootfs (Fedora-based, stable for testing)
```
mook-as, line 4: Presumably, we'll move off finch in the future. But that's a future thing, nothing to do here.

bats/helpers/vm_template.bash
```diff
@@ -0,0 +1,70 @@
...
+#   - "opensuse":           rancher-desktop-opensuse (the target distro for RD2)
+#
+# Usage in test files:
+#   VM_TEMPLATE=$(vm_template)
+# Supports RDD_VM_TYPE on Unix (expands to vmType: <value> when set).
```
mook-as:
```suggestion
# Supports RDD_VM_TYPE on Unix (expands to `vmType: <value>` when set).
```
I realize that doesn't do anything, it just makes reading easier.

bats/tests/33-lima-controllers/limavm-instance.bats
```diff
@@ -36,10 +24,13 @@ local_teardown_file() {
...
     done
-}
-
-local_setup() {
-    skip_on_windows
+    # Clean up any WSL2 distros created by tests.
+    # Use a timeout because wsl.exe --unregister can hang.
+    if is_windows; then
+        for vm in "${VM_NAME}" "invalid-vm"; do
+            timeout 30 bash -c "MSYS_NO_PATHCONV=1 wsl.exe --unregister 'lima-${vm}'" 2>/dev/null || true
```
mook-as, line 31: I don't think I understand why we need the explicit `bash` invocation (as opposed to just `env MSYS_NO_PATHCONV=1 wsl.exe`), but that's probably fine?

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -505,11 +517,21 @@ env:
 
 @test "cleanup LimaVM running test" {
     rdd ctl delete limavm "${VM_NAME}" --namespace "${NAMESPACE}"
-    rdd ctl wait --for=delete "limavm/${VM_NAME}" --namespace "${NAMESPACE}" --timeout=60s
+    # Wrap with an outer timeout because rdd ctl wait --timeout may not be
+    # enforced when the reconciler is stalled. On WSL2, a failed
+    # wsl.exe --unregister can deadlock wslservice.exe, blocking all
+    # subsequent wsl.exe calls (including store.Inspect in the reconciler).
```
mook-as, line 523: Should that be filed as an issue (and ignored for now)? The part where `rdd ctl` can get stalled, that is; `wsl.exe` hanging isn't our issue.

pkg/controllers/app/app/controller.go
```diff
@@ -15,10 +16,26 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 )
 
-// This is temporary and will be removed once the app controller is fully implemented.
+// Embedded Lima template split into platform-specific images and shared
```
mook-as, line 19: Future improvement: OS-specific files (`controller_windows.go`, etc.) so we don't have to think about runtime effects.

pkg/controllers/app/app/lima-images-wsl2.yaml
```diff
@@ -0,0 +1,6 @@
+vmType: wsl2
+images:
+- location: https://deps.runfinch.com/common/x86-64/finch-rootfs-production-amd64-1771357941.tar.gz
```
mook-as, line 3: Presumably we'll use `rancher-desktop-opensuse` in the future; but it doesn't matter right now.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -448,6 +449,11 @@ func (r *LimaVMReconciler) updateCondition(ctx context.Context, limaVM *v1alpha1
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *LimaVMReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// No parent context exists yet; the manager hasn't started.
+	if err := ensureSSHKeys(context.Background()); err != nil {
```
mook-as, line 453: I wonder if we can thread a proper context down from the callers.  I guess `context.TODO()` is more correct here, because making it use something better should be a goal, but it's not clear which?

```suggestion
	if err := ensureSSHKeys(context.TODO()); err != nil {
```

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -42,15 +45,41 @@ func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.
...
+			// the PID may already be recycled. Retrying with a stale PID is
+			// worse than letting Delete proceed without a kill.
+			existingInst.DriverPID = 0
+			existingInst.HostAgentPID = 0
 		}
 		logger.Info("Deleting Lima instance", "instance", limaVM.Name)
-		if err := limainstance.Delete(ctx, existingInst, true); err != nil {
+		// Use a timeout because Lima's WSL2 driver calls wsl.exe --unregister
+		// which can hang indefinitely if the WSL subsystem is degraded.
+		deleteCtx, deleteCancel := context.WithTimeout(ctx, 60*time.Second)
```
mook-as:
```suggestion
		deleteCtx, deleteCancel := context.WithTimeout(ctx, time.Minute)
```

mook-as (dismissed):
LGTM, with follow-ups. I just realized that darwin doesn't really have pidfd-equvialents, which makes things harder to deal with. (But hey, `os.Process` _does_ use pidfd/handle where available, so we can just hold those instead…)

bats/helpers/vm_template.bash
jandubois: Yes, the BATS test for the LimaVM controller is already configurable between `finch` and `opensuse`. I was just worried that our network namespace could introduce more issues in this PR that should be solved separately. This may turn out to be superstition.

bats/tests/33-lima-controllers/limavm-instance.bats
jandubois: We don't really need it, but it allows us to use quoting for the distro name. `timeout` does not really allow you to quote arguments containing whitespace.

I would leave it; we often quote variable expansions, even when we know the variable has no whitespace.

pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: I don't know how this would work. `SetupWithManager` is a `controller-runtime` API that doesn't take a context. We could one define locally in there, but the end effect would be pretty similar.

We can figure this out later; from my point of view `TODO` and `Background` are very similar: I don't have a context, but I need one, so I just create one out of thin air.

mook-as (dismissed):

mook-as (approved):

PR #243, opened 2026-03-21
Encode owner Kind in owned finalizer, document webhook requirement

jandubois:
- Encode the owner's `Kind` in the owned finalizer name (`rdd.rancherdesktop.io/owned-by-App` instead of `rdd.rancherdesktop.io/owned`)
- Document that any resource type carrying an owned finalizer must have a validating webhook that rejects `DELETE` while the finalizer is present
- Add generic `OwnedDeletionGuard[T]` webhook in the base package
- Refactor `ConfigMapValidator` to embed `OwnedDeletionGuard` instead of hand-rolling `ValidateDelete`

### Motivation

The owned finalizer and a validating webhook work together as a single access-control gate: the webhook rejects `DELETE` requests while the finalizer is present, and the owner removes the finalizer when it wants to allow deletion. Without the webhook, a user who deletes an owned resource gets a stuck `Terminating` object with no explanation.

Encoding the owner `Kind` in the finalizer name (`owned-by-App`, `owned-by-LimaVM`) makes the finalizer self-documenting and lets the generic `OwnedDeletionGuard` construct clear rejection messages without per-resource configuration.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-lifecycle.bats
+++ b/bats/tests/33-lima-controllers/limavm-lifecycle.bats
@@ -74,7 +74,7 @@ EOF
 
 @test "verify template ConfigMap has owned finalizer" {
     run -0 rdd ctl get configmap "${TEMPLATE_NAME}" --namespace "${NAMESPACE}" -o jsonpath='{.metadata.finalizers}'
-    assert_output --partial "rdd.rancherdesktop.io/owned"
+    assert_output --partial "rdd.rancherdesktop.io/owned-by-LimaVM"
 }
 
 @test "verify template ConfigMap has owner reference" {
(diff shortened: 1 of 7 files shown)


jandubois:
Automated code review at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260321-005431-pr-243.html

Only 2 seem worth addressing:

  1. Added BATS test for label-removal rejection (limavm-lifecycle.bats)
  2. Fixed doc reference to ConfigMapValidator embedding (controllers.md)
  3. Skipped — empty-string guard (don't want to panic)
  4. Skipped — oldConfigMap vs newConfigMap (identical behavior in practice)
  5. Skipped — allocation pre-check (negligible impact)

Nino-K (approved):

PR #245, opened 2026-03-24
BATS: Use *.test domains for testing

mook-as:
The `.test` TLD is reserved in RFC 2606 to never have any actual host names.  Use that instead of a plain `foo` to avoid issues with networks where attempting to resolve domains that don't exist can result in issues.


jandubois (approved):
Thanks, LGTM

PR #248, opened 2026-03-25
App controller Stop and Start LimaVM

Nino-K:
Allows starting and stopping the LimaVM from the App Controller.

It is related to the following: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/228

### DIFF
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -138,12 +138,12 @@ local_setup() {
 
 @test "verify App conditions preserve LimaVM LastTransitionTime" {
     run -0 rdd ctl get limavm "${VM_NAME}" --namespace "${RDD_NAMESPACE}" -o json
-    local limavm_ts
-    limavm_ts=$(jq -r '.status.conditions[] | select(.type == "Created") | .lastTransitionTime' <<<"${output}")
+    run -0 jq_output '.status.conditions[] | select(.type == "Created") | .lastTransitionTime'
+    local limavm_ts="${output}"
 
     run -0 rdd ctl get app "${APP_NAME}" -o json
-    local app_ts
-    app_ts=$(jq -r '.status.conditions[] | select(.type == "Created") | .lastTransitionTime' <<<"${output}")
+    run -0 jq_output '.status.conditions[] | select(.type == "Created") | .lastTransitionTime'
+    local app_ts="${output}"
 
     [[ "${limavm_ts}" == "${app_ts}" ]]
 }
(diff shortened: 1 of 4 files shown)


bats/tests/32-app-controllers/app.bats
```diff
@@ -150,12 +154,62 @@ local_setup_file() {
...
+        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=120s
+}
+
+@test "wait for App Running condition to become True" {
+    rdd ctl wait --for=condition=Running app/"${APP_NAME}" --timeout=30s
+}
+
+@test "verify App mirrors LimaVM Running condition as True" {
+    run -0 rdd ctl get app "${APP_NAME}" -o json
+    run -0 jq -r '.status.conditions[] | select(.type == "Running") | .status' <<<"${output}"
```
jandubois: Use `jq_output` helper:

```suggestion
    run -0 jq_output '.status.conditions[] | select(.type == "Running") | .status'
```

bats/tests/32-app-controllers/app.bats
```diff
@@ -150,12 +154,62 @@ local_setup_file() {
...
+    assert_output "True"
+}
+
+@test "verify App Running condition preserves LimaVM LastTransitionTime" {
+    run -0 rdd ctl get limavm "${VM_NAME}" --namespace "${RDD_NAMESPACE}" -o json
+    local limavm_ts
+    limavm_ts=$(jq -r '.status.conditions[] | select(.type == "Running") | .lastTransitionTime' <<<"${output}")
```
jandubois: This pattern occurs multiple times:

```suggestion
    run -0 jq_output '.status.conditions[] | select(.type == "Running") | .lastTransitionTime'
    local limavm_ts=$output
```

bats/tests/32-app-controllers/app.bats
```diff
@@ -150,12 +154,62 @@ local_setup_file() {
...
+@test "verify App Running condition observedGeneration matches LimaVM" {
+    run -0 rdd ctl get limavm "${VM_NAME}" --namespace "${RDD_NAMESPACE}" -o json
+    local limavm_gen
+    limavm_gen=$(jq -r '.status.conditions[] | select(.type == "Running") | .observedGeneration' <<<"${output}")
+
+    run -0 rdd ctl get app "${APP_NAME}" -o json
+    local app_gen
+    app_gen=$(jq -r '.status.conditions[] | select(.type == "Running") | .observedGeneration' <<<"${output}")
+
+    [[ "${limavm_gen}" == "${app_gen}" ]]
```
jandubois:
```suggestion
    run -0 jq_output '.status.conditions[] | select(.type == "Running") | .observedGeneration'
    assert_output "${limavm_gen}"
```

jandubois (commented):
Not a full review, just some feedback on BATS patterns.

jandubois:
Full AI review at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260327-123931-pr-248.html

I think only the suggestions I include below require changes.

I know you said using `jq_output` did not work for you, but I don't see why it wouldn't. Please show what you tried, so we can fix it!

---

## Suggestions

1. **Misleading commit message on `aadeeb4`** — `pkg/controllers/app/app/controllers/app_controller.go` [Claude Opus 4.6] (suggestion, regression)

```diff
-    log.Error(cmErr, "Failed to fetch input ConfigMap during cleanup")
+    return ctrl.Result{}, fmt.Errorf("failed to fetch input ConfigMap during cleanup: %w", cmErr)
```

The commit title says "Fix ObservedGeneration in App Controller," but the diff only converts two `log.Error` calls to `return error`. `ObservedGeneration` was already `cond.ObservedGeneration` at the merge base and this commit did not change it. The message body compounds the mismatch: "The LimaVM's `ObservedGeneration` is now correctly propagated."

Fix: amend the commit message to describe the actual change, e.g., "Return errors from ConfigMap Get failures instead of logging and continuing."

2. **Doc says "in the app-namespace" for a cluster-scoped resource** — `docs/design/api_app.md:11` [Codex GPT 5.4, Gemini 3.1 Pro] (suggestion, regression)

Line 9 clarifies that `App` is cluster-scoped, but line 11 says the CLI creates it "in the app-namespace." Cluster-scoped resources do not belong to a namespace. The configuration setting controls `spec.namespace`, which dictates where child resources are created.

Fix:
```diff
-create the `App` object in the "app-namespace", which is a configuration setting
+create the cluster-scoped `App` object, setting its `spec.namespace` to the configured "app-namespace"
```

3. **BATS tests use `$(jq_output ...)` instead of `run -0 jq_output`** — `bats/tests/32-app-controllers/app.bats:175,179,187,189` [Claude Opus 4.6, Gemini 3.1 Pro] (suggestion, regression)

Four places extract values via `$(jq_output ...)` command substitution. When `jq_output` fails inside `$(...)`, BATS produces an opaque error rather than showing the actual output. This echoes Jan's unresolved review feedback.

Fix: use `run -0 jq_output` to capture each value, matching the pattern Jan suggested:
```diff
-    limavm_ts=$(jq_output '.status.conditions[] | select(.type == "Running") | .lastTransitionTime')
+    run -0 jq_output '.status.conditions[] | select(.type == "Running") | .lastTransitionTime'
+    local limavm_ts=$output
```

Apply the same pattern to lines 179, 187, and 189. For the `observedGeneration` test, replace the `[[ ]]` comparison with `assert_output` for better failure diagnostics.

---

## Documentation Assessment

The `api_app.md` updates document mirrored condition semantics, the condition table, and example YAML. Two gaps:

- Line 11's "in the app-namespace" wording conflicts with the cluster-scoped clarification at line 9 (see Suggestion 2).
- `api_lima.md` does not document the `Running=Unknown, Reason=Reconciling` state that `api_app.md:99` now references as mirrored from LimaVM.

jandubois (approved):
Thanks, LGTM

PR #253, opened 2026-03-30
BATS: Use bash 4 syntax for upper case strings

mook-as:
See https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/181


jandubois (approved):
Thanks, LGTM

PR #254, opened 2026-03-30
CI: Lint: Provide token to github CLI

mook-as:
This ensures that the check-spelling step can proceed.  We only provide `contents: read` because we don't care about `pull_request_target` events.


jandubois (approved):
Thanks, LGTM

PR #266, opened 2026-03-31
Enforce godoc comments on exported symbols

### LINKED ISSUE #249: Enforce GoDoc comments on exported symbols
The two revive rules that would enforce this are explicitly disabled:

- exported: this is the rule that requires doc comments on exported symbols
- package-comments: this requires package-level doc comments

Additionally, staticcheck's ST1000 (package doc comments) and ST1022 (comment format for exported error vars) are suppressed as well.

jandubois:
Add missing doc comments to all exported types, functions, variables, and constants. Enable the revive `exported` rule and remove the staticcheck ST1022 suppression (no violations exist).

Fix stuttering type names flagged by the exported rule:
- container.ContainerReconciler → container.Reconciler
- container.ContainerImmutableValidator → container.ImmutableValidator
- containernamespace.ContainerNamespaceDeleteValidator → containernamespace.DeleteValidator
- tail.TailFile → tail.File

Closes #249

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -101,7 +101,6 @@ linters:
       - name: error-strings
       - name: errorf
       - name: exported
-        disabled: true
       - name: increment-decrement
       - name: indent-error-flow
       - name: package-comments
@@ -124,7 +123,6 @@ linters:
       - -QF*
       - -ST1000
       - -ST1001  # duplicates revive.dot-imports
-      - -ST1022
     usetesting:
       os-temp-dir: true
   exclusions:
--- a/pkg/controllers/containers/container/container_reconciler.go
+++ b/pkg/controllers/containers/container/container_reconciler.go
@@ -21,8 +21,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 )
 
-// ContainerReconciler reconciles a Container object.
-type ContainerReconciler struct {
+type reconciler struct {
 	client.Client
 	Scheme   *runtime.Scheme
 	Recorder events.EventRecorder
@@ -33,7 +32,7 @@ type ContainerReconciler struct {
 // +kubebuilder:rbac:groups=containers.rancherdesktop.io,resources=containers/finalizers,verbs=update
 
 // Reconcile implements a container reconciliation loop.
-func (r *ContainerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := logf.FromContext(ctx)
 
 	var container containersv1alpha1.Container
@@ -79,7 +78,7 @@ func (r *ContainerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 }
 
 // setCondition sets or updates a condition in the container status.
-func (r *ContainerReconciler) setCondition(container *containersv1alpha1.Container, conditionType string, status metav1.ConditionStatus, reason, message string) {
+func (r *reconciler) setCondition(container *containersv1alpha1.Container, conditionType string, status metav1.ConditionStatus, reason, message string) {
 	changed := apimeta.SetStatusCondition(&container.Status.Conditions, metav1.Condition{
 		Type:    conditionType,
 		Status:  status,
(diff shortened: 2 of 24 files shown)


pkg/controllers/containers/container/container_reconciler.go
```diff
@@ -21,8 +21,8 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 )
 
-// ContainerReconciler reconciles a Container object.
-type ContainerReconciler struct {
+// Reconciler reconciles Container objects.
+type Reconciler struct {
```
mook-as: Does this need to be exported? This also applies to the other reconcilers / validators we touch in this PR. Don't bother with anything not already affected.

Also, I believe we can make a [doc link](https://go.dev/doc/comment#doclinks) to make it better:

```suggestion
// Reconciler reconciles [Container] objects.
type Reconciler struct {
```

pkg/service/cmd/service.go
```diff
@@ -156,10 +159,12 @@ func PID() int {
 	return pid
 }
 
+// Running reports whether the service process is alive.
 func Running() bool {
 	return PID() != PIDNotFound
 }
 
+// Create initializes a new control plane instance with the given arguments.
```
mook-as: Consider making it more English-like:

```suggestion
// Create a new control plane instance with the given arguments.
```

Same with `Start the service…` etc.

pkg/service/cmd/service.go
```diff
@@ -324,12 +331,15 @@ func Wait(ctx context.Context) error {
 	}
 }
 
+// WaitWithTimeout calls Wait with a 90-second deadline.
```
mook-as: Specific values (_90-second_, _60-second timeout_ below) seem too implementation-specific for the doc comment.

.golangci.yml
```diff
@@ -101,7 +101,6 @@ linters:
       - name: error-strings
       - name: errorf
       - name: exported
-        disabled: true
       - name: increment-decrement
       - name: indent-error-flow
       - name: package-comments
```
mook-as, line 106: Oh, I thought we wanted to enable this one too.

mook-as (dismissed):
Have comments, but since your changes is already a strict improvement I'm okay with dealing with them later.

… _and_ I just realized this is still draft; sorry about that. Feel free to amend all you like.

.golangci.yml
jandubois: I'm not sure, do we? I would do it in a second PR; generate with AI and then take a look if it is useful or not. Personally I'm more interested in the per-symbol comments.

pkg/controllers/containers/container/container_reconciler.go
```diff
@@ -21,8 +21,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 )
 
-// ContainerReconciler reconciles a Container object.
-type ContainerReconciler struct {
+type reconciler struct {
```
mook-as, line 24: It might still be useful to call it `containerReconciler`, but I guess it's not _that_ useful in `package container`…

pkg/controllers/lima/limavm/controller.go
```diff
@@ -288,10 +288,13 @@ type ConfigMapValidator struct {
 
 var _ ctrlwebhookadmission.Validator[*corev1.ConfigMap] = &ConfigMapValidator{}
 
+// ValidateCreate validates template data in newly created ConfigMaps.
```
mook-as, line 291: Possibly also add that it `implements [ctrlwebhookadmission.Validator]` (or whatever the correct thing is).

mook-as (approved):
LGTM. Comments are probably not worth addressing.

pkg/controllers/containers/container/container_reconciler.go
jandubois: Yeah, and it would trigger the "stutter" linter rule that symbol names should not repeat the package name. Which is why `tail.TailFile` is now `tail.File`.

pkg/controllers/lima/limavm/controller.go
jandubois: I don't know if this needs to be in the doc-string. We already "document" just above with:

```
var _ ctrlwebhookadmission.Validator[*corev1.ConfigMap] = &ConfigMapValidator{}
```

mook-as: We won't see that in the godoc view, nor hover tooltips in IDEs.

jandubois: Maybe we need a style guide for Godoc comments, to define how exactly they should be constructed?

PR #268, opened 2026-03-31
Enforce godoc comments on packages

### LINKED ISSUE #249: Enforce GoDoc comments on exported symbols

jandubois:
Enable the revive package-comments rule and the staticcheck ST1000 check. Add doc comments to all packages that lack them.

For multi-file packages, place the doc comment in a dedicated doc.go file so it is easy to find. Single-file packages keep the comment inline. Kubebuilder API packages keep the comment in groupversion_info.go (the kubebuilder convention), and vendored nxadmtail keeps it in its main file (different license headers).

Fix four files where missing blank lines between SPDX headers and the package declaration caused the linter to misinterpret the SPDX block as a malformed doc comment.

Remove duplicate doc comments in pkg/controllers/base (finalizer.go), pkg/apis/rdd/v1alpha1 (configmapreplicaset_types.go), and pkg/util/process (platform files).

Closes #249

### DIFF
--- a/cmd/app-controller/main.go
+++ b/cmd/app-controller/main.go
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
+// Command app-controller runs the app controller as a standalone process, for
+// development and testing outside the embedded control plane.
 package main
 
 import (
--- a/pkg/apis/rdd/v1alpha1/configmapreplicaset_types.go
+++ b/pkg/apis/rdd/v1alpha1/configmapreplicaset_types.go
@@ -2,8 +2,6 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
-// Package v1alpha1 contains the API definitions for the configmap replicaset controller.
-// This package defines custom resources that manage ConfigMaps without any pod dependencies.
 package v1alpha1
 
 import (
--- a/pkg/controllers/app/app/controller.go
+++ b/pkg/controllers/app/app/controller.go
@@ -2,6 +2,9 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
+// Package app registers the App controller. The App controller manages the
+// cluster-scoped App singleton that represents the Rancher Desktop application;
+// it creates and owns a LimaVM and mirrors its conditions back to App status.
 package app
 
 import (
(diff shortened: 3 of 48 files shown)


cmd/app-controller/main.go
```diff
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
+// Package main runs the app controller process.
```
mook-as: As an example, I would instead say here:

> Command app-controller runs the Rancher Desktop Daemon app controller as a standalone process, for development and testing purposes.

`cmd/…` should always have package comments that start with `Command foo …` I think.

pkg/apis/rdd/v1alpha1/configmapreplicaset_types.go
```diff
@@ -2,8 +2,6 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
-// Package v1alpha1 contains the API definitions for the configmap replicaset controller.
-// This package defines custom resources that manage ConfigMaps without any pod dependencies.
```
mook-as, line 6: I would update the doc comment (in `groupversion_info.go`) about what this group is for.

pkg/controllers/app/app/controller.go
```diff
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
+// Package app registers the App controller and its CRD with the controller
```
mook-as: … There should be something about what the app controller does (e.g. that it manages the app singleton).

pkg/controllers/app/demo/controllers/demo_controller.go
```diff
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-FileCopyrightText: The Rancher Desktop Authors
 
+// Package controllers implements the Demo reconciler.
```
mook-as: Maybe add a comment that this is for experimental purposes and will go away in the future.  Basically, _why_ this package exists.

mook-as (dismissed):
I agree they're too terse, but it's better than nothing, and it definitely helps us catch the issue with the license comments being incorrectly turned into doc comments. I'd say we keep it (unless you want to manually improve the comments now, and we can review again).

mook-as (approved):

PR #269, opened 2026-03-31
Add containerEngine and kubernetes fields to the App resource

Nino-K:
Add `spec.containerEngine.name` and `spec.kubernetes.{enabled,version}` to the App resource. The App controller injects these as Lima template params on creation and propagates changes to the LimaVM's template ConfigMap at runtime.

Related: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/259


docs/design/api_app.md
```diff
@@ -58,37 +58,37 @@ spec:
   containerEngine:
     name: moby
   kubernetes:
-    version: 1.30.0
+    enabled: false
+    version: 1.30.2
```
jandubois: Trivial, but let's make it `1.32.2` because we won't support any version before `1.32.0`.

Update additional references below as well.

docs/design/api_app.md
```diff
@@ -58,37 +58,37 @@ spec:
...
     message: Lima instance is running
     lastTransitionTime: "2024-01-01T00:00:05Z"
+    observedGeneration: 1
 ```
 
 - **spec.namespace**: The namespace where the owned `LimaVM` and its ConfigMaps are created. Defaults to `default`. **Immutable after creation** — changing it would orphan resources in the original namespace.
 
 - **spec.running**: Set to `true` to start the LimaVM, `false` to stop it. The App controller propagates this value to `LimaVM.spec.running` on every reconcile.
 
+- **spec.containerEngine.name**: The container engine to use inside the VM. Valid values: `moby` (Docker-compatible, default) and `containerd`. Propagated to the `CONTAINER_ENGINE` Lima template param. When changed on a running App, the App controller updates the LimaVM's template ConfigMap, which triggers the LimaVM controller's template-update path (a VM restart).
```
jandubois: The ConfigMap is **always** updated, even when the instance is not running.

```suggestion
- **spec.containerEngine.name**: The container engine to use inside the VM. Valid values: `moby` (Docker-compatible, default) and `containerd`. Propagated to the `CONTAINER_ENGINE` Lima template param.
```

pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -11,6 +10,26 @@ import (
...
+}
+
+// KubernetesSpec defines the desired Kubernetes configuration.
+type KubernetesSpec struct {
+	// enabled specifies whether Kubernetes should be enabled in the VM.
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled"`
+	// version is the Kubernetes version to use (e.g. "1.30.2").
+	// +optional
+	// +kubebuilder:default="1.30.2"
```
jandubois: I don't think we should have a default here. If anything, we would pick the latest stable version from `k3s-versions.json` as the default, but let's not worry about that right now.

Using `""` as the default is pointless because it matches the zero value, so we can leave it out.

pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -11,6 +10,26 @@ import (
...
+	// Valid values are "moby" (Docker-compatible) and "containerd".
+	// +kubebuilder:validation:Enum=moby;containerd
+	// +kubebuilder:default=moby
+	Name string `json:"name"`
+}
+
+// KubernetesSpec defines the desired Kubernetes configuration.
+type KubernetesSpec struct {
+	// enabled specifies whether Kubernetes should be enabled in the VM.
+	// +kubebuilder:default=false
```
jandubois: I don't think specifying the zero value as a default is useful; it is already the default behaviour anyways.

pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -25,6 +44,14 @@ type AppSpec struct {
 	// +kubebuilder:default="default"
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.namespace is immutable"
 	Namespace string `json:"namespace,omitempty"`
+	// containerEngine specifies the container engine configuration.
+	// +optional
+	// +kubebuilder:default={name:"moby"}
+	ContainerEngine ContainerEngineSpec `json:"containerEngine,omitempty"`
+	// kubernetes specifies the Kubernetes configuration.
+	// +optional
+	// +kubebuilder:default={enabled:false,version:"1.30.2"}
```
jandubois: The defaults are all zero values, so we leave this out too.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,6 +40,48 @@ type AppReconciler struct {
...
+
+	params, _ = tmpl["param"].(map[string]any)
+	if params == nil {
+		params = make(map[string]any)
+	}
+
+	return tmpl, params, nil
+}
+
+func applySpecToTemplate(templateData string, spec v1alpha1.AppSpec) (string, error) {
```
jandubois: I think this mechanism is wrong. The App controller should never parse the existing template; it just creates a new template from the embedded version and appends parameters. This should reuse the same code that is used to create the initial ConfigMap template; we should not have separate create/edit implementations.

jandubois (changes requested):
Latest AI review: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260401-120241-pr-269.html

It has some valid concerns, but my biggest feedback is that the App controller should never unmarshal the template from the ConfigMap; it should always create a new template and let the LimaVM controller deal with it.

Well, since it already has the old ConfigMap in memory, it could compare the old template to the new one and skip the update if they match.

As for the AI suggestion to process the `running` state first, I think it is a little more complex: if the desired state is `Stopped`, then the it should be applied first. But if the desired state is `Running`, then it should be applied last. This potentially saves one restart cycle in different situations.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -172,7 +178,6 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 		return ctrl.Result{}, fmt.Errorf("failed to fetch input ConfigMap: %w", err)
 	}
 
-	// Propagate spec.running from App into the LimaVM.
 	if limaVM.Spec.Running != app.Spec.Running {
```
jandubois: I suggest to duplicate the `limaVM.Spec.Running` update and only update here if the desired state is `false`:

```suggestion
	if limaVM.Spec.Running && !app.Spec.Running {
```

Because that means we stop the VM, update the ConfigMap, and the VM remains stopped.

But if the desired state is `true`, the setting it here will start the VM, and updating the settings will stop and restart it right after. So it is better to leave it stopped, update the settings, and the afterward check if the Running state doesn't match, and the desired state is `true`, and only then start it.

Saves one reboot cycle.

jandubois (changes requested):
Looks good, but I have one optimization suggestion that is easy to implement.

pkg/controllers/app/app/controllers/app_controller.go
Nino-K: changing the `limaVM.Spec.Running && !app.Spec.Running` removes the start case entirely, but I can also see your point. The latest change should do this.

jandubois (approved):
Thanks, LGTM

PR #270, opened 2026-03-31
Fix concurrent scheme access during controller startup

jandubois:
`RegisterWithManager` calls `AddToScheme` (a map write) while webhook setup goroutines call `client.Apply` → `Scheme.ObjectKinds` (a map read). Running both in the same loop causes a fatal "concurrent map read and map write" crash, observed on Windows CI.

Split the single loop into two: register all controllers first, then launch webhook goroutines after all scheme mutations are complete.

Sample crash: https://github.com/rancher-sandbox/rancher-desktop-daemon/actions/runs/23813525401/job/69406800691?pr=267


mook-as (approved):
Sounds plausible.

PR #272, opened 2026-04-02
mock controller: Add second container

mook-as:
This is useful for testing sorting etc.
    
Also mark the mock controller as depending on the test data in the makefile so it gets rebuilt when it is updated.

Additional changes from trying to get this to be pushable:
- Replace `set -eux` with `set -e -u -x` to satisfy the spell checker
- Don't try to check the guest agent binary for license tags


jandubois (approved):
Thanks, LGTM

PR #273, opened 2026-04-07
Add provisioning scripts to bundled Lima template

### LINKED ISSUE #263: Add provisioning scripts to bundled App template
Extract the provisioning scripts from TwoCows and include them in our bundled Lima template. All configurable settings should use the PARAM mechanism.

Generate the template dynamically by combining the static bundled template with a `param` section derived from the App properties.

Note that any `param` variable you set must be referenced in the template, even if just in a comment. Check the TwoCows template; it has comments with some variables that are not otherwise used by the scripts.

The TwoCows provisioning scripts may need to be adjusted to create the correct docker context, or the right kube context name.

---

When this issue is completely implemented, it should be possible enabled Kubernetes and deploy an app to the cluster. Or run a container with `docker`.

Nino-K:
Extract provisioning logic from TwoCows and include it in the bundled Lima template. The provision script enables the configured container engine (moby/containerd) and installs k3s when Kubernetes is enabled.

A script merges the k3s kubeconfig into the host's `~/.kube/config` under the "rancher-desktop" context.

All configurable values are exposed via the `PARAM` mechanism. `HOST_HOME` is added to the appended param block using Lima's `{{.Home}}` template variable.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/263

### DIFF
--- a/pkg/controllers/app/app/controllers/app_controller.go
+++ b/pkg/controllers/app/app/controllers/app_controller.go
@@ -44,7 +44,7 @@ type AppReconciler struct {
 
 func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) string {
 	return baseTemplate + fmt.Sprintf(
-		"\nparam:\n  CONTAINER_ENGINE: %s\n  KUBERNETES_ENABLED: %v\n  KUBERNETES_VERSION: %s\n",
+		"\nparam:\n  CONTAINER_ENGINE: %s\n  HOST_HOME: \"{{.Home}}\"\n  KUBERNETES_ENABLED: %v\n  KUBERNETES_VERSION: %s\n",
 		spec.ContainerEngine.Name,
 		spec.Kubernetes.Enabled,
 		spec.Kubernetes.Version,
--- a/pkg/controllers/app/app/lima-template.yaml
+++ b/pkg/controllers/app/app/lima-template.yaml
@@ -40,6 +40,10 @@ provision:
 
     systemctl stop rancher-desktop.target
 
+    # Write the selected container engine; docker.service and containerd.service
+    mkdir -p /etc/sysconfig
+    echo "CONTAINER_ENGINE=${PARAM_CONTAINER_ENGINE}" >/etc/sysconfig/rancher-desktop
+
     # Enable the requested container engine and disable the other.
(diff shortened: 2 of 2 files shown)


bats/tests/32-app-controllers/app.bats
```diff
@@ -127,6 +127,36 @@ local_setup_file() {
...
+        --namespace "${RDD_NAMESPACE}" -o jsonpath='{.data.template}'
+    assert_output --partial "KUBERNETES_ENABLED: false"
+}
+
+@test "verify template param HOST_HOME is present" {
+    run -0 rdd ctl get configmap "${VM_NAME}-template" \
+        --namespace "${RDD_NAMESPACE}" -o jsonpath='{.data.template}'
+    assert_output --partial "HOST_HOME:"
+}
+
```
jandubois: I think these tests are too tightly bound to the provisioning script. I don't think we need to test the content of the config map; we want to test when you configure `moby` that `dockerd` is running, and when you configure `containerd` that it is not.

You could try to run either `docker info` or `nerdctl info` using `rdd limavm shell` to confirm that the services are working. Similar for Kubernetes.

If you think these tests belong in a later PR, then that is fine too. But these specific tests seem to have little value to me.

bats/tests/32-app-controllers/app.bats
```diff
@@ -211,6 +241,31 @@ local_setup_file() {
...
+    for _ in $(seq 30); do
+        template=$(rdd ctl get configmap "${VM_NAME}-template" \
+            --namespace "${RDD_NAMESPACE}" -o jsonpath='{.data.template}' 2>/dev/null)
+        if grep -q "CONTAINER_ENGINE: containerd" <<<"${template}"; then
+            found=true
+            break
+        fi
+        sleep 1
+    done
+    [[ "${found}" == "true" ]]
```
jandubois: It should be possible to implement the retry loop using our `try` function.

jandubois (changes requested):
Sorry if there was a misunderstanding on what kind of tests we should add!

jandubois (dismissed):
Thanks, LGTM

mook-as (approved):

PR #274, opened 2026-04-07
Add `rdd set` command for App configuration

### LINKED ISSUE #275: Create `rdd set` command to change App settings
Should determine available properties from the App CRD.

```
rdd set running=true kubernetes.enabled=true kubernetes.version=1.35.1
```

This command will start the control plan and the app as needed.

jandubois:
Derive valid property names and types from the App CRD's OpenAPI v3 schema at runtime, so the command automatically supports new properties as they are added to the CRD. Create the `App` with defaults if it does not exist.

The `--help` output shows available properties with their types and descriptions, fetched live from the CRD schema.

The `--dry-run` flag validates changes against the admission controller without persisting them (but does create the `App` if absent, since the admission controller needs the resource to validate against).

### DIFF
--- a/cmd/rdd/set.go
+++ b/cmd/rdd/set.go
@@ -1,0 +1,586 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


cmd/rdd/set.go
```diff
@@ -0,0 +1,586 @@
...
+	obj := &unstructured.Unstructured{
+		Object: map[string]any{
+			"apiVersion": appv1alpha1.GroupVersion.String(),
+			"kind":       appv1alpha1.AppKind,
+			"metadata":   map[string]any{"name": "app"},
+			"spec":       createSpec,
+		},
+	}
+
+	_, err = dynClient.Resource(appGVR).Create(ctx, obj, metav1.CreateOptions{})
```
Nino-K, line 317: It looks as though `--dry-run` creates the app for real when it doesn't exist. In here, when `dryRun=true`, the create call has no option for `DryRun` option, it means the app is created with default values. Which also means after `rdd set --dry-run running=true` on a fresh system, the App exists with `running: false`, the user's `running=true` was never applied.
I'm only raising this because the flag description says "validate changes without persisting them," but the creation is persisted.
Am I understanding this correctly?

Nino-K (commented):
It looks as though `--dry-run` creates the App for real when it doesn't exist.

cmd/rdd/set.go
jandubois: Yes, this is correct. We cannot access the admission controller unless the resource exists. I think it is a reasonable assumption that the user wants to eventually create the app when they do an `rdd set --dry-run`.

I expect the main usage for `--dry-run` to be BATS tests for settings validation.

The reason that the admission controller has to support DryRun is that the Preferences dialog will use it to validate settings before the user has pressed "Apply".

Nino-K (approved):

PR #278, opened 2026-04-08
Bundle k3s versions to app controller

### LINKED ISSUE #260: Bundle a static version of k3s-versions.json to App controller
Grab a k3s-versions.json file from Rancher Desktop 1.x and embed it with the App controller. We will need it to validate k3s versions.

Remove all versions below 1.32.0; we don't support any older versions right now. We might lower the threshold later, but only if we get significant pushback.

Nino-K:
Embed a filtered copy of k3s-versions.json from Rancher Desktop 1.x. Versions below 1.32.0 are removed.
Add a `parseK3sVersions` helper that turns the version list into a lookup set. The reconciler validates `spec.kubernetes.version` against this set when Kubernetes is enabled and sets a Ready=False condition with reason `UnsupportedKubernetesVersion` if the version is not found.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/260


jandubois:
@Nino-K I don't know if this is your latest branch, but the failures are due to the template using the finch image instead of the opensuse one and are therefore missing the rancher-desktop.target unit.

This has already been fixed in `main`, so at a minimum this requires a rebase. Not looking any further because you may have other local changes that you haven't pushed to GitHub yet.

Nino-K:
> @Nino-K I don't know if this is your latest branch, but the failures are due to the template using the finch image instead of the opensuse one and are therefore missing the rancher-desktop.target unit.
> 
> This has already been fixed in `main`, so at a minimum this requires a rebase. Not looking any further because you may have other local changes that you haven't pushed to GitHub yet.

Yes, I just undrafted this and rebased it with all the necessary changes (previous PRs that needed). But, it is ready now.

jandubois:
AI review at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260413-120825-pr-278.html

Most important architecture point: the validation needs to happen in an admission controller, not in the reconciler. Make sure you add BATS tests with --dry-run.

You can copy code/ideas/bats from the notary controller, which is where I prototyped this kind of logic.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -101,6 +103,11 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 		}
 	}
 
+	hostHome, err := os.UserHomeDir()
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get host home directory: %w", err)
+	}
```
jandubois: I don't know why we need to pass the home directory to `applySpecToTempalate`, it could just call `os.UserHomeDir()` itself.

Otherwise add it to the `AppReconciler` struct and only initialize it once. Maybe make `applySpecToTemplate` a reconciler method?

I suspect we will refactor this again in the future anyways, so this can be done in a follow-up commit.

jandubois (dismissed):
Thanks, LGTM

pkg/controllers/app/app/controllers/app_controller.go
Nino-K: Sure, I will make this change in the PR where I add the validating admission controller for app settings.

jandubois (approved):
Thanks, LGTM

PR #280, opened 2026-04-08
Add host-switch virtual network for WSL2 instances

jandubois:
The opensuse distro expects a host-switch process on the Windows host to provide networking via a virtual L2 network over AF_VSOCK. Without it, the distro's network-setup.service blocks on a vsock handshake during boot and never gets an IP address.

Port the host-switch logic from rancher-desktop's networking module into the LimaVM controller as an in-process goroutine. The goroutine performs the vsock handshake with the guest, then runs a gvisor-tap-vsock virtual network providing DNS, DHCP, and NAT.

The host-switch starts before the hostagent (the guest blocks on the handshake during early boot) and stops after the hostagent and watcher.

Also switch the WSL2 App image from Finch rootfs to the opensuse distro.

---

Also: Fix go-mod-k8s-sync checkout for cross-repository PRs

The checkout steps hardcoded the base repository, so fetching the head branch failed for cross-repository pull requests.  Specify the head repository explicitly so the fetch targets the fork where the branch exists.

### DIFF
--- a/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
+++ b/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
@@ -1,0 +1,462 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+//go:build windows
+
+package controllers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/Microsoft/go-winio"
+	"github.com/containers/gvisor-tap-vsock/pkg/types"
+	"github.com/containers/gvisor-tap-vsock/pkg/virtualnetwork"
+	"github.com/go-logr/logr"
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/linuxkit/virtsock/pkg/hvsock"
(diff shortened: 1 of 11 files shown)


pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -0,0 +1,368 @@
...
+		<-ctx.Done()
+		return ln.Close()
+	})
+
+	// Start the port-forwarding API server inside the virtual network.
+	apiAddr := fmt.Sprintf("%s:80", cfg.GatewayIP)
+	vnLn, err := vn.Listen("tcp", apiAddr)
+	if err != nil {
+		logger.Error(err, "Failed to listen on API address", "addr", apiAddr)
+		return
```
Nino-K: This will cause `g.Wait()` to never be called.

pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -0,0 +1,368 @@
...
+	if err != nil {
+		logger.Error(err, "Vsock handshake failed")
+		return
+	}
+
+	cfg := newVirtualNetworkConfig(*subnet)
+	vn, err := virtualnetwork.New(&cfg)
+	if err != nil {
+		logger.Error(err, "Failed to create virtual network")
+		return
```
Nino-K, line 174: We would need to close the listener before this return. So either an explicit `ln.Close()` or add a `defer ln.Close()` up above.

Nino-K (approved):

PR #283, opened 2026-04-09
Add BATS for provisioning scripts

Nino-K:
Adds the BATS test for the provisioning scripts to verify the running container engines. Also, update the makefile to trigger the build when the embedded sources (files) are changed.

### DIFF
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -195,15 +195,43 @@ local_setup_file() {
     assert_output "${app_gen}"
 }
 
-@test "propagating running=false updates LimaVM spec.running" {
-    rdd ctl patch app "${APP_NAME}" --type='merge' -p='{"spec":{"running":false}}'
-    rdd ctl wait --for=jsonpath='{.spec.running}'=false \
-        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=30s
+@test "wait for dockerd to be ready with moby container engine" {
+    try --max 10 --delay 3 -- rdd limavm shell "${VM_NAME}" sudo docker info
 }
 
-@test "wait for LimaVM Running condition to become false after stop" {
+@test "containerd is not running with moby container engine" {
+    run rdd limavm shell "${VM_NAME}" sudo systemctl is-active containerd
+    assert_line "inactive"
+}
+
+@test "switch container engine to containerd without stopping VM" {
+    rdd set containerEngine.name=containerd
+    # The reconciler detects the template change, stops the VM, and restarts it.
+    rdd ctl wait --for=condition=Running=False \
+        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=2m
+    rdd ctl wait --for=condition=Running \
+        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=7m
+}
+
+@test "wait for containerd to be ready with containerd engine" {
+    try --max 10 --delay 3 -- rdd limavm shell "${VM_NAME}" sudo nerdctl --address /run/k3s/containerd/containerd.sock info
+}
+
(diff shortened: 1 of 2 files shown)


bats/tests/32-app-controllers/app.bats
```diff
@@ -195,6 +195,14 @@ local_setup_file() {
     assert_output "${app_gen}"
 }
 
+@test "wait for dockerd to be ready with moby container engine" {
+    try --max 60 --delay 5 -- rdd limavm shell "${VM_NAME}" sudo docker info
```
jandubois: I think a 5 minutes timeout is too long, given that we have already confirmed that the App and VM are in running condition.

bats/tests/32-app-controllers/app.bats
```diff
@@ -195,6 +195,14 @@ local_setup_file() {
...
+}
+
+@test "containerd is not running with moby container engine" {
+    run -1 rdd limavm shell "${VM_NAME}" sudo systemctl is-active containerd
```
jandubois: Is there a way to more directly check that the service is not running, like matching on an error message? `sustemctl` could return an exit code of 1 for a number of conditions.

bats/tests/32-app-controllers/app.bats
```diff
@@ -212,6 +220,26 @@ local_setup_file() {
...
 }
 
+@test "switch to containerd engine and start VM" {
+    rdd set containerEngine.name=containerd running=true
+    rdd ctl wait --for=condition=Running \
+        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=7m
+}
+
+@test "wait for containerd to be ready with containerd engine" {
+    try --max 60 --delay 5 -- rdd limavm shell "${VM_NAME}" sudo nerdctl --address /run/k3s/containerd/containerd.sock info
```
jandubois: Again timeout seems to be too long.

jandubois (commented):
LGTM, only minor feedback.

I wonder if you should have a test where you change the container engine (or kubernetes version) without explicitly stopping and restarting the app, and instead verify that the reconcilers do the whole thing?

bats/tests/32-app-controllers/app.bats
```diff
@@ -195,17 +195,45 @@ local_setup_file() {
...
-@test "propagating running=false updates LimaVM spec.running" {
-    rdd ctl patch app "${APP_NAME}" --type='merge' -p='{"spec":{"running":false}}'
-    rdd ctl wait --for=jsonpath='{.spec.running}'=false \
-        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=30s
+@test "wait for dockerd to be ready with moby container engine" {
+    try --max 10 --delay 3 -- rdd limavm shell "${VM_NAME}" sudo docker info
+}
+
+@test "containerd is not running with moby container engine" {
+    run rdd limavm shell "${VM_NAME}" sudo systemctl is-active containerd
```
jandubois, line 203: I think this will return exit code 3 for "service exists but is inactive", so this might be better (but untested):


```suggestion
    run -3 rdd limavm shell "${VM_NAME}" sudo systemctl is-active containerd
```

bats/tests/32-app-controllers/app.bats
```diff
@@ -195,17 +195,45 @@ local_setup_file() {
...
 }
 
-@test "wait for LimaVM Running condition to become false after stop" {
+@test "stop VM and restore moby engine after containerd test" {
+    rdd set running=false containerEngine.name=moby
     rdd ctl wait --for=condition=Running=False \
         limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=60s
 }
 
+@test "propagating running=false updates LimaVM spec.running" {
```
jandubois, line 231: Test name is misleading; this is just checking the spec, the propagation already happened above.

jandubois (approved):
Thanks, LGTM

PR #285, opened 2026-04-10
ci: add Windows perf instrumentation and disable Compat Telemetry

jandubois:
The `bats-lima-controller` suite has been intermittently timing out on `windows-latest` because gzip decompression of the cached Lima rootfs occasionally takes 100+ seconds against a 120s test deadline. The `windows-latest` runner has only 1 physical core / 2 logical processors and 8 GB RAM, so any background CPU consumer can starve gzip badly.

Investigation showed that `CompatTelRunner.exe` (Microsoft Compatibility Telemetry) was the single largest CPU consumer over a 24-minute BATS job — ~340 CPU-seconds, roughly 22% of one core averaged across the run. Disabling it freed enough headroom that the worst-case `gzip` in limavm-running.bats` dropped from 181s to 5-12s, and the overall runner saturation dropped meaningfully (samples at queue=0 went from 16.5% to 26.8%, samples at CPU>=99% from 24% to 16%).

This commit adds two pieces:

1. Disable `CompatTelRunner` via the Image File Execution Options "Debugger" hook -- a documented Windows mechanism that redirects any attempt to launch `CompatTelRunner.exe` through `taskkill.exe`, which exits immediately. Tried first via `Disable-ScheduledTask` (silently fails on SYSTEM-owned tasks), then via `schtasks.exe` (legacy task names no longer exist on Server 2025), then via binary rename (works but heavy-handed). The IFEO hook is the community-recommended approach, registry-only, launcher-agnostic, and additionally avoids ~80 CPU-seconds of `svchost` retry overhead that the rename approach generated.

2. Two PowerShell diagnostic scripts plus workflow steps to run them:

   - `windows-machine-info.ps1`: one-shot dump of CPU/memory/disk/Defender /WSL state plus a 256 MB sequential write+read benchmark, written to `rdd-logs/_diag/machine-info.txt` at the start of the BATS job.

   - `windows-perf-sampler.ps1`: continuous JSONL sampler that runs in the background for the duration of the bats job. Per-second samples include system CPU/memory/disk counters and the top 10 processes by working set + per-second CPU delta. Written to `rdd-logs/_diag/perf-sampler.jsonl`, uploaded as part of the existing `rdd-logs` artifact.

The instrumentation is generally useful for any Windows GHA runner where intermittent CPU contention may be at play, not just this PR.

### DIFF
--- a/.github/workflows/bats.yaml
+++ b/.github/workflows/bats.yaml
@@ -94,6 +94,58 @@ jobs:
         Get-Service -ErrorAction Continue -Name @('W3SVC', 'docker') |
           Stop-Service
 
+        # Free CPU by disabling Microsoft Compatibility Telemetry. The
+        # windows-latest runner has only 2 vCPUs, and CompatTelRunner.exe
+        # was the single largest CPU consumer in our diagnostic runs
+        # (~340 CPU-seconds over a 24-minute BATS job, ~22% of one core).
+        # On a 2-vCPU machine that crowds out gzip/xz decompression and
+        # causes intermittent BATS timeouts when the runner happens to
+        # be saturated during a Lima cache-hit decompression.
+        #
+        # Disabling scheduled tasks doesn't work on Windows Server 2025:
+        # the legacy task names ("Microsoft Compatibility Appraiser",
+        # "ProgramDataUpdater") no longer exist, and the discovery step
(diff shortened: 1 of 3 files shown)


jandubois:
Superseded by #289 and #290

PR #288, opened 2026-04-10
Windows bats perf instrumentation

mook-as:
Just playing with #285, please ignore me.


jandubois:
I think this can be closed in favour of #292.

I've used `typeperf` instead of `logman` because it was easier to use locally (does not require elevation) and produces the same data. Only concern is killing it by PID, but it seems to work fine in practice, new events seem to get flushed in a timely manner.

mook-as:
Obsoleted by #292 — the point is to not poll with a PowerShell script, and to offload that to something provided with the OS.

PR #289, opened 2026-04-10
ci: disable Compatibility Telemetry on Windows runners

jandubois:
Block CompatTelRunner.exe via an IFEO debugger redirect and disable the DiagTrack service before the BATS job starts. On a 2-vCPU runner, CompatTelRunner consumed ~22% of one core over a 24-minute job, crowding out gzip/xz decompression and causing intermittent timeouts.

Extract the logic into a reusable composite action at .github/actions/disable-compat-telemetry.

### DIFF
--- a/.github/actions/disable-compat-telemetry/action.yaml
+++ b/.github/actions/disable-compat-telemetry/action.yaml
@@ -1,0 +1,56 @@
+name: Disable Compatibility Telemetry
+description: >-
+  Block CompatTelRunner.exe via IFEO debugger redirect and disable the
+  DiagTrack service. Runs only on Windows; harmless no-op otherwise.
+
+runs:
+  using: composite
+  steps:
+  - name: Block CompatTelRunner via IFEO redirect
+    if: runner.os == 'Windows'
+    shell: pwsh
+    run: |
+      # Use Image File Execution Options to redirect CompatTelRunner.exe
+      # to where.exe (which just looks up args on PATH and exits).
+      # This is a documented Windows mechanism that works regardless of
+      # what launches the binary (scheduled task, service, WMI, etc.).
+      # See https://github.com/adolfintel/Windows10-Privacy/issues/4
+      #
+      # On a 2-vCPU GitHub runner, CompatTelRunner consumed ~340 CPU-seconds
+      # over a 24-minute BATS job (~22% of one core), crowding out gzip/xz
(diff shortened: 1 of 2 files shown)


.github/actions/disable-compat-telemetry/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+      # what launches the binary (scheduled task, service, WMI, etc.).
+      # See https://github.com/adolfintel/Windows10-Privacy/issues/4
+      #
+      # On a 2-vCPU GitHub runner, CompatTelRunner consumed ~340 CPU-seconds
+      # over a 24-minute BATS job (~22% of one core), crowding out gzip/xz
+      # decompression and causing intermittent timeouts.
+      $ifeo = 'HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\CompatTelRunner.exe'
+      New-Item -Path $ifeo -Force | Out-Null
+      New-ItemProperty -Path $ifeo -Name Debugger `
+        -Value "$env:WINDIR\System32\taskkill.exe" `
```
mook-as: So this is basically running `taskkill.exe CompatTelRunner.exe /some-args` right? And then `taskkill.exe` just spews "that makes no sense" and then it gets ignored?

I think something like `where.exe` (which just looks up any arguments on `PATH`) would be safer?

.github/actions/disable-compat-telemetry/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+        Stop-Process -Force -ErrorAction SilentlyContinue
+
+  - name: Disable DiagTrack service
+    if: runner.os == 'Windows'
+    shell: pwsh
+    run: |
+      # DiagTrack (Connected User Experiences and Telemetry) is
+      # CompatTelRunner's sibling chatty process.
+      sc.exe stop diagtrack
+      sc.exe config diagtrack start= disabled
```
mook-as: The PowerShell equivalents would be

```suggestion
      Set-Service DiagTrack -StartupType Disabled -Status Stopped
```

mook-as (dismissed):
Looks basically fine; if this actually makes a difference it looks fine.

I see current BATS as ~24m on `main`, and ~23m on branches with this applied, though; that doesn't seem like such a big difference.

mook-as (approved):

PR #290, opened 2026-04-10
ci: add Windows machine-info and perf-sampler instrumentation

jandubois:
Capture runner specs (CPU, memory, disk, Defender, WSL) at the start of the BATS job and sample perf counters once per second throughout. Both produce files under rdd-logs/_diag/ that upload with the existing artifact step, enabling correlation with rdd.stderr timestamps when diagnosing flaky timeouts.

Each is a reusable composite action:
- .github/actions/windows-machine-info (one-shot snapshot)
- .github/actions/windows-perf-sampler (start/stop background sampler)


jandubois:
Superseded by #292

PR #291, opened 2026-04-10
app: Create namespace as needed

### LINKED ISSUE #284: App should create the namespace
Given:

```yaml
apiVersion: app.rancherdesktop.io/v1alpha1
kind: App
metadata:
  creationTimestamp: "2026-04-09T22:00:57Z"
  finalizers:
  - rdd.rancherdesktop.io/cleanup
  generation: 3
  name: app
  resourceVersion: "103"
  uid: 306a4c47-3557-498b-b20f-856f1ddb64e4
spec:
  containerEngine:
    name: moby
  kubernetes:
    enabled: false
  namespace: rancher-desktop
  running: true
```

The app reconciler should be creating the `rancher-desktop` namespace, I believe?  Currently I need to do that manually.

I can work on this to not distract people.

mook-as:
This causes the app reconciler to create the namespace before trying to create LimaVM in it.  It also tries to delete the namespace (if it's owned by the app) on app delete.

To make sure this is exercised, the BATS default namespace has been changed (it was already an option).  However, this surfaced an issue where we were incorrectly assuming the default namespace in BATS.  (As noted, that was already an option that wasn't exercised very much, so it would be a bug.)

### DIFF
--- a/pkg/controllers/app/app/controllers/app_controller.go
+++ b/pkg/controllers/app/app/controllers/app_controller.go
@@ -64,28 +64,16 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	// Handle deletion, delete the LimaVM and wait for it to finish cleaning up.
+	// Handle deletion, delete owned resources.
 	if base.IsBeingDeleted(&app) {
 		log.Info("App resource is being deleted, performing cleanup")
 
 		namespace := app.GetResourceNamespace()
+
+		// Delete the LimaVM and wait for it to finish cleaning up.
 		limaVM := &limav1alpha1.LimaVM{}
 		err := r.Get(ctx, client.ObjectKey{Name: limaVMName, Namespace: namespace}, limaVM)
-		switch {
-		case apierrors.IsNotFound(err):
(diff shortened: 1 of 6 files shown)


pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -110,6 +125,27 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		ns = &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+			},
+		}
+		if err := ctrl.SetControllerReference(&app, ns, r.Scheme); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to set owner reference on namespace: %w", err)
+		}
+		if err := r.Create(ctx, ns); err != nil && !apierrors.IsAlreadyExists(err) {
+			log.Error(err, "Failed to create namespace", "namespace", namespace)
```
jandubois: Logging and then returning the same message as the error seems redundant. I would expect the caller somewhere up the chain to log the error.

Same thing for the `else if` branch below.

jandubois (dismissed):
Thanks, LGTM

I suspect the 2 logging calls will produce duplicate messages. Doesn't hurt though because I expect those error branches will not be executed anyways.

jandubois (approved):
Thanks, LGTM

PR #292, opened 2026-04-11
Windows typeperf sampler

jandubois:
## Summary

- Add a `windows-typeperf-sampler` composite action that runs `typeperf` in the background during BATS jobs, collecting system and per-process counters once per second into a BLG (binary log) file

- On stop, `relog` converts the BLG to TSV and a Go script summarizes the TSV into a human-readable top-N-by-CPU and top-N-by-working-set report

- Add a `windows-machine-info` composite action that captures hardware/OS/Defender/WSL state once at job start for correlation with flaky timeouts

## Why BLG, not direct-to-TSV?

`typeperf`'s direct-TSV mode locks the column set at the first sample, so processes that start after collection begins (rdd, go, gzip) are missed entirely. BLG records the wildcard dynamically; `relog` expands the column set across the full collection window.

## Why typeperf, not PowerShell Get-Counter?

- ~1.5% of one core vs PowerShell's `Get-Counter` loop (important on 2-vCPU runners)
- Captures all processes, not just a top-N snapshot
- Native Windows tool, no PowerShell runtime overhead

## Artifacts produced in `rdd-logs/_diag/`

| File | Source | Description |
|------|--------|-------------|
| `machine-info.txt` | machine-info.ps1 | Hardware/OS/Defender/WSL snapshot |
| `typeperf.tsv` | typeperf → relog | 1s raw counter data for all processes |
| `typeperf-summary.txt` | Go post-processor | Top 10 by CPU and working set per interval |


.github/actions/windows-typeperf-sampler/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+  using: composite
+  steps:
+  - name: Start typeperf
+    if: runner.os == 'Windows' && inputs.command == 'start'
+    shell: pwsh
+    run: |
+      $outputDir = "${{ inputs.output-dir }}"
+      if (-not (Test-Path $outputDir)) {
+        New-Item -ItemType Directory -Force -Path $outputDir | Out-Null
+      }
```
mook-as, line 26: Since you're doing `-Force` anyway, you don't even need to test if the path exists.  `New-Item … -Path $outputDir` would have worked totally fine.

.github/actions/windows-typeperf-sampler/summarize-typeperf.go
```diff
@@ -0,0 +1,242 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// summarize-typeperf reads a TSV file produced by typeperf (via relog
```
mook-as, line 5:
```suggestion
// Command summarize-typeperf reads a TSV file produced by typeperf (via relog
```

mook-as (approved):
FWIW, `logman` works fine in CI (without permissions issues), since #288 worked fine.  But I have zero issues using `typeperf`, even though it means we'll need to kill it and hopefully it wouldn't corrupt anything.

I didn't look very hard at the golang code, but the output seems fine.

PR #299, opened 2026-04-13
Bats runtime fixes

jandubois:
* Disable spotlight metadata collection on macOS runners.
* Prevent inheritance of &3 and &4 by daemon processes, which block BATS from terminating.


mook-as (approved):

jandubois:
This PR failed BATS on macOS once when the runner disconnected (for unknown reasons). Rerunning the failed job succeeded.

PR #300, opened 2026-04-13
ci: capture support bundle for BATS targets

jandubois:
The plain `timeout 1800` wrapper kills bats with SIGKILL and leaves no trace of what was hung, and a test that finishes green while leaking an orphaned qemu also leaves no evidence. Capture a support bundle on every exit (with a dedicated "timeout" section when the watchdog fires) so CI artifacts always carry post-mortem state.

The bundle records:

  * Process tree, Go goroutine stacks via SIGQUIT, open fds, kernel wait channels, and pgid for each interesting process. SIGQUIT/SIGKILL are scoped to processes whose cmdline contains the current RDD_INSTANCE so parallel sibling targets stay untouched.

  * vm_stat/free, top processes by RSS, and the macOS unified log or Linux dmesg filtered for jetsam/OOM events. Go cannot catch SIGKILL, so a daemon killed by the kernel leaves no crash log behind; the kernel messages are the only evidence.

  * rdd ctl get for apps/limavms/containers/images/volumes/ containernamespaces (overview + YAML), rdd ctl get events, and docker ps/inspect when the forwarded socket exists. Every call goes through a poll-loop run_with_timeout helper so a dead or hung API server cannot freeze the bundle capture, and avoids a GNU coreutils dependency (macOS has no native timeout command).

Leak attribution needs the hostagent's pgid. The limavm controller now logs pid and pgid when it starts a hostagent; SetGroup makes the child a new group leader so pgid equals pid, and every qemu forked by that hostagent inherits the pgid. A leaked qemu's pgid can be grepped directly in the rdd.stderr log to find the instance and time it came from.

The wrapper resolves the rdd binary from its own script location rather than PATH, since CI runners do not add <repo>/bin to PATH and bare `rdd` calls fail before any test runs. A fallback to rdd.exe supports the MSYS2 Windows runner.

### DIFF
--- a/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -478,10 +478,7 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
 	// The watcher enqueues reconciles as phase transitions occur.
 	r.startWatcher(ctx, limaVM.Name, limaVM.Namespace, haCmd, inst.Dir, begin)
 
-	logger.Info("Hostagent started, watcher active",
-		"instance", limaVM.Name,
-		"pid", haCmd.Process.Pid,
-		"pgid", haCmd.Process.Pid)
+	logger.Info("Hostagent started, watcher active", "instance", limaVM.Name, "pid", haCmd.Process.Pid)
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -447,7 +447,9 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
 	// network-setup.service performs a vsock handshake during early boot.
 	r.startHostSwitch(ctx, limaVM.Name, inst)
 
-	// Start hostagent in background.
+	// SetGroup makes the hostagent a new process-group leader (pgid == pid
+	// on Unix), which lets bats-with-timeout.sh attribute leaked qemu
+	// grandchildren back to their hostagent ancestor via pgid.
 	haCmd := exec.CommandContext(ctx, rddPath, args...)
 	process.SetGroup(haCmd)
 	haCmd.Stdout = haStdoutW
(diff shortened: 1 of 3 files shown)


jandubois:
For some reason I keep thinking that the script should live inside the `bats/` directory, because in Rancher Desktop we make this a free-standing testkit. On the other hand I like having all scripts in `scripts/` and not scattered around. Opinions?

jandubois:
Rebased on latest `main` to get the bump to opensuse image v0.2.2

scripts/bats-with-timeout.sh
```diff
@@ -0,0 +1,412 @@
...
+# (process tree, pgid, wchan, open fds) at the end so CI artifacts carry
+# the evidence needed to diagnose hangs and leaked processes. On timeout,
+# additionally kills the target with SIGTERM then SIGKILL.
+#
+# Non-destructive state capture is unfiltered so the bundle shows sibling
+# parallel bats targets too; analysis matches them by cmdline substring.
+# Destructive steps (SIGQUIT, SIGKILL of leaked processes) are scoped to
+# our own RDD_INSTANCE so they do not disturb sibling targets.
+#
+# Usage: bats-with-timeout.sh <seconds> <label> <command> [args...]
```
mook-as: What is `label`? It may be useful to have a sample.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -476,7 +478,10 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
 	// The watcher enqueues reconciles as phase transitions occur.
 	r.startWatcher(ctx, limaVM.Name, limaVM.Namespace, haCmd, inst.Dir, begin)
 
-	logger.Info("Hostagent started, watcher active", "instance", limaVM.Name)
+	logger.Info("Hostagent started, watcher active",
+		"instance", limaVM.Name,
+		"pid", haCmd.Process.Pid,
+		"pgid", haCmd.Process.Pid)
```
mook-as: Is `pgid` useful if it's always the same as `pid`? Especially since it makes no sense on Windows.

scripts/bats-with-timeout.sh
mook-as, line 1: I think I want to eventually rewrite the info-gathering in golang, so we get better results on Windows; but there's no need to do so right now.

scripts/bats-with-timeout.sh
```diff
@@ -0,0 +1,412 @@
...
+                dmesg 2>/dev/null | grep -iE 'oom|killed process|out of memory' | tail -50 || true
+            fi
+            ;;
+    esac
+}
+
+# Run a command with a wall-clock timeout. Used by dump_api_state so a
+# dead or hung API server cannot freeze the support bundle capture.
+# Polls at 1-second resolution, matching the main script's timeout loop,
+# and avoids a GNU coreutils `timeout` dependency (not on macOS).
```
mook-as: We install `timeout` on macos, and require it:

https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/30d8408cefb01f5391c1ae59e9b10da1b70983d8/.github/workflows/bats.yaml#L78
https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/30d8408cefb01f5391c1ae59e9b10da1b70983d8/docs/development.md#L13-L16

mook-as (dismissed):
No show-stopping blockers.

mook-as (approved):

PR #302, opened 2026-04-13
Validating admission controller app settings

### LINKED ISSUE #262: Create validating admission controller fo App settings
The admission controller must validate the settings: `containerEngine.name` must be either `moby` or `containerd`.

`kubernetes.version` must be either empty, or a version from the bundled `k3s-versions.json` file. It is not allowed to be empty when `kubernetes.enabled` is true.

The admission controller must be tested with the --dry-run feature supported by Kubernetes itself.

Nino-K:
This PR creates a validating admission webhook for App settings. It validates `containerEngine.name` (moby|containerd), `kubernetes.version` against bundled k3s-versions.json, and requires a non-empty version when `kubernetes.enabled` is true. Move validation out of the reconciler into the webhook so invalid specs are rejected at admission time.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/262

### DIFF
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -463,3 +463,18 @@ spec:
     version: "${K3S_VERSION}"
 EOF
 }
+
+@test "webhook rejects invalid update via patch dry-run" {
+    # The App may not exist here and might be deleted after kubernetes tests, so we create one.
+    create_app false
+
+    run -0 rdd ctl get app app -o jsonpath='{.spec.containerEngine.name}'
+    assert_output "moby"
+
+    run -1 rdd ctl patch app app --dry-run=server \
+        --type=merge -p '{"spec":{"kubernetes":{"enabled":true,"version":"0.0.0"}}}'
+    assert_output --partial "not supported"
+
+    run -0 rdd ctl get app app -o jsonpath='{.spec.containerEngine.name}'
+    assert_output "moby"
+}
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -10,6 +10,8 @@ load '../../helpers/load'
 APP_NAME="app"
 VM_NAME="rd"
 INPUT_CM_NAME="rd"
+APP_VALIDATOR_CONFIG="app-validator"
+K3S_VERSION="1.32.0"
 
 delete_app() {
     # Stop the VM first so the LimaVM finalizer clears quickly.
@@ -296,6 +298,17 @@ EOF
     delete_app
 }
 
+@test "unsupported Kubernetes version is rejected by admission webhook" {
+    delete_app
+    run -1 rdd set running=false kubernetes.enabled=true kubernetes.version=1.31.99
+    assert_output --partial "not supported"
+}
+
+@test "unsupported version does not create App resource" {
+    run -1 rdd ctl get app "${APP_NAME}"
+    assert_output --partial "not found"
+}
+
 @test "valid Kubernetes version allows LimaVM creation" {
     skip_on_windows "Kubernetes tests require further setup on Windows"
     delete_app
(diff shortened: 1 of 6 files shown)


bats/tests/32-app-controllers/app.bats
```diff
@@ -345,3 +358,108 @@ EOF
...
+
+@test "webhook configuration is registered" {
+    rdd ctl wait --for=create ValidatingWebhookConfiguration "${APP_VALIDATOR_CONFIG}" --timeout=60s
+
+    run -0 rdd ctl get ValidatingWebhookConfiguration "${APP_VALIDATOR_CONFIG}" \
+        -o jsonpath='{.webhooks[0].name}'
+    assert_output "app-validator.app.rancherdesktop.io"
+}
+
+@test "dry-run accepts valid App settings" {
```
mook-as, line 372: Totally optional, because I could see argument going either way: I think tests are easier to read if they're structured as base object + modifications specific to the test.  In this case, pass in yq expressions to modify a base object; something like:

```bash
create_app() {
  local input stmt
  read -r -d '' input <<EOF
apiVersion: app.rancherdesktop.io/v1alpha1
kind: App
metadata:
  name: app
spec:
  running: false
  containerEngine:
    name: containerd
  kubernetes:
    enabled: false
EOF
  local stmt
  for stmt in "$@"; do
    input=$(yq "$stmt" <<<"$input")
  done
  echo "$input"
}

…

@test "dry-run accepts valid App settings" {
    run -0 rdd ctl apply --dry-run=server -f - < \
        <(create_app)
}

@test "dry-run accepts valid containerEngine moby" {
    run -0 rdd ctl apply --dry-run=server -f - < \
        <(create_app '.spec.containerEngine.name="moby"')
}
```

But again, I can see arguments going the other way, so feel free to ignore this.

pkg/controllers/app/app/controllers/app_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// ValidateDelete is a no-op; App deletion is governed by the cleanup finalizer.
+func (v *AppValidator) ValidateDelete(_ context.Context, _ *v1alpha1.App) (ctrlwebhookadmission.Warnings, error) {
+	return nil, nil
+}
+
+func (v *AppValidator) validate(app *v1alpha1.App) (ctrlwebhookadmission.Warnings, error) {
+	engine := app.Spec.ContainerEngine.Name
+	if engine != "moby" && engine != "containerd" {
```
mook-as: I'm surprised we don't have an enum/const for this.  Especially given:
https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/a2326f8d9c658bd3611203f5f49444dd4f8bc2bd/pkg/apis/app/v1alpha1/app_types.go#L18

But, indeed, I can find nothing of the sort… Also, as far as I can tell, while the CRDs have those things baked in, it doesn't actually validate. 🤷

pkg/controllers/app/app/controllers/app_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+		return nil, fmt.Errorf("spec.containerEngine.name %q is invalid; must be \"moby\" or \"containerd\"", engine)
+	}
+
+	k8s := app.Spec.Kubernetes
+	if k8s.Enabled && k8s.Version == "" {
+		return nil, errors.New("spec.kubernetes.version must not be empty when spec.kubernetes.enabled is true")
+	}
+
+	if k8s.Version != "" {
```
mook-as, line 54: That means we verify the version when it's disabled. Probably a good idea, I guess…

pkg/controllers/app/app/controllers/k3s_versions.go
```diff
@@ -0,0 +1,36 @@
...
+	if err := json.Unmarshal([]byte(data), &parsed); err != nil {
+		return nil, fmt.Errorf("failed to parse k3s versions: %w", err)
+	}
+	set := make(map[string]struct{}, len(parsed.Versions))
+	for _, v := range parsed.Versions {
+		// Strip leading "v" and "+k3s*" suffix: "v1.32.0+k3s1" → "1.32.0"
+		bare := strings.TrimPrefix(v, "v")
+		if idx := strings.Index(bare, "+"); idx >= 0 {
+			bare = bare[:idx]
+		}
```
mook-as:
```suggestion
		bare, _, _ = strings.Cut(bare, "+")
```

mook-as (dismissed):

jandubois (commented):
AI review: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260414-144550-pr-302.html

Only suggestions, but I think some are worthwhile to address.

pkg/controllers/app/app/controllers/app_webhook.go
jandubois: No, Kubernetes doesn't allow you assign arbitrary values to enum fields. You don't need to validate again.

We still should have constants for the engine names, but we can add them when we need them.

mook-as: Doesn't that mean the `engine` check should be removed, then?

pkg/controllers/app/app/controllers/app_webhook.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (v *AppValidator) validate(app *v1alpha1.App) (ctrlwebhookadmission.Warnings, error) {
+	engine := app.Spec.ContainerEngine.Name
+	if engine != "moby" && engine != "containerd" {
+		return nil, fmt.Errorf("spec.containerEngine.name %q is invalid; must be \"moby\" or \"containerd\"", engine)
+	}
+
+	k8s := app.Spec.Kubernetes
+	if k8s.Enabled && k8s.Version == "" {
+		return nil, errors.New("spec.kubernetes.version must not be empty when spec.kubernetes.enabled is true")
```
mook-as, line 51: I just realized that we're returning one error at a time, instead of all of them at once, so if there are multiple issues the users needs to fix them sequentially. That's a bit annoying, but probably okay? We can merge this PR without fixing this.

mook-as (changes requested):

jandubois (dismissed):
Thanks, LGTM

AI review: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260415-120235-pr-302.html

Nothing important, but I1 shows some dead code that can be removed. I would prefer removing because it can give the wrong impression that admission controllers need to implement default schema constraints, and I don't want that repeated when we add more properties.

pkg/controllers/app/app/controllers/app_webhook.go
jandubois: Oh, I totally missed that. For now it doesn't matter because we don't have many validations, but this is certainly something we want to do in the future as we add more properties.

jandubois (approved):
Thanks, LGTM

While the feedback of combining validation errors is valid, in the current form we can only have one or the other error, so the joining can be implemented later.

mook-as (approved):
Yeah, I was expecting my review to just be dismissed once the engine check went away. 👍

PR #304, opened 2026-04-13
app: Use `instance.ShortDir()` for docker socket

### LINKED ISSUE #294: Docker socket should be created in the ShortDir
Right now the location is hardcoded to `"{{.Home}}/.rd/docker.sock"`.

At a minimum it should be using `.rd2`, but really we need to use `instance.ShortDir()` and fix the path programmatically.

We already compose the template by concatenating multiple YAML snippets, so this would just mean moving the `portForwards` section out of the static template and generating it at runtime.

mook-as:
This mounts the docker socket at `instance.ShortDir()`, instead of hard- coding at `~/.rd/docker.sock`.  This matches the original design (see `docs/design/cmd_app.md`).

Fixes #294

### DIFF
--- a/pkg/controllers/app/app/controllers/app_controller.go
+++ b/pkg/controllers/app/app/controllers/app_controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
 	limav1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/lima/v1alpha1"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
 )
 
 const (
@@ -50,13 +51,16 @@ func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) (string, er
 	if err != nil {
 		return "", fmt.Errorf("failed to get host home directory: %w", err)
 	}
-	return baseTemplate + fmt.Sprintf(
-		"\nparam:\n  CONTAINER_ENGINE: %s\n  HOST_HOME: %q\n  KUBERNETES_ENABLED: %v\n  KUBERNETES_VERSION: %s\n",
-		spec.ContainerEngine.Name,
-		toLinuxPath(hostHome),
-		spec.Kubernetes.Enabled,
-		spec.Kubernetes.Version,
-	), nil
+	return strings.Join([]string{
+		baseTemplate,
+		"param:",
+		fmt.Sprintf("  CONTAINER_ENGINE: %s", spec.ContainerEngine.Name),
+		fmt.Sprintf("  HOST_HOME_GUEST: %q", toLinuxPath(hostHome)),
+		fmt.Sprintf("  HOST_SHORT_DIR: %q", instance.ShortDir()),
+		fmt.Sprintf("  KUBERNETES_ENABLED: %v", spec.Kubernetes.Enabled),
+		fmt.Sprintf("  KUBERNETES_VERSION: %s", spec.Kubernetes.Version),
+		"",
+	}, "\n"), nil
 }
 
 // toLinuxPath converts a host path to a Linux-accessible path inside a Lima VM.
@@ -70,6 +74,8 @@ func toLinuxPath(hostPath string) string {
 	if len(hostPath) >= 2 && hostPath[1] == ':' {
 		drive := strings.ToLower(string(hostPath[0]))
 		rest := strings.ReplaceAll(hostPath[2:], `\`, `/`)
+		// /mnt/ is the mount point for drvfs disks in WSL2, per the default
+		// value of `[automount] root=` in `/etc/wsl.conf`.
 		return "/mnt/" + drive + rest
 	}
 	return hostPath
--- a/pkg/controllers/app/app/lima-template.yaml
+++ b/pkg/controllers/app/app/lima-template.yaml
@@ -27,14 +27,13 @@ portForwards:
   hostPortRange:
   - 0
   - 0
-  hostSocket: "{{.Home}}/.rd/docker.sock"
+  hostSocket: "{{.Param.HOST_SHORT_DIR}}/docker.sock"
 provision:
 - mode: system
   script: |
     #!/bin/bash
 
     # All param need to be mentioned for Lima validation.
-    # $PARAM_HOST_HOME - used by the kubeconfig probe
 
     set -o errexit -o nounset -o pipefail -o xtrace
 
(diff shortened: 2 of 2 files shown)


pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +46,19 @@ type AppReconciler struct {
...
-	)
+func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) (string, error) {
+	params := map[string]string{
+		"CONTAINER_ENGINE":   spec.ContainerEngine.Name,
+		"HOST_HOME":          "{{.Home}}",
+		"HOST_SHORT_DIR":     filepath.ToSlash(instance.ShortDir()),
+		"KUBERNETES_ENABLED": strconv.FormatBool(spec.Kubernetes.Enabled),
+		"KUBERNETES_VERSION": spec.Kubernetes.Version,
+	}
+	serializedParams, err := yaml.Marshal(map[string]any{"param": params})
```
Nino-K: Just fyi, I intentionally used the `fmt.Sprintf` to 1) save us an import, and 2)better performance.

Nino-K (commented):
The "{{.Home}}" might not work, as I realized that when I was debugging bats. I explained further in this commit: https://github.com/rancher-sandbox/rancher-desktop-daemon/pull/278/changes/b7b69e6cade9230fd55ddda27197d6ff039e6d8b

pkg/controllers/app/app/controllers/app_controller.go
mook-as: Ah, okay, I'll redo this (once #278 is merged) that way, then, hopefully with more readable formatting.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +48,43 @@ type AppReconciler struct {
...
-		spec.Kubernetes.Version,
-	)
+func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) (string, error) {
+	hostHome, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get host home directory: %w", err)
+	}
+	params := map[string]any{
+		"CONTAINER_ENGINE":   spec.ContainerEngine.Name,
+		"HOST_HOME":          toLinuxPath(hostHome),
```
jandubois: We need to have some convention to designate that a path is expected to be valid inside the guest and not on the host, e.g. by using a `_GUEST` suffix:


```suggestion
		"HOST_HOME_GUEST":    toLinuxPath(hostHome),
```

That way we can have:

**HOST_HOME**: the home directory on the host
**HOST_HOME_GUEST**: the host home directory inside the guest
**GUEST_HOME_GUEST**: the home directory of the guest

I realize that `GUEST_HOME_GUEST` looks really ugly, but on the bright side, we are very likely not going to need it because scripts inside the guest can just use `HOME`.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +48,43 @@ type AppReconciler struct {
...
+		"HOST_SHORT_DIR":     toLinuxPath(instance.ShortDir()),
```
jandubois: This one actually needs to be a host path because it is used to construct the host path of the docker socket.


```suggestion
		"HOST_SHORT_DIR":     instance.ShortDir(),
```

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +48,43 @@ type AppReconciler struct {
...
-		spec.ContainerEngine.Name,
-		spec.Kubernetes.Enabled,
-		spec.Kubernetes.Version,
-	)
+func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) (string, error) {
+	hostHome, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get host home directory: %w", err)
+	}
+	params := map[string]any{
```
jandubois: Param values are always strings.

```suggestion
	params := map[string]string{
```

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +48,43 @@ type AppReconciler struct {
...
+		"HOST_SHORT_DIR":     toLinuxPath(instance.ShortDir()),
+		"KUBERNETES_ENABLED": spec.Kubernetes.Enabled,
+		"KUBERNETES_VERSION": spec.Kubernetes.Version,
+	}
+
+	var builder strings.Builder
+	builder.WriteString(baseTemplate)
+	builder.WriteString("\nparam:\n")
+	for _, key := range slices.Sorted(maps.Keys(params)) {
+		fmt.Fprintf(&builder, "  %s: %v\n", key, params[key])
```
jandubois: I wonder if we should quote the value to make it more likely to be valid YAML? Probably not necessary, but feels safer:


```suggestion
		fmt.Fprintf(&builder, "  %s: %q\n", key, params[key])
```

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -42,13 +48,43 @@ type AppReconciler struct {
...
+// Inside a WSL2 Lima VM the Windows filesystem is mounted at /mnt/<drive>/...,
+// so we convert it to WSL2 supported path. On other platforms the path is returned unchanged.
+func toLinuxPath(hostPath string) string {
+	if goruntime.GOOS != "windows" {
+		return hostPath
+	}
+	if len(hostPath) >= 2 && hostPath[1] == ':' {
+		drive := strings.ToLower(string(hostPath[0]))
+		rest := strings.ReplaceAll(hostPath[2:], `\`, `/`)
+		return "/mnt/" + drive + rest
```
jandubois, line 79: This should have a comment that `/mnt/` is just the default setting for the `root` value in the `[automount]` group in `wsl.conf`.

At some point we may want to parse that file at startup.

jandubois (changes requested):
I know this is still draft, but I had a quick look...

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -50,13 +51,16 @@ func applySpecToTemplate(baseTemplate string, spec v1alpha1.AppSpec) (string, er
...
-	return baseTemplate + fmt.Sprintf(
-		"\nparam:\n  CONTAINER_ENGINE: %s\n  HOST_HOME: %q\n  KUBERNETES_ENABLED: %v\n  KUBERNETES_VERSION: %s\n",
-		spec.ContainerEngine.Name,
-		toLinuxPath(hostHome),
-		spec.Kubernetes.Enabled,
-		spec.Kubernetes.Version,
-	), nil
+	return strings.Join([]string{
+		baseTemplate,
+		"params:",
```
jandubois: Typo:

```suggestion
		"param:",
```

mook-as:
Windows CI is failing:

> C:\\msys64\\usr\\bin\\ssh.exe … -o IdentityFile='/c/Users/runneradmin/.rdbats-lima-controller/lima/_config/user' …
> ** WARNING: connection is not using a post-quantum key exchange algorithm. …
> lima@127.0.0.1: Permission denied (publickey,password,keyboard-interactive).

It does appear to run various scripts from `…/cidata/boot.Linux/00-*.sh` etc., so at least the init stuff is running.

The post-quantum warning may be because the machine is running Leap?  I see the same on a mac, so it's not relevant.

jandubois (approved):
Thanks, LGTM

PR #305, opened 2026-04-14
limavm: use patch to tolerate concurrent writes during Prepare

jandubois:
**Bug.** On Windows CI, `limavm_controller.Reconcile` holds a local `LimaVM` copy for ~3 minutes while `limainstance.Prepare()` downloads and decompresses the opensuse distro image. If any write lands on the `LimaVM` during that window — for example, the app_controller propagating `spec.running=true` after a test patches the `App` — the post-Prepare `Status().Update()` fails with 409 Conflict, the reconcile aborts before removing the sentinel file, and the next reconcile interprets the sentinel as "Prepare crashed" and deletes the freshly prepared instance directory. Observed in run 24412470185 on `windows-latest`.

**Fix.** Switch the post-Prepare status write to `client.MergeFrom` + `Status().Patch()`. Without `OptimisticLock`, `MergeFrom` omits `metadata.resourceVersion` from the patch body, so concurrent writes to unrelated fields stop triggering 409s. This matches the `updateCondition` pattern already used ~12× in `limavm_lifecycle.go` for the same reason. The same cleanup lands on the app_controller's template-ConfigMap update, which had the same conflict-prone shape (not a test-breaker — controller-runtime re-queues on error — but noisy in the logs and trivially fixable).

**Test timeout.** Bump the `wait for LimaVM Created condition to be set` test from 150s to 5m. 150s was calibrated for macOS/Linux where Prepare takes ~34s; Windows consistently takes ~3 minutes because xz runs through Lima's stdin progress-reader pipe and sits idle ~90% of the time. A comment in the test points at the upstream Lima slowness, which is tracked separately for a future contribution.

### DIFF
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -128,9 +128,13 @@ local_setup_file() {
 }
 
 @test "wait for LimaVM Created condition to be set" {
-    # 150s allows time for the opensuse distro image download (~350MB).
+    # Download the opensuse distro image (~350MB) and decompress it. The
+    # Windows CI runner is much slower than macOS/Linux here: xz runs
+    # through Lima's stdin-wrapper pipe and is starved on every read,
+    # stretching a ~34s operation to ~3 minutes. Budget generously until
+    # that upstream slowness is addressed.
     rdd ctl wait --for=condition=Created \
-        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=150s
+        limavm/"${VM_NAME}" --namespace "${RDD_NAMESPACE}" --timeout=5m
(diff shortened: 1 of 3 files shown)


mook-as (approved):
Looks fine.

> A comment in the test points at the upstream Lima slowness, which is tracked separately for a future contribution.

I see a description; is there an issue filed (here, or upstream) that we are tracking?

jandubois:
> > A comment in the test points at the upstream Lima slowness, which is tracked separately for a future contribution.
> 
> I see a description; is there an issue filed (here, or upstream) that we are tracking?

Not yet. I'm working on getting some comparison benchmarks from AI first and post the results as an issue, if this turns out to be actionable.

PR #306, opened 2026-04-14
BATS: run bats-app on its own CI runner

jandubois:
bats-app and bats-lima each boot a WSL2 VM, and Lima's WSL2 driver hard-codes SSH to guest port 22 per instance. When `make --jobs` launches both suites concurrently on Windows, the two VMs collide on localhost:22 forwarding and one of them sees persistent "Permission denied" during SSH readiness probes until the 1800s wrapper kills the job. Run 24422550777 (PR #278 merge) is a recent instance.

Split bats-app out of the bats-all dependency chain and add a matrix dimension to the BATS workflow so bats-all and bats-app run on separate runners. Locally, the default target still runs both via recursive make so that typing `make` keeps covering the full suite without the collision. The bats-controllers grouping loses bats-app for the same reason.

As a side effect, bats-app (which is expected to grow) no longer blocks on the rest of the suite in CI.

### DIFF
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -13,6 +13,7 @@ BATS_TIMEOUT ?= 1800
 BATS = $(if $(filter-out 0,$(BATS_TIMEOUT)),../scripts/bats-with-timeout.sh $(BATS_TIMEOUT)) $(BATS_CORE)
 
 default: bats-all
+.PHONY: default
 
 # Check if BATS core exists
 check-bats:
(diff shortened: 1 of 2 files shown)


bats/Makefile
```diff
@@ -80,8 +89,8 @@ bats-containers: check-bats build-rdd build-mock-controller
 	RDD_INSTANCE=bats-containers-controller $(BATS) tests/34-containers-controllers/
 .PHONY: bats-containers
 
-# Run all controller tests.
-bats-controllers: bats-builtin bats-rdd bats-app bats-lima bats-containers
+# Run all controller tests except bats-app (see `default` target above).
+bats-controllers: bats-builtin bats-rdd bats-lima bats-containers
```
mook-as: `bats-all` is still useful locally; can we add a new `bats-ci` target or something along those lines for this? Or some other way to exclude `bats-app` only in CI?

jandubois: Sure, kept `bats-all` and now have `bats-app` and `bats-non-app`.

Not really happy with the `non-app` name, but can't think of anything better. At least it is descriptive.

mook-as (approved):
I hate it less? But yeah, don't see any better ways forward in the short term.

PR #307, opened 2026-04-15
scripts/bats-with-timeout: don't start rdd

mook-as:
If rdd is not running, `rdd ctl ...` is going to start it first.  That leads to things falling over if we're trying to probe a successful run since it can mean we end up trying to examine the logs (in a later CI step) while the instance is being torn down.

To fix this, check that the instance is up before doing all of this.

Also, for future bullet-proofing, try to get the list of API resources that contain our domain for the object types.

### DIFF
--- a/scripts/bats-with-timeout.sh
+++ b/scripts/bats-with-timeout.sh
@@ -210,22 +210,28 @@ dump_api_state() {
         return
     fi
 
-    echo "=== rdd ctl get (overview) ==="
-    timeout --kill-after=1 10 \
-        "$rdd_bin" ctl get apps,limavms,containers,images,volumes,containernamespaces \
-        --all-namespaces 2>&1 || true
+    if "$rdd_bin" service status 2>/dev/null | grep --quiet 'control plane has been started: true'; then
+        local resource_types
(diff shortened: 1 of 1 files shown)


scripts/bats-with-timeout.sh
```diff
@@ -210,22 +210,28 @@ dump_api_state() {
         return
     fi
 
-    echo "=== rdd ctl get (overview) ==="
-    timeout --kill-after=1 10 \
-        "$rdd_bin" ctl get apps,limavms,containers,images,volumes,containernamespaces \
-        --all-namespaces 2>&1 || true
+    if "$rdd_bin" service status 2>/dev/null | grep --quiet 'control plane has been started: true'; then
```
mook-as, line 213: As a future improvement point, I'd like to make `rdd service status` be able to just show one line instead of it all (depending on parameters), and also to be able to respond via exit code.  But I didn't want to put that in this PR.

jandubois: Furthermore I've been thinking of moving `rdd svc status` to `rdd status` so it can cover App conditions too.

jandubois (approved):
Thanks, LGTM

PR #308, opened 2026-04-15
CI: BATS: split out lima too, drop parallel runs

mook-as:
Split out lima too, because that's also one of the slower suites due to having to start up the VM.  Also drop running things in parallel, since that seems to just lead to slower total run time instead due to slow CI.

While we're at it, redo the makefile so that we can just exclude specific targets (as provided in `$BATS_EXCLUDES`) rather than having a special target.  This silences some `make` warnings.


jandubois (approved):
Thanks, LGTM

PR #309, opened 2026-04-15
engine: mirror Docker container engine state into Kubernetes

### LINKED ISSUE #279: Let `rdd set` wait for the desired state
In most scripting you will want to wait for the App to reach the desired state before running the following commands.

I think `rdd set` should wait by default, just like most other `rdd` commands. This will require an `AppReady` or similar condition on the App resource (see #264).

A `--wait=false` option would skip the wait, again the same as most other `rdd` command that wait for some state before exiting. We also need to add the corresponding `--timeout` option, but with a larger default than other commands; maybe even use 5min here.

### LINKED ISSUE #296: Mirror moby containers/images/volumes to rdd resources
Watch the container engine and make sure each object has a corresponding mirrored resource in the control plane. Watch for changes and apply the changes to the mirrors. When a mirror gets deleted, delete the actual object in dockerd.

jandubois:
Add the `engine` controller under `pkg/controllers/app/engine/`, plus supporting changes in `rdd set`, the discovery ConfigMap, the App and Container CRDs, and the design docs. The controller mirrors Docker containers, images, and volumes into `Container`, `Image`, and `Volume` resources under the containers.rancherdesktop.io API group.

## Engine controller

The reconciler watches App for `Running` plus `containerEngine.name=moby` and runs a `dockerWatcher` goroutine that connects to the host Docker socket, performs a full sync, then streams Docker events until the context cancels. A dead watcher signals a transient disconnect: the reconciler reconnects and re-runs fullSync, which is idempotent and preserves mirror identities. Every other App state (VM stopped, backend switched to containerd) stops the watcher and sweeps every mirror resource.

The reconciler applies mirror resources with server-side apply under a dedicated field manager (`engine-controller`). Finalizer-only applies use a separate manager (`engine-controller-finalizer`) to retain spec ownership and avoid tripping CRD required-field validation. Each mirror carries the `engine.rancherdesktop.io/docker-mirror` finalizer, which forwards K8s-side deletes to Docker before removing the mirror.

The reconciler writes a new `ContainerEngineReady` condition on App with reasons `Connected`, `ConnectFailed`, `Stopped`, or `NotApplicable`. `NotApplicable` reports `Status=True` on non-moby backends so that `rdd set running=true containerEngine.name=containerd` finishes waiting; consumers that need mirrors must gate on the reason, not the status alone.

## `Container.spec.state = unknown`

`ContainerSpec.State` gains an `unknown` value, now the default. Engine-created mirrors use `unknown`, meaning "mirror Docker status but take no action." Setting the field to `running` or `created` expresses user intent; the reconciler dispatches ContainerStart, ContainerUnpause, or ContainerStop until the container reaches the desired state. Any non-running state (created, exited, dead) satisfies `created`. The immutable-spec webhook now permits `spec.state` transitions while holding every other field immutable.

Container ports (both spec and status) become atomic maps (`x-kubernetes-list-type=map`, `listMapKey=name`) so SSA merges them correctly instead of replacing the whole array.

## `rdd set` waits for desired state

`rdd set` gains a `--timeout` flag (default 5m, `0` disables) and waits synchronously on the App after patching:

- `running=true` waits for `ContainerEngineReady=True`.
- `running=false` waits for the App's `Running` condition to leave `True` — stricter than "engine disconnected", because the VM must actually stop.
- Every other spec change returns as soon as the patch lands.

The wait filters condition snapshots by `observedGeneration >= post-write generation`, so a stale `ContainerEngineReady=True` from a previous reconcile or a prior backend cannot prematurely satisfy it. Both `EngineReconciler` and `AppReconciler` stamp `ObservedGeneration` on status updates.

## Discovery ready annotation

`rdd.rancherdesktop.io/ready=true` lands on the discovery ConfigMap once every enabled controller has installed its CRDs. Clients that depend on CRDs (`rdd svc start --wait`, `rdd set`) now wait for the annotation rather than ConfigMap existence alone, eliminating the race that surfaced as `the server could not find the requested resource`. The control plane sets the annotation inside the shared manager's startup goroutine after `installControllerCRDs` establishes every CRD, or immediately after `InitDiscovery` when no controllers are configured.

## App status multi-writer safety

`AppReconciler` now mirrors LimaVM conditions under `retry.RetryOnConflict` plus a re-`Get`, matching the pattern `EngineReconciler` uses for `ContainerEngineReady`. Both controllers write to `App.Status.Conditions`, and controller-runtime serializes reconciles within a controller but not across controllers; without the retry they 409-loop through requeues.

## Supporting changes

- `instance.DockerSocket()` returns the host-side Docker socket path (`~/.rdN/docker.sock`). The Lima template gains `HOST_DOCKER_SOCKET` and drops the raw `HOST_SHORT_DIR` parameter. `rdd svc paths docker_socket` exposes the path so BATS and other tooling can find it.
- `Volume` becomes selectable on `.status.name`, so field selectors can query the original Docker name after it has been hashed into an RFC 1123 subdomain (`vol-<sha256>`) for `metadata.name`.
- A new BATS suite `32-app-controllers/engine.bats` covers image pull/untag/dangling flows, container lifecycle, volumes with RFC-1123-invalid names, and bidirectional deletion. Skipped on Windows until the Docker socket access pattern is wired up for WSL2.
- Design-doc updates under `docs/design/`: `api_app.md` (new condition), `api_containers.md` (engine mirroring, terminology, spec.state, finalizer lifecycle), `cmd_app.md` (`--timeout`, wait semantics), `discovery.md` (ready annotation).

## Known limitations

- `containerEngine.name=containerd` reports `ContainerEngineReady` with reason `NotApplicable` and takes no mirroring action. The condition will be renamed when containerd mirroring lands.
- `reconcileContainerSpecs` and `processFinalizers` both list all resources per reconcile; splitting into per-resource reconcilers with watch predicates is the long-term fix.
- Docker socket access still needs wiring for Windows/WSL2.


docs/design/api_containers.md
```diff
@@ -26,6 +44,45 @@ unspecified.
...
+### Finalizer lifecycle
+
+Each mirror carries the `engine.rancherdesktop.io/docker-mirror`
+finalizer. A K8s-side delete triggers the finalizer handler, which
+deletes the corresponding engine object and then strips the finalizer
+so the mirror can be garbage-collected.
+
+An engine-side delete (for example, `docker rm`) goes the other way:
+the engine controller strips the finalizer and deletes the mirror
+directly, without calling back into the engine.
```
mook-as: Wouldn't a k8s-side delete trigger the finalizer, which deletes the engine object, and just lets that cascade via the engine-side-delete path? (In which case, describing the engine-side delete first might be easier.)

docs/design/api_containers.md
```diff
@@ -26,6 +44,45 @@ unspecified.
...
+rdd ctl wait --for=condition=ContainerEngineReady=True app/app
+```
+
+When the VM shuts down or the container engine becomes unreachable, the
+controller removes all mirror resources and sets `ContainerEngineReady` to
+`False`.
+
+### Finalizer lifecycle
+
+Each mirror carries the `engine.rancherdesktop.io/docker-mirror`
```
mook-as: Presumably we'd use a `engine.rancherdesktop.io/containerd-mirror` for the other mirror controller? Or should that be the same finalizer?

docs/design/api_containers.md
```diff
@@ -91,12 +148,23 @@ status:
...
+expressing intent, and the controller takes no start/stop action.
+
+To drive a container from the K8s API, set `spec.state` to `running`
+or `created`. The engine controller starts or stops the underlying
+container accordingly. `status.status` always reflects the actual
+Docker state.
 
-We will need to support a variety of actions on containers:
+Valid values for `spec.state`:
+- `unknown` — engine-managed; no action taken (default)
```
mook-as, line 163: Presumably, it's not legal to _set_ the state to `unknown`; it can only be reached when the container is initially created by the controller.

docs/design/cmd_app.md
```diff
@@ -2,15 +2,24 @@
...
 Set one or more properties on the App singleton resource. Properties use dot notation for nested fields.
 
 Valid property names and types are derived from the App CRD's OpenAPI schema at runtime, so the command automatically supports new properties as they are added to the CRD.
 
 If the App resource does not exist, it is created with default settings before the specified values are applied.
 
-- `--dry-run`: Validate changes against the API server's admission controller without persisting them. If the App does not exist, it is created with defaults (the VM will not start) so that the admission controller can validate the patch.
+By default, `rdd set` waits for the desired state before returning:
+
+- `running=true` waits for `ContainerEngineReady=True` — the engine watcher has connected to Docker on the `moby` backend, or the condition reports `NotApplicable` on `containerd`. If the controller set excludes engine (for example, `--controllers=-engine`, or on Windows where registration is skipped), the wait falls back to `Running=True`.
```
mook-as: Is all that based on the OpenAPI schema too, or is that specifically coded? I assume the latter (without reading the code); in that case, it is probably good to explicitly note that in the `By default, …` line:

> By default, `rdd set` waits for the desired state before returning when setting the following specific properties:

mook-as: I don't understand the second sentence (_… excludes engine_); I suspect that means something along the lines of:

> If the engine controller is not running (for example, `rdd` was started with `--controllers=-engine`), the wait falls back to `Running=true`.

I don't recall any registration being skipped on Windows; did I miss that?

docs/design/cmd_app.md
```diff
@@ -2,15 +2,24 @@
...
+- `running=false` waits for the App's `Running` condition to leave `True`, meaning the VM has actually stopped. This is stricter than "container engine disconnected".
+- Other property changes do not wait. Pure backend swaps (for example, `rdd set containerEngine.name=containerd` without a `running` argument) return as soon as the patch is accepted.
+
+The wait filters on the App's post-patch `metadata.generation`, so a leftover `ContainerEngineReady=True` from a prior backend cannot prematurely satisfy it.
+
+- `--dry-run`: Validate changes against the API server's admission controller without persisting them. If the App does not exist, it is created with defaults (the VM will not start) so the admission controller can validate the patch. The wait is skipped in dry-run mode.
```
mook-as, line 23: So `rdd set --dry-run` can actually cause changes on the server (specifically, creating the `app`)?  That's… really odd. But I see that the text was already there pre-change…

docs/design/cmd_app.md
```diff
@@ -2,15 +2,24 @@
...
+- `--timeout=DURATION`: How long to wait (default `5m`). `--timeout=0` skips the wait and returns as soon as the patch is accepted.
```
mook-as: Does the timeout expiring change the exit status? Even in the case of `--timeout=0`?

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
 
 ## Lifecycle
 
+The control plane creates the ConfigMap with an empty `data` map as soon as the API server is ready, before any controller manager starts. Its `creationTimestamp` therefore marks the control plane start time, which [`rdd ctl await --since=startup`](cmd_service.md) uses to filter condition transitions.
```
mook-as, line 39: Having both `rdd ctl wait` _and_ `rdd ctl await` sounds like a **terrible** idea.  But I guess that wasn't introduced in this PR… Filed #311 because it's just that bad.

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
...
+
 A controller manager patches the ConfigMap on startup (adding its API group key) and removes the key on shutdown. The ConfigMap itself is deleted when the last key is removed.
```
mook-as: So if I understand correctly, assuming we start RDD with all controllers disabled and add an external one:

1. Control plane creates `configmap/rdd-controller-manager`
2. Controller manager adds an entry on controller start
3. Controller manager removes the entry on controller stop
4. Controller manager finds the configmap empty, so it is completely deleted
5. Control plane is still running, everything breaks

Perhaps the controller manager should never delete the configmap; it should only be deleted by the control plane on shutdown?

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
...
 
-The control plane also cleans up the entire ConfigMap during service shutdown and on startup (to clear stale entries from a previous crash).
+The control plane also clears the ConfigMap on shutdown and on startup, to drop stale entries from a previous crash.
```
mook-as: Per the previous comment, then, the control plane should:

- On startup, clear the configmap (before registering any controllers), resetting the `creationTimestamp`.
- On shutdown, delete the configmap. Or just clear it (including the… `creationTimestamp`?), and never delete.  Delete sounds saner here.

jandubois:
Latest AI review is https://rancher-sandbox.github.io/rancher-desktop-daemon/20260415-103429-pr-309.html

I don't plan to address any of the findings in this PR unless convinced otherwise.

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
...
+
+## Ready annotation
+
+The annotation `rdd.rancherdesktop.io/ready=true` signals that every enabled controller has installed its CRDs **and** registered its data entry in the ConfigMap. Clients that depend on CRDs (for example, `rdd set`, which fetches the App CRD schema) or on the enabled-controller list (for example, `rdd set running=true`, which checks whether the engine controller is present) must wait for the annotation; otherwise they race startup and see either `the server could not find the requested resource` or a stale, empty controller list.
```
mook-as, line 45: What object does this annotation live on? `configmap/rdd-controller-manager`?

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
...
+
+The control plane sets the annotation in one of two places:
+
+1. Immediately after creating the ConfigMap, if no controllers are enabled.
```
mook-as, line 49: How does this interact with external controllers? That is, the control plane starts with all controllers disabled, then the user runs one externally the way we do the mock controller? I expect the answer is simply _the annotation does not include external controllers_, but that should be documented.

docs/design/discovery.md
```diff
@@ -36,9 +36,24 @@ The ConfigMap is named `rdd-controller-manager` in the `rdd-system` namespace. E
...
+2. Otherwise, inside the shared controller manager's startup goroutine, after `installControllerCRDs` has established every CRD **and** `registerDiscovery` has written the controller-manager entry into `configMap.Data`.
+
+Ordering both steps before the annotation keeps the "ready = clients may proceed" contract consistent: any client that waits for the annotation sees both CRDs and the enabled-controller list, not just one of the two.
+
+The ConfigMap is recreated on every control plane startup, so a stale `ready` from a previous crash is always cleared before CRD installation begins.
```
mook-as, line 54: There is technically a race between when the control plane starts and we delete the stale configmap, but I don't think that's likely to be an issue in practice.

mook-as (commented):
This reviews `docs/` only; I'll start reviewing actual code after this, but you can look at my comments first for concurrency.

bats/tests/32-app-controllers/app.bats
```diff
@@ -38,7 +38,7 @@ EOF
 }
 
 local_setup_file() {
-    setup_rdd_control_plane "app,limavm"
+    setup_rdd_control_plane "*"
```
mook-as:
```suggestion
    setup_rdd_control_plane
```
The `*` is the default, and it's actually _more_ confusing to include it.

bats/tests/32-app-controllers/engine-docker.bats
mook-as, line 1: This should be called `engine-docker.bats`, and we can have a second one later for `engine-containerd.bats`.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    # exercise — the engine controller is part of the default controller
+    # set, so no explicit --controllers selection is needed.
+    rdd svc delete
+    rdd set running=true
+}
+
+# --- Startup ---
+
+@test "ContainerNamespace moby exists" {
+    rdd ctl wait --for=create --namespace="${NAMESPACE}" \
```
mook-as: Shouldn't this use `RDD_NAMESPACE` instead, so that the helpers in `controller.bash` will work correctly? That might involve overriding it in `local_setup_file`, though.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+        containernamespace/moby --timeout=10s
```
mook-as:
```suggestion
        ContainerNamespace/moby --timeout=10s
```
Primarily for ease of (human) parsing, secondarily for the spell checker (though it's not an issue here).

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -0,0 +1,409 @@
...
+
+@test "ContainerNamespace moby exists" {
+    rdd ctl wait --for=create --namespace="${NAMESPACE}" \
+        containernamespace/moby --timeout=10s
+}
+
+# --- Image mirroring ---
+
+@test "docker pull creates Image resource" {
+    docker pull busybox
```
mook-as, line 41: TODO: avoid Docker Hub images due to rate limiting.

```suggestion
    docker build -t foo - <<-EOF
      FROM scratch
      ENTRYPOINT ["/"]
    EOF
```

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    # (the container still references it), so the UnTag path produces
+    # a dangling Image mirror rather than deleting the mirror outright.
+    # Tagging the dangling image then must prune the dangling mirror
+    # and leave only the new tagged one — the symmetric direction of
+    # the untag test above.
+    docker pull alpine:latest
+    run -0 docker inspect alpine:latest --format '{{.Id}}'
+    alpine_id=${output}
+
+    docker run -d --name dangling-pin alpine:latest sleep 3600
```
mook-as: You actually only need `docker create`, no need to actually run it.

Note that the `sleep 3600` _seems_ like it's adding a one hour time limit to the test, but it's not: the container is not removed after it exits.

In general, though, I think `sleep inf` more clearly indicates that you never expect that to end.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    docker rm --force dangling-pin
+    docker rmi dangling-tag-test:v1
+}
+
+# --- Container lifecycle mirroring ---
+
+@test "docker run creates Container resource with status=running" {
+    docker run -d --name test-lifecycle busybox sleep 3600
+
+    run -0 docker inspect test-lifecycle --format '{{.Id}}'
```
mook-as:
```suggestion
    run -0 docker run -d --name test-lifecycle busybox sleep inf
```
`docker run -d` already outputs the container id, no need to ignore the output _then_ ask for the container id.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    docker stop test-lifecycle
+    rdd ctl wait --for=jsonpath='{.status.status}'=exited \
+        --namespace="${NAMESPACE}" container/"${cid}" --timeout=30s
+}
+
+@test "docker rm deletes Container resource" {
+    run -0 docker inspect test-lifecycle --format '{{.Id}}'
+    cid=${output}
+
+    docker rm test-lifecycle
```
mook-as:
```suggestion
    docker rm --force test-lifecycle
```
Not strictly necessary, but this should allow this test to pass if the previous one failed to stop it for some reason.

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -0,0 +1,409 @@
...
+        --field-selector "status.id=${alpine_id}" \
+        -o jsonpath='{.items[*].status.repoTag}'
+    assert_output "dangling-tag-test:v1"
+
+    # Cleanup.
+    docker rm --force dangling-pin
+    docker rmi dangling-tag-test:v1
+}
+
+# --- Container lifecycle mirroring ---
```
mook-as, line 122: Actually, I wonder if image/container/volume should be in different test files… Mainly to make the files shorter and easier to understand (because there wouldn't be state left over from previous sections). I realize that might make things take slightly longer due to new VM spinups, though, so it would need to be a balance…

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    rdd ctl wait --for=delete --namespace="${NAMESPACE}" volume \
+        --field-selector "status.name=My_Vol_Ume" --timeout=30s
+}
+
+# --- Deletion via the API removes the Docker object ---
+
+@test "deleting Container resource removes Docker container" {
+    docker run -d --name test-delete busybox sleep 3600
+
+    run -0 docker inspect test-delete --format '{{.Id}}'
```
mook-as:
```suggestion
    run -0 docker run -d --name test-delete busybox sleep 3600
```

Same as before, `docker run -d` outputs the container id. I guess I should stop mentioning already-mentioned issues (like the `sleep 3600` vs `sleep inf`)…

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+        --namespace="${NAMESPACE}" container/"${cid}" --timeout=30s
+
+    # Resolve the Image mirror name from its repoTag.
+    run -0 rdd ctl get image --namespace="${NAMESPACE}" \
+        --field-selector "status.repoTag=busybox:latest" -o name
+    assert_output
+    image_ref=${output}
+
+    # Docker will refuse to remove an image referenced by a running
+    # container. With I3 fixed, processImageFinalizers leaves the
```
mook-as: What is I3? It sounds like one of your AI-review points, in which case referring to the review comment in code isn't appropriate (because the context is totally gone by the time people read the code). In fact, most of this comment (the whole second sentence) sounds like it shouldn't live in the test.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    # finalizer in place and the Image mirror stays (in Terminating
+    # state) until the image is actually removable.
+    rdd ctl delete "${image_ref}" --namespace="${NAMESPACE}" --wait=false
+
+    run -0 rdd ctl get "${image_ref}" --namespace="${NAMESPACE}" \
```
mook-as: Is there a race here? Should this be `rdd ctl wait --force=jsonpath=…` instead?

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+        -o jsonpath='{.metadata.deletionTimestamp}'
+    assert_output
+    run -0 rdd ctl get "${image_ref}" --namespace="${NAMESPACE}" \
+        -o jsonpath='{.metadata.finalizers[0]}'
+    assert_output "engine.rancherdesktop.io/docker-mirror"
```
mook-as: Am I understanding correctly then this is more like `docker rmi` instead of `docker rmi --force`, in that the image does _not_ get removed? Except it does again, after all users are gone?

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -0,0 +1,409 @@
...
+        --type=merge -p '{"spec":{"state":"running"}}'
+
+    rdd ctl wait --for=jsonpath='{.status.status}'=running \
+        --namespace="${NAMESPACE}" container/"${cid}" --timeout=30s
+
+    run -0 docker inspect test-state --format '{{.State.Status}}'
+    assert_output "running"
+}
+
+@test "patching spec.state=created stops a paused container" {
```
mook-as, line 277: This whole test feels weird, mainly because we don't really have a paused state. We're poking into the container spec to set it to a state that shouldn't be user-settable to test this.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+
+    rdd ctl wait --for=jsonpath='{.status.status}'=running \
+        --namespace="${NAMESPACE}" container/"${cid}" --timeout=30s
+
+    docker pause test-unpause
+    rdd ctl wait --for=jsonpath='{.status.status}'=paused \
+        --namespace="${NAMESPACE}" container/"${cid}" --timeout=30s
+
+    rdd ctl patch container "${cid}" --namespace="${NAMESPACE}" \
+        --type=merge -p '{"spec":{"state":"running"}}'
```
mook-as: We should probably assert the state before the patch, too, so we know exactly what we're changing _from_.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+    rdd set --timeout=10s running=false
+}
+
+@test "rdd set --timeout=0 skips the wait entirely" {
+    # --timeout=0 preserves the pre-wait-by-default behavior: the
+    # command returns as soon as the patch is accepted, regardless of
+    # whether ContainerEngineReady has caught up. Use a no-op patch
+    # (running=false on an already-stopped VM) so the test does not
+    # race VM startup.
+    rdd set --timeout=0 running=false
```
mook-as: I don't understand what the test actually does, other than checking that `--timeout=0` doesn't throw an error: it doesn't actually check that we _didn't_ wait. I'm not actually sure we can.

docs/design/api_containers.md
jandubois: The whole `spec.state` will be removed soon, so let's not worry about it.

bats/tests/32-app-controllers/engine.bats
```diff
@@ -0,0 +1,409 @@
...
+}
+
+@test "restarting VM restores ContainerEngineReady and moby namespace" {
```
mook-as: It's not clear if checking that it creates the moby namespace because that's implicit, or if it restores it because it used to exist.  This is not a very clear test, and I'm not sure it needs to exist.

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -0,0 +1,409 @@
...
+
+@test "deleting containernamespace/moby completes without a finalizer hang" {
+    # moby ContainerNamespace has no mirror finalizer, so a user delete
+    # must return promptly rather than get trapped in Terminating.
+    rdd ctl delete containernamespace/moby --namespace="${NAMESPACE}" --timeout=10s
+    run -0 rdd ctl get containernamespaces --namespace="${NAMESPACE}" --output=name
+    refute_output
+}
+
+# --- containerd backend ---
```
mook-as, line 381: As noted before, this should live in a separate file; we will need to test the containerd backend with the whole set once it's implemented.

cmd/rdd/limavm_shell.go
```diff
@@ -172,8 +172,15 @@ func limaVMShellAction(cmd *cobra.Command, args []string) error {
...
 
+	// ConnectTimeout caps the TCP handshake at 30s. ServerAliveInterval=30
+	// with ServerAliveCountMax=3 closes a wedged session after ~90s of
+	// unanswered keepalives. Interactive shells and long-running commands
+	// ack the keepalives and stay connected.
 	sshArgs = append(sshArgs, []string{
 		"-o", fmt.Sprintf("LogLevel=%s", logLevel),
+		"-o", "ConnectTimeout=30",
+		"-o", "ServerAliveInterval=30",
+		"-o", "ServerAliveCountMax=3",
```
mook-as: This feels like it should live in a different PR (that can get merged first).

mook-as (commented):
This round is for BAT tests.

Except I forgot to stop and added an extra unrelated comment.  Appropriately, it's about a hunk that feels like it should be in a separate PR, for a comment that should be in a separate round of comments…

docs/design/cmd_app.md
jandubois: I don't understand what you are asking about here? All the properties come from the schema, but the wait condition is hardcoded and will be switched to "Settled" in a followup PR. The exceptions for `containerd` are temporary and will be removed when the Engine controller gets `containerd` support.

jandubois: This is trying to comment on the fact that the Engine controller is part of the Containers APIGroup. So if you want to use the Mock controller you have to use `--controllers=containerd,-engine` because you need the `Container` etc controller (and CRDs), but the must disable Engine to use Mock. Otherwise these 2 would fight over who is the authoritative source of the Container/Image/Volume resources.

docs/design/cmd_app.md
jandubois: The `--dry-run` is restricted to the setting change attempt. You must have settings in the first place to validate a change.

It is an edge case that should not really happen much for actual users? Why would you want to do a dry run of settings changes if you don't even have an app? That's not really a meaningful use case.

docs/design/cmd_app.md
jandubois: I have to double-check. My intention was that with `--timeout=0` the exit status is 0 if the admission controller has accepted the change. With a non-zero timeout the exit code would not be 0 unless the wait was successful.

We should probably define different exit codes for "request failed" vs. "wait timed out". This applies to all other commands that have a `--wait` option as well.

I just remember that in `kubectl wait --timeout=0` means wait indefinitely, so this was probably a poor choice, and we should have a `--wait=false` option instead (which is what we already do for the various `rdd svc` subcommand. I'll add this as a comment to #196.

docs/design/discovery.md
jandubois: Agreed, but that is not new in this PR. I've created #312 to fix it.

docs/design/discovery.md
jandubois: Not sure; what problem does this solve? Then we need a different mechanism to determine when the control plane has started.

docs/design/discovery.md
jandubois: On the discovery ConfigMap

docs/design/discovery.md
jandubois: It cannot include external controller. You could attach additional controllers at any time. But yeah, the docs could mention that.

pkg/instance/instance.go
```diff
@@ -119,3 +119,10 @@ var ShortDir = sync.OnceValue(func() string {
 var LimaHome = sync.OnceValue(func() string {
 	return filepath.Join(ShortDir(), "lima")
 })
+
+// DockerSocket returns the path to the Docker socket for this instance
+// (e.g., ~/.rd2/docker.sock). This is the host-side socket that Lima
+// port-forwards from the guest's /var/run/docker.sock.
+var DockerSocket = sync.OnceValue(func() string {
+	return filepath.Join(ShortDir(), "docker.sock")
```
mook-as, line 127: Not relevant for this PR: this needs to be fixed for Windows, because that's `\\.\pipe\docker_engine` or whatever.

docs/design/cmd_app.md
mook-as: Okay, if _the wait condition is hardcoded_ then the documentation should say that; currently it's ambiguous if it's somehow derived from the CRDs.  It's just unclear (because I'm reading the docs _before_ reading the code).

mook-as: Makes sense; I just needed to know that `--controllers=-engine` was an argument to `rdd`, because it wasn't specified, so the context defaulted to `rdd app --controllers=-engine`…

Also I still don't understand the Windows bit.

docs/design/cmd_app.md
mook-as: Right, hence my first thought would be that `rdd set --dry-run` when the app doesn't exist should fail instead of modifying the backend in any way.  But that's kind of trivial and we can just punt on it.

docs/design/api_containers.md
jandubois: That feels more complicated. It adds further latency for the resource to hang around in deleted state, waiting for the watch to catch the delete event and only then remove the finalizer. There isn't anything you can do with the resource once it is in deleted state, so why no let the deletion continue?

docs/design/discovery.md
mook-as: What I mean is, after the (external) controllers go away, the config map ends up missing even while the control plane is still running (and therefore it is not longer possible to determine when the control plane has started).

docs/design/discovery.md
mook-as: Okay, please update the documentation to include that. (I assume it's just queued on your local disk before you push an update once I'm done the review.)

docs/design/discovery.md
jandubois: Yes, there is a race, but it is almost impossible to prevent. We would need e.g. a private endpoint to connect to the api server, and then connect a public listener to it once the initialization is done. But that is a lot of overhead to close a millisecond inconsistency after a dirty shutdown/crash. Not worth it IMHO.

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+
+// waitForDesiredState waits for the App to reach the state
+// implied by the properties that were set. running=true waits
+// for ContainerEngineReady=True when the engine controller is
+// enabled, or Running=True otherwise. running=false waits for
+// the App's Running condition to go False — stricter than
+// "container engine disconnected" because the VM must actually
+// stop. Other property changes do not wait.
+//
+// minGen is the App's metadata.generation after our write. The
```
mook-as: Since this function isn't doing the writing, the whole first sentence is just confusing. Just the second one (_The predicate accepts only…_) is clear enough, I think.

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+// >= minGen, so a stale ContainerEngineReady=True from a previous
+// backend cannot satisfy the wait before the reconciler has observed
+// the new spec. Both EngineReconciler and AppReconciler stamp
+// ObservedGeneration on update.
+//
+// TODO: changes that trigger a VM restart without changing "running"
+// (for example, containerEngine.name alone) are still not waited on.
+// A dedicated "Reconciled" condition on App would let the CLI detect
+// when the full reconcile chain has settled for any property change.
+func waitForDesiredState(ctx context.Context, config *rest.Config, properties map[string]string, timeout time.Duration, minGen int64) error {
```
mook-as, line 490: Hmm, this only examines the properties given, right? So potentially this can get wedged (up to the timeout) if:

- User runs `rdd set ..things...` (e.g. enable container engine)
- We wait for the state implied by the properties
- In the background, something else updates the app, modifying the properties that are set (e.g. disable container engine)
- Since the desired state conflicts with the passed in properties, it never passes.

To rephrase: would it make more sense to watch the app object for its state to match its spec?  As noted in the TODO above, it would just eventually transition to a `Reconciled` condition, and this function just ends up checking that `Reconciled` is `True` where its `ObservedGenerate` is at least `minGen`.

Of course, maybe getting wedge is a _feature_, in the sense that `rdd set running=true` will fail if somebody sets `running=false` in the background.  In that case probably just note it in a comment and call it a day.  (When we do the Reconciled condition, we'd need to factor out the state-checking logic so we can wait for Reconciled, and re-run the same logic the reconciler does and report if the reconciled state does not match the desired state at the point we entered this function.  But in that case we'd probably get the app state as an input, instead of a random property map.)

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+	enabled, err := discovery.GetEnabledControllers(ctx)
+	if err != nil {
+		return false, fmt.Errorf("failed to list enabled controllers: %w", err)
+	}
+	for _, name := range enabled {
+		if name == engineControllerName {
+			return true, nil
+		}
+	}
+	return false, nil
```
mook-as:
```suggestion
	return slices.Contains(enabled, engineControllerName), nil
```

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+}
+
+// watchCondition watches the App singleton until the predicate
+// returns true or the context expires. watchtools.UntilWithSync
+// re-lists transparently on benign watch-channel closures (API
+// timeouts, proxy disconnects, resource-version compaction).
+func watchCondition(ctx context.Context, config *rest.Config, satisfied func(*unstructured.Unstructured) bool) error {
+	dynClient, err := dynamic.NewForConfig(config)
```
mook-as, line 558: No need to put this into the code (just a comment on the PR is fine), but why do we need to use the dynamic client here instead of using `pkg/apis/app/…`?

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+	if err != nil {
+		return fmt.Errorf("failed to create dynamic client: %w", err)
+	}
+
+	lw := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			options.FieldSelector = "metadata.name=app"
```
mook-as, line 566: If the app is a singleton, is the field selector required? (I guess the API might require it and I just didn't know?)

cmd/rdd/set.go
```diff
@@ -395,17 +440,176 @@ func patchApp(ctx context.Context, c client.Client, app *appv1alpha1.App, specMa
...
+		return satisfied(obj), nil
+	}
+
+	if _, err := watchtools.UntilWithSync(ctx, lw, &unstructured.Unstructured{}, precondition, condition); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			return errors.New("timed out waiting for App state")
+		}
+		if errors.Is(err, context.Canceled) {
+			return errors.New("wait cancelled")
+		}
```
mook-as, line 598: Any particular reason we're _not_ wrapping the underlying errors here (so we can't `errors.Is` in the caller)? In fact, just passing these two errors as-is is probably fine?

bats/tests/32-app-controllers/engine-docker.bats
jandubois: The main reason I don't want to do this is that it means spinning up a separate VM, and each VM start adds several minutes to the total BATS runtime. Especially on Windows this could add 7-10 minutes. So I don't want to do this unless we really run into a problem here.

bats/tests/32-app-controllers/engine.bats
jandubois: Yes, that is how it works. Do you think this is not correct?

bats/tests/32-app-controllers/engine-docker.bats
jandubois: Yes, it is weird. Which is one of the many reasons I want to get rid of `spec.state`, at which point this will be rewritten anyways.

bats/tests/32-app-controllers/engine.bats
mook-as: No, just confirming my understanding. The user might get confused if the point between the removal request and the actual removal takes a long time (say, an hour), but that will probably be fine if we end up showing the state in the UI as "deleting" or "pending delete". (It'll not be the state in `.status.status`; maybe we can stick it into a status condition, or just calculate that in the frontend.)

bats/tests/32-app-controllers/engine-docker.bats
jandubois: Will move to a separate file when we add containerd support. For now I've renamed the other file.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -298,28 +299,37 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
-			log.Error(err, "Unable to update App status")
-			return ctrl.Result{}, err
+	// Mirror LimaVM status conditions onto the App status. The engine
+	// reconciler writes ContainerEngineReady on the same object, so
+	// app's resourceVersion from the initial Get can be stale.
+	// retry.RetryOnConflict + re-Get matches
+	// EngineReconciler.setEngineCondition; without it, concurrent
+	// writers 409-loop through controller-runtime requeues. Truncate
+	// messages defensively against a future bypass of the LimaVM
+	// controller's source-side truncation.
```
mook-as: The last sentence about truncating was more understandable in the old version, I think. Maybe just copy it over, if it's still correct?

It may also be useful to move that to the `TruncatedConditionMessage` call instead. (Or do the truncation and put it in a temporary variable, so we can have the comment there.)

pkg/external/main.go
```diff
@@ -60,6 +60,19 @@ func RunControllers(apiGroupName string) int {
 	var controllersToStart []base.Controller
 
 	for _, controller := range base.GetAllControllers() {
+		// A mistaken blank import can pull a foreign-group controller
+		// into the global registry. Filter by API group so this binary
+		// only manages controllers it is named for; a stray import
+		// cannot cross-register into the wrong shared manager. Log
+		// and skip rather than starting under the wrong group: this
+		// is a build-time bug in our own code, not a user error.
```
mook-as: That sounds like a good reason to `panic()` rather than just log; that way things will get fixed faster? (That is, it prevents us shipping with the mistake.)

pkg/service/controllers/discovery.go
```diff
@@ -411,6 +421,39 @@ func InitDiscovery(ctx context.Context, client kubernetes.Interface) error {
...
+		ctx,
+		ControllerManagerConfigMapName,
+		types.StrategicMergePatchType,
+		patchData,
+		metav1.PatchOptions{},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to mark control plane as ready: %w", err)
+	}
+	klog.InfoS("Marked control plane as ready",
```
mook-as: Not `klog.FromContext(ctx).Info(…)`? Not that it really matters, of course.

docs/design/discovery.md
jandubois: Sorry, you were right; the ConfigMap should not be deleted because this could be triggered by an external controller. Fixed.

docs/design/discovery.md
jandubois: Added to the docs

bats/tests/32-app-controllers/engine-docker.bats
jandubois: I want to ignore this for now. Did you run into rate limits?

bats/tests/32-app-controllers/engine.bats
jandubois: I've removed the test. I've also changed it so `--timeout=0` waits indefinitely, like `kubectl wait --timeout=0`, and defined a range of exit codes for different exit reasons. No test for `--wait=false` though

pkg/service/controllers/discovery_test.go
```diff
@@ -185,3 +185,56 @@ func TestControllerManagerDiscoveryGroup(t *testing.T) {
...
+	assert.NilError(t, err, "failed to create ControllerManagerDiscoveryGroup")
+	assert.NilError(t, d.RegisterControllerManager(t.Context(), ControllerManagerInput{
+		HealthPort: 1234,
+	}), "failed to register controller manager")
+
+	cm, err = client.CoreV1().ConfigMaps(RDDSystemNamespace).Get(t.Context(), ControllerManagerConfigMapName, metav1.GetOptions{})
+	assert.NilError(t, err, "failed to get discovery config map after register")
+	assert.Equal(t, cm.Annotations[ReadyAnnotation], "true", "expected ready annotation to survive controller manager registration")
+
+	// InitDiscovery on a second startup clears the stale ready annotation.
```
mook-as: Should that also update the creation timestamp?

mook-as (commented):
That should be everything except `pkg/controllers/app/engine/…`

bats/tests/32-app-controllers/engine-docker.bats
mook-as: Nope, and yeah it's totally fine to ignore any comments (until I find something that's really a blocker, which I don't expect).

bats/tests/32-app-controllers/engine.bats
mook-as: I haven't looked at your changes yet, but I found that there was a https://pkg.go.dev/k8s.io/client-go/tools/watch#ContextWithOptionalTimeout that handles `--wait` durations.  Yeah, I have no idea why it lives _there_ of all places.

pkg/controllers/app/engine/controller.go
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Package engine registers the engine controller. The engine controller
```
mook-as, line 5: This text will need to (eventually, doesn't have to be now) be rewritten to be not docker-specific; probably around the same time we implement containerd reconciliation.

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,451 @@
...
+	cli, err := dockerclient.New(
+		dockerclient.WithHost("unix://" + socketPath),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Docker client: %w", err)
+	}
+
+	// Verify the connection by pinging Docker.
+	pingCtx, pingCancel := context.WithTimeout(ctx, 5*time.Second)
+	defer pingCancel()
```
mook-as, line 59: Should this be in an IIFE so we don't run `pingCancel` way later? I guess it doesn't matter _that_ much, since we call `w.run(…)` in a goroutine…

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,451 @@
...
+	case <-w.done:
+		return false
+	default:
+		return true
+	}
+}
+
+// run is the main watcher goroutine. since is the Docker events
+// "Since" filter captured before fullSync.
+func (w *dockerWatcher) run(ctx context.Context, since string) {
```
mook-as, line 144: API comment: it might be clearer to have `since` be a `time.Time`, and we can do the conversion in here, as `Since: strconv.FormatInt(since.Unix(), 10)`.

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,451 @@
...
+		Filters: eventFilter,
+	})
+
+	for {
+		select {
+		case <-ctx.Done():
+			log.Info("Docker watcher stopping")
+			return
+		case err := <-result.Err:
+			if ctx.Err() != nil {
```
mook-as, line 184: The docs for `ctx.Err()` says this is only non-nil if `ctx.Done()` is closed? I guess there can be a race, so this is fine… do we need a `log.Info` in this case?

pkg/instance/instance.go
jandubois: Correct. Just setting it to the different string isn't enough though, we also need to implement the named pipe (or copy the code from Rancher Desktop 1.x).

cmd/rdd/set.go
jandubois: This is where the new `Settled` condition comes in. `rdd set` will wait for `Settled` to become true, which means there is nothing left for the reconciler to do. If `running` is true, then we may want to wait for `Settled` AND `Ready` to make sure the container engine and/or kubernetes are also usable.

cmd/rdd/set.go
jandubois: There is no strong reason for using it beyond that we already have a `client.Client` and therefore don't need to add all the scaffolding.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,568 @@
...
+	appName = "app"
+
+	// containerNamespace is the Docker container namespace name;
+	// Docker uses a single namespace called "moby".
+	containerNamespace = "moby"
+
+	// apiNamespace is the Kubernetes namespace for mirror resources.
+	apiNamespace = "rancher-desktop"
+
+	// controllerName is the SSA field owner for engine-controller applies.
```
mook-as, line 41: Took me a while to figure out SSA is "server-side apply". I think that's just me being dumb, though.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,568 @@
...
+	// required-field validation.
+	finalizerFieldOwner = controllerName + "-finalizer"
+
+	// mirrorFinalizer is added to mirror resources so user deletions
+	// are forwarded to Docker before the resource is removed.
+	mirrorFinalizer = "engine.rancherdesktop.io/docker-mirror"
+
+	// conditionContainerEngineReady goes to True once the engine
+	// controller has connected to Docker and finished the initial sync.
+	conditionContainerEngineReady = "ContainerEngineReady"
```
mook-as: It should be mentioned that this is a status condition on App objects.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,568 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	appv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+	containersv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/containers/v1alpha1"
+)
+
+const (
+	appName = "app"
```
mook-as, line 35: Hmm, I think we should have a constant in `app/v1alpha1` instead of it just randomly living everywhere.

pkg/controllers/app/app/controllers/app_controller.go
mook-as, line 1: Not for this PR, but I just realized that we don't have a validator to ensure we only create the `app` singleton, do we?

cmd/rdd/set.go
mook-as: … but we didn't pass in the `client.Client`, and we make a new one here? Meh, whatever, we can think about changing it later.

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,451 @@
...
+	// fullSync is idempotent, so replaying a few extra minutes is safe.
+	since, err := dockerEventsSince(ctx, cli)
+	if err != nil {
+		log := logf.FromContext(ctx).WithName("docker-watcher")
+		log.V(1).Info("Failed to query Docker daemon time, falling back to biased host clock",
+			"error", err)
+		since = strconv.FormatInt(time.Now().Add(-2*time.Minute).Unix(), 10)
+	}
+
+	if err := w.fullSync(watchCtx); err != nil {
```
mook-as, line 90: At this point, `watchCancel` cannot have been called yet (the first opportunity is when this returns with an error), so `watchCtx` is the same as `ctx`.  It's just confusing, but I guess harmless?

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,451 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	containersv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/containers/v1alpha1"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
+)
+
+// dockerWatcher manages a Docker client connection and event stream. It
+// performs a full sync on connect and then watches for incremental changes.
+type dockerWatcher struct {
+	cli *dockerclient.Client
```
mook-as, line 33: I think things would be clearer if this is named `moby` instead of `cli`.

jandubois:
@mook-as I believe I have addressed all review feedback.

The branch has been rebased on latest `main` to get the updated GHA workflow.

Review feedback is in individual commits, with smaller changes collected in a single additional commit.

Let me know if I missed anything!

pkg/controllers/app/engine/controllers/sync_volumes.go
```diff
@@ -0,0 +1,142 @@
...
+	sum := sha256.Sum256([]byte(dockerName))
+	return fmt.Sprintf("vol-%x", sum)
+}
+
+// syncContainerNamespace creates the "moby" ContainerNamespace mirror.
+// This resource has no mirror finalizer: Docker has no corresponding
+// engine object for the reverse delete, and cleanupMirrorResources
+// sweeps it unconditionally on VM stop, so a finalizer with no
+// handler would only trap user deletes in Terminating.
+func (w *dockerWatcher) syncContainerNamespace(ctx context.Context) error {
```
mook-as: Seems odd that this lives _here_; maybe a new `sync.go` instead? `sync_misc`?

pkg/controllers/app/engine/controllers/sync_volumes.go
```diff
@@ -0,0 +1,142 @@
...
+	containersv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/containers/v1alpha1"
+	containersv1alpha1apply "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/containers/v1alpha1/applyconfiguration/containers/v1alpha1"
+)
+
+// volumeMirrorName returns a deterministic RFC 1123 subdomain name
+// for a Docker volume. Docker permits uppercase and underscores,
+// which are invalid in K8s object names, so the Docker name is
+// hashed with a "vol-" prefix. The original is preserved in
+// status.name.
+func volumeMirrorName(dockerName string) string {
```
mook-as, line 31: That makes is harder to pick out which is which, but we can figure that out later.

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,452 @@
...
+
+// removeMirrorResource strips the finalizer from a mirror resource and
+// deletes it, used when Docker has already deleted the underlying
+// object. The Update is wrapped in retry.RetryOnConflict to survive a
+// stale cache; NotFound is treated as success. obj is a template —
+// each retry iteration starts from a fresh DeepCopyObject.
+func (w *dockerWatcher) removeMirrorResource(ctx context.Context, obj client.Object, name string) error {
+	key := client.ObjectKey{Name: name, Namespace: w.apiNamespace}
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		latest := obj.DeepCopyObject().(client.Object)
```
mook-as: Do we need to deep copy that many times? Wouldn't doing it once at the start of the function be enough, since it'll just keep getting overwritten? We also wouldn't need to make a new copy for `deleteTarget`, since that would already have the namespace/name set (when fetching).

pkg/controllers/app/engine/controllers/sync_containers.go
```diff
@@ -0,0 +1,266 @@
...
+	// Apply succeeds on a fresh resource and fails loudly on a
+	// concurrent conflict.
+	var existing containersv1alpha1.Container
+	err := w.k8s.Get(ctx, client.ObjectKey{Name: inspect.ID, Namespace: w.apiNamespace}, &existing)
+	if apierrors.IsNotFound(err) {
+		applyConfig := containersv1alpha1apply.Container(inspect.ID, w.apiNamespace).
+			WithFinalizers(mirrorFinalizer).
+			WithSpec(containersv1alpha1apply.ContainerSpec().WithState(containersv1alpha1.ContainerStatusUnknown))
+		if err := w.k8s.Apply(ctx, applyConfig, client.FieldOwner(controllerName)); err != nil {
+			return fmt.Errorf("failed to create container %s: %w", inspect.ID, err)
```
mook-as, line 126: On failure to apply, that means we fail if a container is created between lines 120 and 125; shouldn't we just continue to updating the status in that case?

pkg/controllers/app/engine/controllers/sync_containers.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+	} else if err != nil {
+		return fmt.Errorf("failed to get container %s: %w", inspect.ID, err)
+	} else if existing.DeletionTimestamp == nil {
+		// Re-assert the mirror finalizer on every sync so a user who
+		// `kubectl edit`s it away cannot bypass the engine-side Docker
+		// cleanup on a later delete. Skip re-assertion once the mirror
+		// is Terminating: adding a finalizer to a deleting object is
+		// rejected, and processContainerFinalizers is about to strip
+		// the finalizer anyway.
```
mook-as, line 136: That _Sounds_ like it would be a thing for validators, but I can see that causing other issues…

pkg/controllers/app/engine/controllers/sync_containers.go
```diff
@@ -0,0 +1,266 @@
...
+			// portNames is sorted: ContainerPort.Bindings is atomic
+			// under SSA, and Docker returns dual-stack bindings in
+			// either order.
+			sortedPorts := slices.Clone(ports)
+			sort.Slice(sortedPorts, func(i, j int) bool {
+				if sortedPorts[i].HostIP.String() != sortedPorts[j].HostIP.String() {
+					return sortedPorts[i].HostIP.String() < sortedPorts[j].HostIP.String()
+				}
+				return sortedPorts[i].HostPort < sortedPorts[j].HostPort
+			})
```
mook-as:
```suggestion
			sortedPorts := slices.SortedFunc(slices.Values(ports), func(pb1, pb2 mobynetwork.PortBinding) int {
				if pb1.HostIP.String() != pb2.HostIP.String() {
					return strings.Compare(pb1.HostIP.String(), pb2.HostIP.String())
				}
				return strings.Compare(pb1.HostPort, pb2.HostPort)
			})
```

pkg/controllers/app/engine/controllers/sync_containers.go
```diff
@@ -0,0 +1,266 @@
...
+			// Sort bindings by (HostIP, HostPort) for the same reason
+			// portNames is sorted: ContainerPort.Bindings is atomic
+			// under SSA, and Docker returns dual-stack bindings in
+			// either order.
+			sortedPorts := slices.Clone(ports)
+			sort.Slice(sortedPorts, func(i, j int) bool {
+				if sortedPorts[i].HostIP.String() != sortedPorts[j].HostIP.String() {
+					return sortedPorts[i].HostIP.String() < sortedPorts[j].HostIP.String()
+				}
+				return sortedPorts[i].HostPort < sortedPorts[j].HostPort
```
mook-as: Note that `HostPort` is a string here, but I don't think that's relevant because it's still stable…

pkg/controllers/app/engine/controllers/sync_containers.go
```diff
@@ -0,0 +1,266 @@
...
+				WithName(portName.String()).
+				WithBindings(bindings...))
+		}
+	}
+	statusApply.WithPorts(applyPorts...)
+
+	statusConfig := containersv1alpha1apply.Container(inspect.ID, w.apiNamespace).
+		WithStatus(statusApply)
+
+	err = w.k8s.SubResource("status").Apply(ctx, statusConfig,
```
mook-as:
```suggestion
	err = w.k8s.Status().Apply(ctx, statusConfig,
```
It's the exact same thing, but guards against typos.

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,452 @@
...
+		errs = append(errs, fmt.Errorf("failed to sync containers: %w", err))
+	}
+	if err := w.syncAllImages(ctx); err != nil {
+		errs = append(errs, fmt.Errorf("failed to sync images: %w", err))
+	}
+	if err := w.syncAllVolumes(ctx); err != nil {
+		errs = append(errs, fmt.Errorf("failed to sync volumes: %w", err))
+	}
+
+	log.Info("Full sync complete")
```
mook-as: Should this also log at least the number of errors, if it's not zero?

pkg/controllers/app/engine/controllers/sync_volumes.go
```diff
@@ -0,0 +1,142 @@
...
+		WithLabels(vol.Labels).
+		WithOptions(vol.Options).
+		WithMountPoint(vol.Mountpoint).
+		WithScope(vol.Scope)
+
+	if t, err := time.Parse(time.RFC3339Nano, vol.CreatedAt); err == nil {
+		statusApply.WithCreatedAt(metav1.NewTime(t))
+	}
+
+	err = w.k8s.SubResource("status").Apply(ctx,
```
mook-as:
```suggestion
	err = w.k8s.Status().Apply(ctx,
```

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,452 @@
...
+				log.Info("Docker event stream closed")
+				return
+			}
+			// Transient handleEvent errors (API 503, SSA conflict past
+			// its internal retry) are logged and dropped. Container
+			// events self-heal on the next state change; image pull and
+			// volume create events fire once, so a dropped apply leaves
+			// the mirror missing until the next full resync. A periodic
+			// fullSync tick is deferred until the rest of the mirroring
+			// work lands.
```
mook-as: If I understand that last sentence correctly, we expect to implement a periodic full sync as a TODO? That sounds reasonable, but we should actually mark it as a TODO (that we can grep for) / file an issue for it.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+	if watcherRunning && !r.watcher.alive() {
+		diedWatcher = r.watcher
+		r.watcher = nil
+		watcherRunning = false
+	}
+	r.watcherMu.Unlock()
+	watcherDied := diedWatcher != nil
+	if watcherDied {
+		diedWatcher.stop()
+	}
```
mook-as: `watcherDied` is true if `diedWatcher` is not `nil`; that can only be the case if the chunk at line 112 run, so `watcherRunner && !r.watcher.alive()`.  That means `r.watcher.alive()` must have returned `false`, which must mean `w.done` was closed.  That can only happen if `w.run()`'s `defer close(w.done)` was run, which can be true if the context for `w.run()` was closed (i.e. `w.cancel()` had been called), or an error was reported from dockerd's event stream, or that stream was closed.  So I guess this exists to catch the latter two cases? That and to ensure the docker client was closed…

I guess this still makes sense (for cleanup)? It's just rather convoluted.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+
+	// Treat a dead watcher as a transient disconnect and fall through.
+	// If wantWatcher is still true below, a fresh watcher's fullSync
+	// reconciles drift in place — existing mirror resources keep their
+	// identity, so downstream clients see no churn. An actual stop or
+	// backend change is handled by the !wantWatcher branch, which
+	// sweeps the mirrors.
+	if watcherDied {
+		log.Info("Docker watcher died, will attempt to reconnect")
+	}
```
mook-as, line 128: Should this be in the same block as the previous statement? It's odd to have the comment after the initial reaction to the condition…

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+func (r *EngineReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Result, error) {
+	log := logf.FromContext(ctx)
+
+	var app appv1alpha1.App
+	if err := r.Get(ctx, client.ObjectKey{Name: appName}, &app); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	r.apiNamespace = app.GetResourceNamespace()
+
+	running := meta.IsStatusConditionTrue(app.Status.Conditions, "Running")
```
mook-as: I think we should list at least some of the known conditions as consts in `app_types.go`, so it's easier to find places that reference them.

It's not clear from `api_app.md` whether the `Running=True` state is true when the VM is alive, or when the VM has started the container engine, or when we have forwarded the container engine socket, or what. We may need more granular states.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+	return errors.Join(errs...)
+}
+
+// deleteAllOfType lists and removes every resource of the given kind
+// in r.apiNamespace. Engine is authoritative for Container, Image, and
+// Volume in the App's namespace: this loop deletes every object, not
+// just engine-owned mirrors. Coexistence with another writer to these
+// kinds requires an engine-owned label and matching filters here and
+// in the sync_*.go full-sync prune paths. Finalizer removal retries
```
mook-as, line 312: Hmm, we will eventually (post-PR) figure out how to coexist with the mock controller.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+			if err := r.cleanupMirrorResources(ctx); err != nil {
+				log.Error(err, "Failed to clean up mirror resources")
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, r.setEngineCondition(ctx, &app, terminalStatus, terminalReason, terminalMessage)
+	}
+
+	if !watcherRunning {
+		log.Info("App is running, starting Docker watcher")
```
mook-as, line 169: Is this supposed to be reentrancy-safe? There's a lock and stuff, but here `watcherRunning` may be outdated.

Maybe we just need extra comments around `EngineReconciler` to document if we allow multiple reconcilers, and around `watcherMu` to describe what it's actually guarding against.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+		w.stop()
+	}
+}
+
+// setEngineCondition updates the ContainerEngineReady condition on
+// the App. The App controller also writes App.Status.Conditions (to
+// mirror LimaVM conditions) and controller-runtime does not serialize
+// reconciles across controllers, so a naive Update races and 409s.
+// retry.RetryOnConflict + re-Get matches removeMirrorResource and
+// deleteAllOfType.
```
mook-as: I'm having issues parsing this comment.  I think only the first sentence should be the function documentation; the rest needs to go _inside_ the function as implementation detail, and needs to be rephrased:

```golang
	// The App controller also updates the status conditions (to mirror LimaVM
	// conditions).  As multiple controllers can update the status conditions
	// at the same time, we need to retry on conflict.
```

I don't even understand how `removeMirrorResource` and `deleteAllOfType` are relevant here.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+func (r *EngineReconciler) setEngineCondition(ctx context.Context, app *appv1alpha1.App, status metav1.ConditionStatus, reason, message string) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
mook-as, line 260: Might be useful to comment that we don't want server-side apply because we want to record the generation we saw.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+	return ctrl.Result{}, errors.Join(specsErr, finErr)
+}
+
+// startWatcher creates and starts a Docker watcher goroutine. The
+// watcher inherits watcherCtx, which cancels only on manager
+// shutdown; startWatcher deliberately drops Reconcile's ctx so a
+// future ReconciliationTimeout or per-request deadline cannot kill
+// the watcher the moment Reconcile returns.
+func (r *EngineReconciler) startWatcher(_ context.Context) error {
```
mook-as: This should be called `startWatcherAndSync` or something, to be clearer that this only returns after the full sync is complete.  It was otherwise unclear on line 189 why that was valid.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+	// Both passes issue List calls per reconcile, and every
+	// Container/Image/Volume watch event triggers a reconcile — so
+	// cost is O(N) per child event. The long-term fix is to split
+	// these into per-resource reconcilers with watch predicates
+	// (deletionTimestamp set, spec.state changed).
+	//
+	// Run both passes unconditionally and join their errors. Early
+	// return on specsErr would stall every mirror's finalizer behind
+	// a single stuck spec.state actuation: the next reconcile would
+	// hit the same broken container first and skip finalizers again.
```
mook-as, line 204: This comment is starting from the middle of the thought; it should first say what the passes are.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+	}
+
+	var containers containersv1alpha1.ContainerList
+	if err := r.List(ctx, &containers, client.InNamespace(r.apiNamespace)); err != nil {
+		return err
+	}
+
+	var errs []error
+	for i := range containers.Items {
+		c := &containers.Items[i]
```
mook-as, line 381:
```suggestion
	for _, c := range containers.Items {
```

I guess you're trying to avoid a copy or something? Not sure that's worth it…

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,452 @@
...
+//
+//	desired \ actual  | running | paused   | created | exited | dead | pausing/restarting | removing | unknown
+//	------------------+---------+----------+---------+--------+------+--------------------+----------+--------
+//	running           | nil     | unpause  | start   | start  | start| start              | start    | start
+//	created           | stop    | stop     | nil     | nil    | nil  | stop               | stop     | stop
+//
+// ContainerStop handles paused and restarting containers natively, so
+// the created branch always dispatches Stop from any non-stopped
+// state. The running branch handles paused explicitly because Docker
+// rejects ContainerStart on a paused container.
```
mook-as: Everything after the fist paragraph should go inside the function as implementation documentation, not API contract.  Or it should just go away, because that's not what we actually implement (the implementation makes more sense and doesn't look at _all_ the current states).

The bit about pause -> running should go to where it's implemented, not as the overview.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -0,0 +1,570 @@
...
+		img := &images.Items[i]
+		if img.DeletionTimestamp == nil || !controllerutil.ContainsFinalizer(img, mirrorFinalizer) {
+			continue
+		}
+		// An Image mirror with no engine-populated status has no
+		// Docker reference to forward the delete to (bare-skeleton
+		// user create, or the startup race before applyImage ran).
+		// Strip the finalizer and let the Delete proceed — symmetric
+		// with processVolumeFinalizers' empty-Status.Name guard.
+		if img.Status.ID != "" || img.Status.RepoTag != "" {
```
mook-as, line 471: The comment sounds like you should guard again `img.Status == nil` too?

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -0,0 +1,452 @@
...
+			// work lands.
+			if err := w.handleEvent(ctx, msg); err != nil {
+				log.Error(err, "Failed to handle Docker event",
+					"type", msg.Type, "action", msg.Action, "actor", msg.Actor.ID)
+			}
+		}
+	}
+}
+
+// enqueueReconcile triggers a reconcile in the engine reconciler.
```
mook-as: It should say that it only triggers a reconcile if one is not already queued.

pkg/controllers/app/engine/controllers/sync_containers_test.go
```diff
@@ -0,0 +1,55 @@
...
+			wantName:      "",
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.input, func(t *testing.T) {
+			ns, name := parseContainerName(tc.input)
+			if ns != tc.wantNamespace || name != tc.wantName {
+				t.Errorf("parseContainerName(%q) = (%q, %q), want (%q, %q)",
+					tc.input, ns, name, tc.wantNamespace, tc.wantName)
+			}
```
mook-as: This might be easier with `assert`:

```suggestion
			assert.DeepEqual(
				t,
				[]string{ns, name},
				[]string{tc.wantNamespace, tc.wantName},
			)
```

pkg/controllers/app/engine/controllers/sync_images.go
```diff
@@ -0,0 +1,267 @@
...
+// Docker image ID and its RepoTags: one name per tag (hashed over
+// id+tag), or a single name hashed over id alone for a dangling image.
+// The "img-" prefix and SHA-256 match the vol-<sha256(name)> scheme
+// used for Volumes, keeping names short, RFC 1123 valid, and
+// consistent across resource types.
+//
+// syncAllImages shares this helper to seed activeNames from the list
+// response alone, so a transient Inspect failure cannot classify
+// existing mirrors as stale.
+func imageMirrorNames(id string, repoTags []string) []string {
```
mook-as, line 38: `api_containers.md` needs to be updated

mook-as (commented):
Nothing here too scary, I think

pkg/controllers/app/engine/controllers/sync_containers.go
jandubois: We just have a failure and it will self-heal on the next reconciler run which will be scheduled as a consequence of the failure. Seems simpler than adding extra code for this.

pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: It unfortunately is, but I think a simpler flow would either stall concurrent reconciles or leak the old client. Any ideas how to simplify?

mook-as (dismissed):
It's basically fine; we can merge it as-is and do follow-ups, or do as much as you agree with right now and then punt the rest.

pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: It is idiomatic to avoid copies; we use it in several other places too.

pkg/controllers/app/engine/controllers/engine_reconciler.go
mook-as: I think we if just had a `defer w.cli.Close()` in `dockerWatcher#run` (between the `defer w.enqueueReconcile()` and the `defer close(w.done)`) we can drop the `stop` call here?  Then the contract is that `dockerWatcher.cli` is valid until `run` returns, and also that `run` does all the cleanup. `dockerWatcher#stop` can drop the `Close` too, because `w.cancel()` handles that.

mook-as (approved):

PR #315, opened 2026-04-16
app: add Settled condition, wait for it in `rdd set`

### LINKED ISSUE #264: .status updates for containerEngine and kubernetes settings
The `api_app.md` file shows some `.status` fields, like the observed values for the new properties. This still requires discussion on how we will detect changes in a timely manner, and the best way is to make them available in the `.status` field.

I think we will want to add a bunch of condions, like `ContainerEngineRunning` and `KubernetesRunning`, but the details still need to be figured out.

We should also have a "global" condition like `AppReady` which indicates that the observed status exactly matches the desired status. This could be used by `rdd set` to wait for the desired state.

jandubois:
## Summary

`rdd set` used to branch on what was being set: `running=true` waited for `ContainerEngineReady=True` (or `Running=True` when the engine controller wasn't enabled), `running=false` waited for `Running` to leave `True`, and any other property change returned immediately. The third case meant `rdd set containerEngine.name=containerd` landed the patch and walked away before the VM had restarted, and left an open TODO in `set.go` to plumb a dedicated "reconciled" signal through the App reconciler.

Add an `App.Status.Conditions[Settled]` aggregate and use it as the single synchronisation point for every `rdd set` call.

## Settled

`Settled` answers *has the reconcile chain caught up with the current spec?*. It goes True when the mirrored LimaVM `Running` condition reports a terminal reason (`Started` for a desired-running app, `Stopped` for a desired-stopped app) and, when an engine controller is registered in the cluster, the engine has observed the current App generation with `ContainerEngineReady=True`. Transient reasons like `Starting`, `Stopping`, `Reconciling`, or `RestartNeeded` hold Settled at False even if the VM is momentarily running, so `rdd set` does not return before the chain stabilises.

`Settled.ObservedGeneration` tracks the App generation the reconcile pass saw. The CLI filters on `ObservedGeneration >= post-write generation`, so a stale `Settled=True` from a previous spec cannot satisfy the wait before the App controller has observed the new spec.

For a stopped app (`desired=false`) Settled short-circuits on the engine condition: the engine controller tears its watcher down when the VM stops and no longer writes `ContainerEngineReady`, so waiting on it would hang `rdd set running=false` on an already-stopped app. A stopped VM with reason `Stopped` is settled regardless of what `ContainerEngineReady` currently says.

## Engine-aware Settled

The App reconciler queries `ControllerManagerDiscovery` on each reconcile to decide whether the engine controller is registered anywhere in the control plane. When it is, Settled gates on `ContainerEngineReady`; when it is not, the engine condition is ignored. This works for both the embedded service and split-manager deployments where engine runs in a separate process.

## rdd set

`waitForDesiredState` collapses to a single predicate: any property change waits for `Settled=True` with `ObservedGeneration >= minGen`. This removes the running-specific branching and the discovery lookup for `isEngineControllerEnabled`; the App reconciler itself now knows whether the engine is expected.

## BATS

`set.bats` intentionally runs with only the app controller registered, so the App reconciler fails to start (no LimaVM CRD to watch) and Settled is never written. The four real-set tests pass `--wait=false` (using the flag added in dbe1083) to skip the wait. `engine-docker.bats` exercises the full Settled path: the existing `rdd set --timeout=10s running=false` and containerd-backend tests pick it up without test changes.


pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +349,91 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+// spec?". It goes True when the VM has reached the desired running
+// state with a terminal reason (Started/Stopped) and the engine has
+// observed and processed the current generation. A transient phase
+// (Starting, Stopping, Reconciling, RestartNeeded) holds Settled at
+// False even if the VM is momentarily running, so `rdd set` does not
+// return before the chain stabilises.
+//
+// engineEnabled is false when no controller in this process writes
+// ContainerEngineReady (Windows, or `--controllers` excluding engine);
+// in that case the engine condition is ignored.
```
mook-as:
```suggestion
// engineEnabled is false when no controller in this process writes
// ContainerEngineReady; in that case the engine
// condition is ignored.
```
We really need to stop scattering Windows references everywhere inappropriately; that fact that it doesn't work is just a bug that we're punting on, not a permanent state of things, much like containerd.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +349,91 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		Type:               v1alpha1.AppConditionSettled,
+		ObservedGeneration: app.Generation,
+	}
+
+	switch {
+	case runningCond == nil:
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = "WaitingForLimaVM"
+		settled.Message = "Waiting for LimaVM to report its state"
+	case desiredRunning && runningCond.Reason != "Started":
```
mook-as: Shouldn't this check that `runningCond.Status` is `ConditionTrue` instead?

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +349,91 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		settled.Message = runningLimaVMMessage(runningCond, "Started")
+	case !desiredRunning && runningCond.Reason != "Stopped":
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = runningCond.Reason
+		settled.Message = runningLimaVMMessage(runningCond, "Stopped")
+	case !engineEnabled:
+		settled.Status = metav1.ConditionTrue
+		settled.Reason = "Settled"
+		settled.Message = "App has reached the desired state"
+	case !desiredRunning:
```
mook-as, line 434: You can technically get here with:

Expression | State
--- | ---
`desiredRunning` | `False`
`runningCond.Reason` | `Stopped`
`engineEnabled` | `True`

That should be a logic error (you can't want to have engine but not running), so maybe we should assert on that?

docs/design/api_app.md
```diff
@@ -106,10 +111,12 @@ status:
...
 
-  `Created` and `Running` are mirrored, so their `lastTransitionTime` reflects the LimaVM transition rather than the copy — the timestamp is meaningful for staleness checks. `ContainerEngineReady.observedGeneration` is stamped with the App's generation when the engine controller writes the condition, so `rdd set` can distinguish stale snapshots from fresh ones.
+  `Created` and `Running` are mirrored from LimaVM, so their `lastTransitionTime` reflects the LimaVM transition rather than the copy — the timestamp is meaningful for staleness checks. `ContainerEngineReady.observedGeneration` is stamped with the App's generation when the engine controller writes the condition. `Settled` is computed by the App controller from the mirrored `Running` and the engine-written `ContainerEngineReady`, and carries the App's current `observedGeneration`. `rdd set` waits for `Settled=True` with `observedGeneration >= post-patch generation` so stale snapshots cannot prematurely satisfy the wait.
```
mook-as: > … carries the App's current `observedGeneration`.

Did you mean the `App`'s `generation`? The app doesn't have an `observedGeneration`, other than as part of status conditions.

mook-as: > `ContainerEngineReady.observedGeneration` is stamped with the App's generation when the engine controller writes the condition.

Did you mean the engine reconciler?

I think it's stamping the generation of whenever it's doing the _update_, instead of whatever snapshot it thinks the state should be, so there can be awkward drifts (where it gets the app to check the desired state, that app gets updated from under it, and then it stamps a generation that's newer than the initial observed state).

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +349,91 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		settled.Reason = "Settled"
+		settled.Message = "App has reached the desired state"
+	case engineCond == nil:
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = "WaitingForEngine"
+		settled.Message = "Waiting for container engine condition"
+	case engineCond.ObservedGeneration < app.Generation:
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = "EngineStale"
+		settled.Message = "Container engine has not yet observed this generation"
```
mook-as: I think a more understandable message would be something like
> Container engine needs to be synchronized

pkg/controllers/base/controller.go
```diff
@@ -118,6 +119,44 @@ func GetAllControllers() []Controller {
 	return defaultRegistry.GetAllControllers()
 }
 
+// SetEnabledControllers records which controllers started in this
+// process after --controllers filtering. The controller-manager entry
+// point calls it once before any controller begins reconciling.
+// Reconcilers that need to know whether a sibling runs in the same
+// process should query IsControllerEnabled instead of GetAllControllers,
+// which reflects only the compile-time registry.
+func (r *Registry) SetEnabledControllers(names []string) {
```
mook-as: If these are only for the in-process controllers, the name should say that; currently it sounds like it would check for all controllers regardless of which procss.

`SetInProcessControllers` etc maybe?

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -323,6 +335,8 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 				LastTransitionTime: cond.LastTransitionTime,
 			}) || changed
 		}
+		settled := computeSettledCondition(latest, r.EngineEnabled)
```
mook-as: Why this only care about in-process controllers? It sounds like if I run all controllers externally, it should still report if the engine state matches.  That is, it sounds like it should be looking at `ControllerManagerDiscovery.IsControllerRunning`?

mook-as (changes requested):

pkg/controllers/app/app/controllers/app_controller.go
jandubois: I don't understand. Isn't that explained by the comment below? If you want the App stopped, and it is not running, then the engine condition doesn't matter. I'm probably not getting what you are saying.

mook-as: Oh, you're right, I think.

docs/design/api_app.md
```diff
@@ -106,10 +111,15 @@ status:
...
+  The `Created` and `Running` conditions are mirrored from LimaVM, so their `lastTransitionTime` reflects the LimaVM transition rather than the copy — the timestamp is meaningful for staleness checks. The engine reconciler stamps `ContainerEngineReady.observedGeneration` with the App's `metadata.generation` at the time it writes the condition; if the App's generation advances between the reconciler's read and write, the stamp reflects the write-time generation rather than the observed one. The App reconciler computes `Settled` from the mirrored `Running` and the engine-written `ContainerEngineReady`, and stamps its own `observedGeneration` with the App's current `metadata.generation`. `rdd set` waits for `Settled.status=True` with `observedGeneration >= post-patch metadata.generation`, so stale snapshots cannot prematurely satisfy the wait.
```
mook-as: Importantly, computing `Settled` stamps it with the generation of `app` that was _observed_, not the generation _as it was when it was written_.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +375,92 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = "WaitingForLimaVM"
+		settled.Message = "Waiting for LimaVM to report its state"
+	case desiredRunning && runningCond.Status != metav1.ConditionTrue:
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = runningCond.Reason
+		settled.Message = runningLimaVMMessage(runningCond, "Started")
+	// Status=False covers Stopped as well as the transient Starting/Stopping
+	// reasons, so we must match the terminal Stopped reason explicitly.
+	case !desiredRunning && runningCond.Reason != "Stopped":
```
mook-as, line 426: I guess there's no need to check `runningCond.Status`, because `True` wouldn't have `Stopped` as a reason?

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +375,92 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = runningCond.Reason
+		settled.Message = runningLimaVMMessage(runningCond, "Stopped")
+	case !engineEnabled:
+		settled.Status = metav1.ConditionTrue
+		settled.Reason = "Settled"
+		settled.Message = "App has reached the desired state"
```
mook-as: I guess we get here on:
- desired: running; running condition: true; engine not enabled.
- desired: false; running condition: stopped; engine not enabled.

That is to say, desired state matches running condition, and engine not enabled. Makes sense.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +375,92 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+		settled.Message = "Waiting for LimaVM to report its state"
+	case desiredRunning && runningCond.Status != metav1.ConditionTrue:
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = runningCond.Reason
+		settled.Message = runningLimaVMMessage(runningCond, "Started")
+	// Status=False covers Stopped as well as the transient Starting/Stopping
+	// reasons, so we must match the terminal Stopped reason explicitly.
+	case !desiredRunning && runningCond.Reason != "Stopped":
+		settled.Status = metav1.ConditionFalse
+		settled.Reason = runningCond.Reason
```
mook-as, line 428: The documentation says this should be `WaitingForLimaVM`?  Not sure which should be updated.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -335,6 +375,92 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
...
+	}
+	return settled
+}
+
+// runningLimaVMMessage builds the Settled message when LimaVM's
+// Running reason does not match the desired state. Failure reasons
+// (ending in "Failed") propagate LimaVM's diagnostic message; other
+// reasons get a concise "has not yet reached <desired>" text.
+func runningLimaVMMessage(runningCond *metav1.Condition, desired string) string {
+	if strings.HasSuffix(runningCond.Reason, "Failed") && runningCond.Message != "" {
```
mook-as, line 468: [LimaVM documentation](https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/d1c228f497ada6d12deffc5e78eca5fec7ec2c45/docs/design/api_lima.md#example-limavm-object) should specify that `Failed` conditions can get forwarded, so we don't end up creating a differently-named state.

pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+
+func (f fakeDiscovery) GetEnabledControllers(_ context.Context) ([]string, error) {
+	return f.enabled, f.err
+}
+
+func Test_computeSettledCondition(t *testing.T) {
+	t.Parallel()
+
+	// makeApp builds an App carrying the given conditions. Generation
+	// defaults to 2 so stale ObservedGeneration values have room below.
```
mook-as: But there is no default value, all callers have to pass in `2` manually?

pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+	makeApp := func(generation int64, running bool, conds ...metav1.Condition) *v1alpha1.App {
+		app := &v1alpha1.App{Spec: v1alpha1.AppSpec{Running: running}}
+		app.Generation = generation
+		app.Status.Conditions = append(app.Status.Conditions, conds...)
+		return app
```
mook-as: Is this the same as:

```suggestion
		return &v1alpha1.App{
			ObjectMeta: metav1.ObjectMeta{Generation: generation},
			Spec:       v1alpha1.AppSpec{Running: running},
			Status: v1alpha1.AppStatus{
				Conditions: conds,
			},
		}
```

pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+			wantReason:    "Starting",
+			wantMessage:   "LimaVM has not yet reached Started",
+		},
+		{
+			name:          "StartFailed surfaces LimaVM message",
+			app:           makeApp(2, true, running("StartFailed", "template failed to parse", metav1.ConditionFalse, 2)),
+			engineEnabled: true,
+			wantStatus:    metav1.ConditionFalse,
+			wantReason:    "StartFailed",
+			wantMessage:   "template failed to parse",
```
mook-as: If want to test message forwarding, the message should be more obviously nonsensical so we know it wasn't looked at. Something like `ExplosionFailed` / _the virtual machine did not explode_.

pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -0,0 +1,235 @@
...
+		{
+			name: "engine ready at older generation is stale",
+			app: makeApp(2, true,
+				running("Started", "VM is running", metav1.ConditionTrue, 2),
+				engine("Ready", "engine is ready", metav1.ConditionTrue, 1),
+			),
+			engineEnabled: true,
+			wantStatus:    metav1.ConditionFalse,
+			wantReason:    "EngineStale",
+			wantMessage:   "Container engine needs to be synchronized",
```
mook-as: I feel like we should make the messages constants if we want to assert on them…

pkg/controllers/base/controller.go
```diff
@@ -118,6 +119,46 @@ func GetAllControllers() []Controller {
...
+// SetInProcessControllers records the in-process controllers on the global registry.
+func SetInProcessControllers(names []string) {
+	defaultRegistry.SetInProcessControllers(names)
+}
+
+// IsControllerInProcess reports whether a controller with the given name
+// is running in this process. Returns false if SetInProcessControllers
+// has not been called, so external consumers in tests that skip
+// controller startup see a consistent "nothing in process" view.
+func (r *Registry) IsControllerInProcess(name string) bool {
```
mook-as: I don't think anybody calls `IsControllerInProcess` anymore (because we switched to `ControllerDiscovery`), so this (and `SetInProcessControllers`) can be removed.

mook-as (changes requested):

jandubois:
I believe I've addressed all requested changes.

mook-as (approved):

PR #316, opened 2026-04-16
engine: replace Container.spec.state with action annotations

### LINKED ISSUE #298: Replace `container.spec.state` with annotations
The `spec.state` mechanism does not work for `Container` objects because the control plane is not the source of truth, the container engine is.

E.g. if the user sets `spec.state=running`, the reconciler can start a stopped container. But if the user then tries to pause or stop the container from the CLI, the reconciler will always restart it again automatically.

We need a mechanism that triggers an action once, not one that prescribes the new state forever (edge-trigger vs. level-trigger).

We already have this mechanism for LimaVM resources which can be restarted or reset by applying an annotation, which the reconciler will remove once the action is initiated. We should do the same for Containers, with a different annotation for each possible container state (start/stop/pause/unpause/...).

jandubois:
Stacked on top of #309 (engine-v3). Do not merge before that lands. PR #315 stacks on this one and adds the App Ready/Settled conditions that depend on these types.

## Summary

`Container.spec.state` was a level-triggered desired-state field (`unknown` / `running` / `created`), but Docker's restart policy and direct `docker start` / `docker stop` are concurrent writers — the reconciler thrashed against them. The mirror cannot own state Docker itself mutates.

Replace the field with a one-shot annotation `containers.rancherdesktop.io/action: start|stop|pause|unpause|restart` on `metadata`. The reconciler dispatches the Docker call, records the outcome in a new `status.lastAction` field (`action` / `state=InProgress|Succeeded|Failed` / `error` / `requestedAt` / `completedAt`), then removes the annotation. A single annotation holds at most one pending action; writing a new value atomically replaces any pending one, so there is no queue and no accumulating history.

Pause/unpause pre-check Docker state and skip when the container is already in the target state — without the pre-check, controller-runtime's informer cache can deliver the same annotation to two consecutive reconcile passes before the second one sees the removal from the first, and the duplicate dispatch would flip `lastAction` to `Failed`. Start/stop are naturally idempotent (304 Not Modified); restart is the one exception and may double-restart on crash recovery.

`ContainerCreateRequest.spec.state` is untouched — a create request is inherently one-shot, so edge semantics already apply.

## API and SSA fallout

`ContainerSpec` becomes an empty struct. `Container.Spec` is now optional; the CRD no longer lists `spec` in its top-level `required`. The immutable-spec webhook simplifies to "spec is immutable on update" — trivially true for an empty struct, but enforces the intent that nothing belongs there.

With spec gone, the two-field-owner SSA dance in `sync_containers.go` is unnecessary: the old code split finalizer reasserts onto a separate field manager (`engine-controller-finalizer`) to avoid pruning `spec`. Empty spec means no spec to prune, so `finalizerFieldOwner` is deleted and `applyContainer` collapses the create and finalizer-reassert branches into a single ForceOwnership apply under `engine-controller`.

## Docs

Rewrote the "Containers", "Container state", and "Change container state" sections of `docs/design/api_containers.md` around the annotation model and `lastAction`.

## Tests

`bats/tests/32-app-controllers/engine-docker.bats` rewrites the four `spec.state` tests into five action tests (start, stop, pause+unpause, failure path on stopped container, restart). A helper `request_action` sets the annotation and waits for the reconciler to consume it; `assert_last_action` checks the recorded outcome.

## Test plan

- [x] `bats-app` passes 105/105 locally on macOS.
- [x] `go build ./...`, `go vet ./...`, `go test -run ^$ ./...`
- [ ] `bats-non-app` — not run locally; trusting CI.
- [ ] Windows — not run locally.


jandubois:
I don't think we need any CLI support for setting container actions. The feature exists solely for the benefit of the GUI. Terminal users can use `docker` and `nerdctl` to talk to the container engine directly and should not go through the control plane mirrors.

docs/design/api_containers.md
```diff
@@ -210,8 +220,52 @@ An admission controller will ensure that we cannot have multiple
...
-stops the container.
+
+Set the `containers.rancherdesktop.io/action` annotation on the
+`Container` to request a one-shot action. The engine reconciler reads
+the annotation, dispatches the matching Docker call, records the
+outcome in `status.lastAction`, and removes the annotation.
+
+Valid values:
+
+| action    | Docker call         |
```
mook-as: I don't think the `Docker call` column is relevant.

docs/design/api_containers.md
```diff
@@ -146,23 +150,29 @@ status:
...
   - type: Dead
     status: False
+  # Outcome of the most recent action. Persists until the next action
+  # overwrites it, regardless of any observable state changes (e.g. a
+  # direct `docker stop`) in between. The `error` field is set only
+  # when `state` is `Failed`.
+  lastAction:
+    action: pause
+    state: Succeeded
+    observedAt: "2026-04-15T10:30:00Z"
```
mook-as, line 160: Should that be using `observeAt`, or `observedGeneration`? Hmm, I guess annotations don't affect the generation, so it wouldn't help?

At this is basically a slightly-odd status condition, right? Not that making it a status condition is necessary the right thing to do here… let's not, just yet, until we think through the rest.

`lastAction` | status condition
--- | ---
(nothing) | `type` = `ActionSucceeded`
`action` | `reason`
`state` | `status`
`error` | `message`
`observedAt` | `observedGeneration`
`completedAt` | `LastTransitionTime`

The main upside to changing is that we already have `rdd ctl wait-condition`… :D  Yeah, probably not useful here.

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -241,95 +242,245 @@ local_setup_file() {
...
+
+    rdd ctl wait --for=jsonpath='{.status.status}'=running \
         --namespace="${RDD_NAMESPACE}" container/"${cid}" --timeout=30s
-    run -0 docker inspect test-unpause --format '{{.State.Status}}'
-    assert_output "paused"
 
-    rdd ctl patch container "${cid}" --namespace="${RDD_NAMESPACE}" \
-        --type=merge -p '{"spec":{"state":"running"}}'
+    # Record the pre-restart StartedAt so we can verify the container
+    # was actually restarted rather than left untouched.
```
mook-as, line 410: Another option is to inspect the pid, and make sure that changed (but we're still running).

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -340,65 +342,197 @@ func (w *dockerWatcher) removeMirrorResource(ctx context.Context, obj client.Obj
...
+		if err != nil {
 			return err
 		}
-		log.Info("Starting container", "id", c.Name)
-		_, err := w.cli.ContainerStart(ctx, c.Name, dockerclient.ContainerStartOptions{})
-		return err
-	case containersv1alpha1.ContainerStatusCreated:
-		if isStopped(actual) {
+		if !running {
+			return fmt.Errorf("container %s is not running", id)
```
mook-as: Mention that we're trying to unpause in the error message?

pkg/controllers/app/engine/controllers/docker_watcher.go
```diff
@@ -340,65 +342,197 @@ func (w *dockerWatcher) removeMirrorResource(ctx context.Context, obj client.Obj
...
-	case containersv1alpha1.ContainerStatusCreated,
-		containersv1alpha1.ContainerStatusExited,
-		containersv1alpha1.ContainerStatusDead:
-		return true
+// containerRunState reports whether Docker currently shows the container as
+// running and as paused. Pause uses paused to skip a no-op dispatch; unpause
+// also checks running so that unpause on a stopped container reports a
+// failure instead of a silent success.
+func (w *dockerWatcher) containerRunState(ctx context.Context, id string) (running, paused bool, err error) {
+	inspect, err := w.cli.ContainerInspect(ctx, id, dockerclient.ContainerInspectOptions{})
```
mook-as, line 465: Doing a full inspect seems a bit expensive but I don't have better ideas that aren't a pain to implement, so I guess it's fine.

mook-as (dismissed):
Just some tiny things

pkg/controllers/app/engine/controllers/docker_watcher.go
jandubois: It only happens when you execute an action on the container, so I think it is fine.

jandubois:
I dropped the docker call table, and I've improved the error message. I don't want to do anything else.

The first round of CI had a failure in `bats-lima` on Windows. It is a flake and seems to have to do with `nxadm/tail` not always following the new file after a rotation. Will investigate separately.

mook-as (approved):

PR #320, opened 2026-04-16
  Manage Docker CLI context on ContainerEngineReady transitions

### LINKED ISSUE #318: Manage docker context
When `rdd` is configured to use the `moby` engine, it needs to also create and delete the corresponding docker context (this is "documented" in `docs/design/bootstrap.md`).

### Desired behaviour

- When the App's `ContainerEngineReady` condition transitions to `True` with reason `Connected`:
    - Create (or update) a Docker CLI context named after the instance (e.g. `rancher-desktop-2`) pointing at `instance.DockerSocket()`.
    - Probe the user's current Docker context. If no current context is set, or if it fails a liveness check (e.g. `client.Ping` with a short timeout), switch the default to the new context. If the current context is healthy, leave it alone — the user can switch manually.
- When `ContainerEngineReady` transitions away from `True` or the App is deleted:
    - Remove the `rancher-desktop-{suffix}` context. If it was the active default, Docker falls back to default.

### Where the code goes

The natural home is the engine reconciler at `pkg/controllers/app/engine/controllers/engine_reconciler.go`. It is the component that owns the `ContainerEngineReady` condition.

Reuse the existing watcher-lifetime context pattern already in the reconciler (`watcherCtx` / `watcherCancel`): the context is established by the manager's `RunnableFunc` and cancelled on shutdown, so a goroutine derived from it survives individual `Reconcile` calls but is guaranteed to be torn down when the instance stops.

### Why a goroutine

Probing the existing context requires a Docker API round-trip that may hang (unreachable endpoint, stale socket). Doing this inline would block `Reconcile` and any other work sharing the reconcile worker. The goroutine should:

  - Derive its context from `watcherCtx` (cancelled on manager shutdown) and additionally be cancelled if `ContainerEngineReady` leaves `True` before the probe finishes — otherwise a slow probe could promote our context after the engine has already gone away.
  - Guard against concurrent invocations (mutex or a single in-flight token) so repeated reconciles while the probe is running do not fan out.
  - Use a bounded timeout on the liveness probe; treat timeout as "not healthy".

Nino-K:
When `ContainerEngineReady` transitions to True for the moby engine, a Docker CLI context is created (named after the instance, e.g., `rancher-desktop-2`) pointing to `instance.DockerSocket()`.

A background probe then checks the user's current Docker context. If the current context is missing or points to an invalid socket, the newly created context is set as the default.

When `ContainerEngineReady` transitions away from True, any in-flight probe is cancelled. If the active context is ours, it is cleared, and the associated context directory is removed.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/318

### DIFF
--- a/bats/tests/32-app-controllers/engine-docker.bats
+++ b/bats/tests/32-app-controllers/engine-docker.bats
@@ -405,3 +405,89 @@ local_setup_file() {
     run -0 rdd ctl get ContainerNamespaces --namespace="${RDD_NAMESPACE}" --output=name
     refute_output
 }
+
+# --- Docker context management ---
+
+# docker_context_dir returns the ~/.docker/contexts/meta/<hash> directory for
+# the named context. Docker derives the sub-directory from sha256(name).
+docker_context_dir() {
+    local name="$1"
+    local hash
+    # sha256sum on Linux, shasum on macOS
+    if command -v sha256sum &>/dev/null; then
+        hash=$(printf '%s' "${name}" | sha256sum | awk '{print $1}')
+    else
+        hash=$(printf '%s' "${name}" | shasum -a 256 | awk '{print $1}')
+    fi
+    echo "${HOME}/.docker/contexts/meta/${hash}"
+}
+
+@test "moby engine creates Docker context for the instance" {
+    # Restart with moby (the containerd test above may have left the engine in
+    # containerd mode).
+    rdd set running=true containerEngine.name=moby
+    rdd ctl wait --for=condition=ContainerEngineReady \
+        app/app --timeout=30s
+
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+    local meta_file
+    meta_file="$(docker_context_dir "${context_name}")/meta.json"
+
+    assert_file_exists "${meta_file}"
+
+    run -0 jq -r '.Name' "${meta_file}"
+    assert_output "${context_name}"
+
+    run -0 rdd service paths docker_socket
+    local socket_path=${output}
+    run -0 jq -r '.Endpoints.docker.Host' "${meta_file}"
+    assert_output "unix://${socket_path}"
+}
+
+@test "moby engine sets currentContext when no healthy context exists" {
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+
+    # Save and clear any existing currentContext so the probe finds no
+    # healthy context and promotes ours. Restored in teardown.
+    local saved_context
+    saved_context=$(jq -r '.currentContext // empty' "${HOME}/.docker/config.json" 2>/dev/null || true)
+
+    # Clear the current context and restart the engine so the probe runs fresh.
+    jq 'del(.currentContext)' "${HOME}/.docker/config.json" >"${HOME}/.docker/config.json.tmp" &&
+        mv "${HOME}/.docker/config.json.tmp" "${HOME}/.docker/config.json"
+
+    rdd set running=false
+    rdd set running=true containerEngine.name=moby
+    rdd ctl wait --for=condition=ContainerEngineReady app/app --timeout=30s
+
+    # The probe goroutine runs asynchronously; give it a moment.
+    try --max 6 --delay 1 -- \
+        bash -c "jq -r '.currentContext' '${HOME}/.docker/config.json' | grep -qx '${context_name}'"
+
+    run -0 jq -r '.currentContext' "${HOME}/.docker/config.json"
+    assert_output "${context_name}"
+
+    # Restore the original context if there was one.
+    if [[ -n "${saved_context}" ]]; then
+        jq --arg ctx "${saved_context}" '.currentContext = $ctx' \
+            "${HOME}/.docker/config.json" >"${HOME}/.docker/config.json.tmp" &&
+            mv "${HOME}/.docker/config.json.tmp" "${HOME}/.docker/config.json"
+    fi
+}
+
+@test "stopping moby engine removes Docker context and clears currentContext" {
+    rdd set running=false
+
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+    run_e -0 docker_context_dir "${context_name}"
+    local context_dir="${output}"
+
+    # removeDockerContext runs asynchronously after the reconciler processes
+    # the Running=False transition; poll until the directory is gone.
+    try --max 10 --delay 1 -- test ! -d "${context_dir}"
+    assert_dir_not_exists "${context_dir}"
+
+    # currentContext should either be gone or point to something else.
+    run jq -r '.currentContext // empty' "${HOME}/.docker/config.json"
+    refute_output "${context_name}"
+}
--- a/pkg/controllers/app/engine/controllers/docker_context.go
+++ b/pkg/controllers/app/engine/controllers/docker_context.go
@@ -1,0 +1,226 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package controllers
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"os"
+	"path/filepath"
+	"time"
+
+	dockerclient "github.com/moby/moby/client"
+)
+
+// dockerContextProbeTimeout is the maximum time allowed to ping a Docker
+// socket when checking the user's current context is healthy.
+const dockerContextProbeTimeout = 3 * time.Second
+
+// dockerContextMeta is the on-disk format of a Docker CLI context metadata file.
+type dockerContextMeta struct {
+	Name      string                        `json:"Name"`
+	Metadata  map[string]any                `json:"Metadata"`
+	Endpoints map[string]dockerEndpointMeta `json:"Endpoints"`
+}
+
+type dockerEndpointMeta struct {
+	Host          string `json:"Host"`
+	SkipTLSVerify bool   `json:"SkipTLSVerify"`
+}
+
+func dockerConfigDir() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".docker"), nil
+}
+
+// contextMetaPath returns the path to the meta.json file for the named context.
+// Docker creates the directory name using sha256(contextName).
+func contextMetaPath(name string) (string, error) {
+	dir, err := dockerConfigDir()
+	if err != nil {
+		return "", err
+	}
+	sum := sha256.Sum256([]byte(name))
+	return filepath.Join(dir, "contexts", "meta", hex.EncodeToString(sum[:]), "meta.json"), nil
+}
+
+func createReplaceDockerContext(name, socketPath string) error {
+	path, err := contextMetaPath(name)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(path), 0o700); err != nil {
+		return err
+	}
+	meta := dockerContextMeta{
+		Name:     name,
+		Metadata: map[string]any{},
+		Endpoints: map[string]dockerEndpointMeta{
+			"docker": {Host: "unix://" + socketPath},
+		},
+	}
+	data, err := json.MarshalIndent(meta, "", "\t")
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(path, append(data, '\n'), 0o600)
+}
+
+func deleteDockerContext(name string) error {
+	path, err := contextMetaPath(name)
+	if err != nil {
+		return err
+	}
+	err = os.RemoveAll(filepath.Dir(path))
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	return err
+}
+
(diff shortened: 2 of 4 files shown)


bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -405,3 +405,90 @@ local_setup_file() {
...
+    # containerd mode).
+    rdd set running=true containerEngine.name=moby
+    rdd ctl wait --for=condition=ContainerEngineReady \
+        app/app --timeout=30s
+
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+    local meta_file
+    meta_file="$(docker_context_dir "${context_name}")/meta.json"
+
+    run -0 test -f "${meta_file}"
```
mook-as:
```suggestion
    assert_file_exists "${meta_file}"
```

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -405,3 +405,90 @@ local_setup_file() {
...
+
+    run -0 jq -r '.Name' "${meta_file}"
+    assert_output "${context_name}"
+
+    local socket_path
+    socket_path=$(rdd svc paths docker_socket)
```
mook-as:
```suggestion
    run -0 rdd service paths docker_socket
    local socket_path=${output}
```

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -405,3 +405,90 @@ local_setup_file() {
...
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+
+    # Save and clear any existing currentContext so the probe finds no
+    # healthy context and promotes ours. Restored in teardown.
+    local saved_context
+    saved_context=$(jq -r '.currentContext // empty' "${HOME}/.docker/config.json" 2>/dev/null || true)
+
+    # Clear the current context and restart the engine so the probe runs fresh.
+    jq 'del(.currentContext)' "${HOME}/.docker/config.json" >"${HOME}/.docker/config.json.tmp" &&
+        mv "${HOME}/.docker/config.json.tmp" "${HOME}/.docker/config.json"
```
mook-as, line 457: If we accept `yq`, `yq --inplace --input-format=json --output-format=json 'del(.currentContext)' "${HOME}…"` would work too.

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -405,3 +405,90 @@ local_setup_file() {
...
+
+    rdd set running=false
+    rdd set running=true containerEngine.name=moby
+    rdd ctl wait --for=condition=ContainerEngineReady app/app --timeout=30s
+
+    # The probe goroutine runs asynchronously; give it a moment.
+    try --max 6 --delay 1 -- \
+        bash -c "jq -r '.currentContext' '${HOME}/.docker/config.json' | grep -qx '${context_name}'"
```
mook-as, line 465: That sounds like `ContainerEngineReady` shouldn't be set until the the context is set up correctly.

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -405,3 +405,90 @@ local_setup_file() {
...
+
+    local context_name="rancher-desktop-${RDD_INSTANCE}"
+
+    # removeDockerContext runs asynchronously after the reconciler processes
+    # the Running=False transition; poll until the directory is gone.
+    try --max 10 --delay 1 -- \
+        bash -c "! test -d '$(docker_context_dir "${context_name}")'"
+
+    run test -d "$(docker_context_dir "${context_name}")"
+    assert_failure
```
mook-as:
```suggestion
    local context_name="rancher-desktop-${RDD_INSTANCE}"
    run_e -0 docker_context_dir "${context_name}"
    local context_dir="${output}"

    # removeDockerContext runs asynchronously after the reconciler processes
    # the Running=False transition; poll until the directory is gone.
    try --max 10 --delay 1 -- test ! -d "${context_dir}"

    assert_dir_not_exists "${context_dir}"
```

There's no need to repeatedly calculate `context_dir`.  `test` is an actual executable (`/usr/bin/test`), no need to run `bash -c …`.  And there's a specific assertion for directory not existing, which should produce better error messages.

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -0,0 +1,203 @@
...
+
+	dockerclient "github.com/moby/moby/client"
+)
+
+// dockerContextProbeTimeout is the maximum time allowed to ping a Docker
+// socket when checking the user's current context is healthy.
+const dockerContextProbeTimeout = 3 * time.Second
+
+// dockerContextMeta is the on-disk format of a Docker CLI context metadata file.
+type dockerContextMeta struct {
```
mook-as, line 25: I wonder if we can just use https://pkg.go.dev/github.com/docker/cli@v29.4.0+incompatible/cli/context/store#ContextStore ?

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+	err = os.RemoveAll(filepath.Dir(path))
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	return err
+}
+
+// getDockerContextSocket returns the Unix socket path that the named context
```
mook-as:
```suggestion
// getDockerContextSocket returns the socket path that the named context
```
It's _close enough_ for Windows, which defaults to named pipes… or are we going to skip all this on Windows and just use the default path?

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -245,6 +262,84 @@ func (r *EngineReconciler) stopWatcher() {
 	if w != nil {
 		w.stop()
 	}
+	r.removeDockerContext()
+}
+
+// manageDockerContext creates the instance Docker context and, in a goroutine,
+// probes the user's current context; if it is absent or unhealthy, switches
+// the default to the new context. At most one probe runs at a time.
+func (r *EngineReconciler) manageDockerContext(socketPath string) {
```
mook-as, line 276: Huh, somehow I got the impression from the PR description that the probe is long-running, but re-reading it, it doesn't actually say that.  Not sure why I had that impression…

mook-as (changes requested):

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -0,0 +1,203 @@
...
+// points at. Returns an empty string if the context does not exist or has no
+// docker endpoint.
+func getDockerContextSocket(name string) (string, error) {
```
mook-as: Actually, this gets the socket of an arbitrary context; that means it'll fall over when the context is, say, TCP.

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	if err != nil {
+		return "", err
+	}
+	var meta dockerContextMeta
+	if err := json.Unmarshal(data, &meta); err != nil {
+		return "", err
+	}
+	host := meta.Endpoints["docker"].Host
+	return strings.TrimPrefix(host, "unix://"), nil
```
mook-as: We're removing the `unix://` prefix and only to add it back in `probeDockerContext`? Why not just leave it as-is?

pkg/controllers/app/engine/controllers/docker_context.go
Nino-K: I thought this made sense to change, but once I started to change it, I realized that `github.com/docker/cli` is not in `go.mod`. Adding it would pull in a large dependency tree just to replace ~30 lines of
  simple JSON writing we fully control. The ContextStore API also abstracts over multiple backends (TLS, named pipes), which we don't need yet, we only ever write a single Unix socket endpoint.

pkg/controllers/app/engine/controllers/docker_context.go
Nino-K: I think this is going to happen on Windows too. But, regardless, I dropped the `Unix` for now.

pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: The probe can be long-running when the `client.Ping()` call hangs...

mook-as: I meant long-running as in it watches the context in a loop and fixes things when they deviate, but now that I type that out it sounds like an obectively terrible idea.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -245,6 +267,104 @@ func (r *EngineReconciler) stopWatcher() {
...
+	// Docker context management uses unix:// sockets; Windows requires
+	// npipe:// which is not yet implemented.
```
mook-as, line 278: We should either have `TODO:` that we can grep for, or an issue filed.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -245,6 +267,104 @@ func (r *EngineReconciler) stopWatcher() {
...
+			if r.contextProbeGen == myGen {
+				r.contextProbeCancel = nil
+			}
+			r.contextMu.Unlock()
+			cancel()
+		}()
+
+		current, err := getCurrentDockerContext()
+		if err != nil {
+			log.Error(err, "Failed to read current Docker context")
```
mook-as, line 316: We're just going to continue with the probe when the get fails? We don't get an error if the file doesn't exist, so that would be if the file has issues or we failed to read.  That would mean data loss as we potentially clobber, right?

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -245,6 +267,104 @@ func (r *EngineReconciler) stopWatcher() {
...
+	r.contextMu.Lock()
+	if r.contextProbeCancel != nil {
+		r.contextProbeCancel()
+		r.contextProbeCancel = nil
+	}
+	r.contextMu.Unlock()
+
+	// Wait for the probe goroutine to finish before deleting the context
+	// directory. The goroutine's probeCtx was just cancelled, so Ping will
+	// return within dockerContextProbeTimeout (3s) at most.
```
mook-as, line 357: I think it's technically possible to have another call to `manageDockerContext` trigger while the probe is off doing its thing, causing `contextProbeWg` to go above 1? I've not fully thought it out, though, and even in that case we probably don't care.

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -0,0 +1,226 @@
...
+func contextMetaPath(name string) (string, error) {
+	dir, err := dockerConfigDir()
+	if err != nil {
+		return "", err
+	}
+	sum := sha256.Sum256([]byte(name))
+	return filepath.Join(dir, "contexts", "meta", hex.EncodeToString(sum[:]), "meta.json"), nil
+}
+
+func createReplaceDockerContext(name, socketPath string) error {
```
mook-as, line 55: I was going to ask about locking around modifying these files, and then realized other processes can do that too so it's a bit pointless.

mook-as (approved):
Just needs to have an issue filed to keep track of the Windows thing I think

jandubois:
I haven't had time to look at this yet, but I want us to make sure we will not lose the user's `config.json`, so any replacement must be done atomically, similar how we handle updates to `.bash_profile` etc.

Nino-K:
@mook-as 

> I meant long-running as in it watches the context in a loop and fixes things when they deviate, but now that I type that out it sounds like an obectively terrible idea.

I thought the current design, one-time probe on the engine start and clean removal on stop, actually seems right. A polling loop that continuously reasserts the context would lead to bad scenarios like: 

- Fight the user every time they deliberately `run docker context use ...`

- I think it would add I/O and goroutine overhead for no real benefit 🤷

pkg/controllers/app/engine/controllers/engine_reconciler.go
Nino-K: I guess it would make sense to return here, since an error here means the config file exists but could not be read or parsed. So the idea would be to not proceed with writing currentContext, which would overwrite a file we cannot safely round-trip?? I can either make the change here in this PR or do in when I'm doing the Windows npipe stuff.

pkg/controllers/app/engine/controllers/engine_reconciler.go
Nino-K: You are right, the count can go above one, but it shouldn't be a problem, I don't think. The `Wait()` just blocks until the count reaches 0, it should work correctly regardless of the count.

Nino-K:
> I haven't had time to look at this yet, but I want us to make sure we will not lose the user's `config.json`, so any replacement must be done atomically, similar how we handle updates to `.bash_profile` etc.

@jandubois I already addressed it in the AI review that you sent me. I think it was the issue I3 if I recall correctly.

mook-as (approved):

jandubois (approved):
I'm fine with merging this as-is, but there are several issues in https://rancher-sandbox.github.io/rancher-desktop-daemon/20260417-123813-pr-320.html that should be addressed in a follow-up PR.

PR #323, opened 2026-04-17
implement log passthrough

mook-as:
Initial implementation of log passthrough for docker only.

### DIFF
--- a/bats/tests/32-app-controllers/engine-docker.bats
+++ b/bats/tests/32-app-controllers/engine-docker.bats
@@ -28,6 +28,18 @@ local_setup_file() {
     export RDD_NAMESPACE
 }
 
+# Make a WebSocket request, with the endpoint given as the first argument in the
+# same form as `kubectl get --raw`; the output is captured in `${output}`.
+do_websocket() { # endpoint
+    local endpoint=$1
+    run_e -0 rdd ctl config view --minify --flatten --output=jsonpath='{.clusters[].cluster.server}'
+    local server_url=${output}
+    run_e -0 rdd ctl config view --minify --flatten --output=jsonpath='{.users[].user.token}'
+    local token=${output}
+    run -0 curl --header "Authorization: Bearer ${token}" --insecure --max-time 30 \
+        "${server_url/http/ws}${endpoint}"
+}
+
 # --- Startup ---
 
 @test "ContainerNamespace moby exists" {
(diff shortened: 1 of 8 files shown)


jandubois:
AI review at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260417-165817-pr-323.html

mook-as:
Digging into things adjacent to the AI review, I realized that the [`Client.ContainerLogs`](https://pkg.go.dev/github.com/moby/moby/client#Client.ContainerLogs) is… impossible to work with.  Per its docs:

> The stream format on the response uses one of two formats:
> - If the container is using a TTY, there is only a single stream (stdout) and data is copied directly from the container output stream, no extra multiplexing or headers.
> - If the container is *not* using a TTY, streams for stdout and stderr are multiplexed.

Normally, you can examine the [Content-Type](https://docs.docker.com/reference/api/engine/version/v1.54/#tag/Container/operation/ContainerAttach) to figure out which you have; however, the `CotnainerLogs` API returns a [result](https://pkg.go.dev/github.com/moby/moby/client#ContainerLogsResult) that doesn't expose the header (… looking at the implementation, it doesn't even look at it).

In contrast, [`ContainerAttach`](https://pkg.go.dev/github.com/moby/moby/client#Client.ContainerAttach) returns something that [exposes `MediaType`](https://pkg.go.dev/github.com/moby/moby/client#HijackedResponse.MediaType).  But we can't rewrite logs to use that API either, because that doesn't return history logs (i.e. I believe it's effectively `tail=0`).

jandubois:
Opus 4.7 says:

>  The standard fix (what the docker CLI itself does in `cli/command/container/logs.go`): call `ContainerInspect` first and branch on `Config.Tty`.
>
> ```
>  inspect, err := r.watcher.cli.ContainerInspect(req.Context(), containerID)
>  // ...
>  if inspect.Config.Tty {
>      _, err = io.Copy(writer, reader)
>  } else {
>      _, err = stdcopy.StdCopy(writer, writer, reader)
>  }
> ```
>  `TTY` is fixed at container creation and can't change, so there's no race concern. This also gives him a natural place to return `404` via `errdefs.IsNotFound(err)` on the inspect — before opening the WebSocket, which is cleaner than discovering it mid-stream.

mook-as:
Went through a few rounds of local AI review. Unfortunately, that basically tripled the patch size because it kept finding things… :D

mook-as:
AI Feedback:

(I notice that all the code snippets are wrong, making them unreadable.)

> I1. Unsynchronized read of r.apiNamespace in HandleLogs

Yes, there's a race, but how did you get the container ID to construct the URL in the first place?  The only way to do that is if you had a stale one (i.e. the server restarted), in which case the UI will fail to reconnect.

The proper fix here is to fix the locking, which is needed anyway because the lock is being held during the initial sync (which can take arbitrarily long).  So I'd prefer we punt this.

> S1. watcherMu field comment and EngineReconciler struct doc went stale

See above; the locking here needs a re-think.

> S2. Docs promise UTF-8 message text, but the endpoint streams raw binary 

No, any text inside of the raw binary stream must be UTF-8 (because that's what xterm.js requires).  We do not support, say, the binary stream containing KOI-8 encoded text.

> S3. Container ID validator is too permissive: accepts any-length and uppercase hex

Irrelevant; it just checks that the container ID is valid enough that we have no worries using it.  The real check is fetching the container object.  That will be a 404 on an invalid container ID; that seems reasonable.  I've actually hit that in BATS, because I originally used a truncated ID (from `docker ps`) and that 404ed.

> S4. websocketWriter.Write leaks a half-open writer when the inner Write fails 

There is no `NextWriter`; when the `Write` fails, the copy exits, and we close the connection.  While the review acknowledges that, it somehow things we might… do something different in the future?  It's a log endpoint, there's nothing else it _can_ do.  There's no retry, because making the client do that makes way more sense.

> S5. HandleLogs doc comment omits the follow query parameter

Irrelevant; we're not trying to document the API endpoint completely in the godoc.  That's what the markdown is for.

> S6. GetPassthroughEndpoints sort differs between engine and mock/demo controllers

That's an existing issue (in the old controllers), doesn't make sense to fix it in this already-giant PR.

> S7. "Docker watcher not running" is misleading on the containerd backend

There _isn't_ a containerd backend yet.  This will get fixed once there is.

> S8. New BATS tests issue log requests before waiting for the Container mirror

Eh, maybe?

> S9. Negative paths for the new endpoint are untested

… If I feel like it

> S10. conn.SetReadLimit(4096) is a magic number

The log endpoint is one way, there's nothing to read (other than pong).

> S11. MarkControlPlaneReady fires before controllers start, and silently proceeds on discovery-registration failure

I don't think I've touched any of the relevant code? Why is this even here?  The handler nil check is there because we now return nil if the watcher isn't around.

---

I might fix S8/S9 if human reviews flag more things to fix, but they're not exactly worth doing if there's no other reason to.

jandubois (approved):
Thanks, LGTM

Tested with https://github.com/rancher-sandbox/rancher-desktop-app/pull/48
CleanShot 2026-04-23 at 11 52 05@2x

PR #326, opened 2026-04-17
Fix deadlock in vendored nxadm/tail (observed as Windows CI hostagent-watcher hang)

jandubois:
## Summary

Fixes a deadlock in the shared `InotifyTracker` that `github.com/nxadm/tail` inherits from its upstream design. The mechanism is platform-independent, but we have only observed it on Windows CI so far, where it reliably hangs the LimaVM controller's hostagent watcher after a few rotations and leaves the reconciler wedged forever. Most recently seen on PR #316 run [24538861468](https://github.com/rancher-sandbox/rancher-desktop-daemon/actions/runs/24538861468) (`BATS (windows-latest, bats-lima)`), where tests 95/100 timed out because the watcher for the 4th hostagent restart received zero events and `stopWatcher()` blocked on `<-state.done` forever.

## Root cause

Our `LimaVMReconciler` watches the hostagent's JSON event stream by calling `lima/pkg/hostagent/events.Watch`, which uses `github.com/nxadm/tail`. `nxadm/tail` shares a single `fsnotify.Watcher` process-wide through an `InotifyTracker` singleton. Upstream runs one goroutine that serves **four** jobs from a single `select`:

1. Synchronous `Add`/`Remove` RPCs from individual tails (blocks waiting for a reply from fsnotify's reader thread).
2. Drains `watcher.Events`.
3. Drains `watcher.Errors`.
4. Shuts down on `close`.

Every fsnotify backend has an **unbuffered `Errors` channel** (`fsnotify.go:253`: `make(chan error)`, no buffer, on every OS). Every backend also has multiple `sendError` call sites on its reader thread. Call sites, by backend:

| Backend | sendError sites | Representative triggers |
|---------|-----------------|-------------------------|
| `backend_windows.go` | 7 | `ReadDirectoryChanges` buffer overflow ("buffer larger than it is"), CancelIo/CloseHandle failures, IOCP errors |
| `backend_inotify.go` (Linux) | 5 | inotify read errors, **`ErrEventOverflow` (IN_Q_OVERFLOW)**, EINTR |
| `backend_kqueue.go` (macOS/BSD) | 3 | kevent read failure, per-watch errors |
| `backend_fen.go` (Solaris) | 4 | `port.Get` failures, unexpected event source |

Deadlock chain:

1. Tracker's `run` is inside `fsnotify.Remove(...)`, blocked on `<-in.reply`.
2. `readEvents` tries to send an error → blocks on `Errors <- err`.
3. Nobody drains `Errors` (that was `run`'s job, but it's stuck in #1).
4. `readEvents` never processes the pending input; `fsnotify.Remove` never returns.
5. Every subsequent `tail.Open` caller piles up on unbuffered `shared.watch`; every `tail.Stop` hangs in `Tomb.Wait`.

### Why we only see it on Windows

Empirical, not fundamental. Windows's `ReadDirectoryChanges` with fsnotify's 4 KB default buffer overflows easily under bursty RENAME/CREATE events; on a 2-vCPU windows-latest runner with Compat Telemetry noise, the "buffer larger than it is" error fires within a handful of rotation cycles. Linux `fs.inotify.max_queued_events` defaults to 16384 (often much higher under systemd) and our hostagent rotation rate doesn't get near it. macOS kqueue flakes haven't been reported but the BATS surface there is smaller too. The fix is structural and applies regardless of which backend emits the error.

### Evidence from the failing CI job

| time        | event                                                                  |
|-------------|------------------------------------------------------------------------|
| 23:28:33.18 | Controller logs `Hostagent started, watcher active` (pid=6704)         |
| 23:28:37.83 | hostagent writes `{sshLocalPort:22}` to ha.stdout.log                  |
| 23:28:40.96 | hostagent writes `{running:true}`                                      |
| …           | **5 minutes of complete watcher silence** — no event received           |
| 23:33:32.98 | host-switch vsock handshake 5 min timeout (watcher never cancelled it) |
| 23:33:40.37 | reconciler's `Stopping Lima instance`                                  |
| 23:33:41.84 | `Hostagent process exited cleanly`                                     |
| —           | `Watcher stopped` is **never** logged                                  |

The "buffer larger than it is" sendError was the trigger confirmed by a standalone repro.

### 1. `nxadmtail: absorb upstream watch/, util/, winfile/ subpackages`

`pkg/util/nxadmtail` previously vendored only `tail.go` (with the Pulumi/nxadm#70/#71 StopAtEOF patches) and still pulled `nxadm/tail/{watch,util,winfile}` as runtime dependencies. Copy those subpackages in-tree and rewrite imports to local paths so `nxadm/tail` moves to `// indirect` in go.mod. No behavior change; the `InotifyTracker` at this commit is still the unfixed upstream version.

### 2. `tail: rename pkg/util/nxadmtail to pkg/util/tail, merge with CLI wrapper`

The reviewer of the original vendoring PR ([#156](https://github.com/rancher-sandbox/rancher-desktop-daemon/pull/156)) had suggested `pkg/util/tail`, but that name was taken by a small CLI wrapper. Fold both into `pkg/util/tail`. Two `File` functions collided, so:

- lib constructor `File` → `Open` (`tail.Open(path, cfg)` reads as "open a tail session").
- wrapper helper `File` → `Stream` (`tail.Stream(ctx, w, path, follow)` reads as "stream the file to writer").

Only external call sites: `cmd/rdd/service.go` and `cmd/rdd/limavm.go`, updated to `tail.Stream`. Also updates the `LTAG_EXCLUDES` list in the Makefile so ltag keeps honoring the vendored MIT-licensed files at their new path.

### 3. `tail/watch: fix deadlock in the shared InotifyTracker`

Split `InotifyTracker.run()` into three goroutines:

```go
// Drain Events forever.
go func() { for ev := range t.watcher.Events { t.sendEvent(ev) } }()

// Drain Errors forever. *The key to the fix.*
go func() { for err := range t.watcher.Errors { /* log if non-EINTR */ } }()

// Handle Add/Remove RPCs. Can block on fsnotify without starving drainers.
for {
    select {
    case winfo := <-t.watch:  t.error <- t.addWatch(winfo)
    case winfo := <-t.remove: t.error <- t.removeWatch(winfo)
    }
}
```

The deadlock chain breaks at step 3: with a dedicated `Errors` drainer, `sendError` returns immediately, `readEvents` processes the pending RPC, no circular wait. **One `fsnotify.Watcher` is still shared process-wide** — Linux's `fs.inotify.max_user_instances` budget (default 128) is unaffected. HP originally chose the shared design for exactly this reason, and the fix preserves it.

Commit also adds `TestInotifyTrackerNoDeadlockOnRepeatedRotation` — a `GOMAXPROCS=1` stress test that self-execs a writer subprocess and runs 200 rotate/tail/stop cycles mirroring the LimaVM restart pattern. Reliably red on the pre-fix tracker (typically fails between cycle 2 and 30), green on the fixed tracker with the exact CI-log error (`Windows system assumed buffer larger than it is`) logged and recovered from. Marked `testing.Short()`-skippable because a full run is ~7 minutes.

Running this same test on Linux later exposed a second, distinct race in the same subsystem; see §6.

### 4. `rdd: switch hostagent watcher to local tail-based events package`

Even with our fork fixed, rdd still went through `lima/pkg/hostagent/events.Watch`, which uses the unpatched upstream `nxadm/tail`. Add `pkg/hostagent/events` with a small `Watch(ctx, stdoutPath, stderrPath, begin, onEvent)` backed by the local `pkg/util/tail`. Its `Event`/`Status` types mirror the subset of Lima's JSON schema rdd actually consumes (`Running`, `Degraded`, `Exiting`, `Errors`, `SSHLocalPort`); other fields unmarshal as zero values. The `propagateStderr` flag upstream has is dropped — rdd never set it.

`pkg/controllers/lima/limavm/controllers/hostagent_watcher.go` swaps to the local package. **After this commit rdd has no direct dependency on `nxadm/tail` and no live call path through Lima's `events.Watch`.**

### 5. `tail: resolve golangci-lint issues in the absorbed code`

The absorbed upstream files triggered 19 golangci-lint findings under the project's lint config. Fix without changing observable behavior: dissolve the `util` subpackage (package name tripped revive meaningless-name, `os.Exit` tripped deep-exit); un-export the tracker's internal API (`Watch`→`track`, `Events`→`eventsFor`, etc.); switch the EINTR check to `errors.As`/`errors.Is`; join `fallthrough` arms into comma-separated cases; drop an ineffectual re-assignment; and fix the test file's forbidden `t.Fatalf`/shadowed `max`/uncontexted `exec.Command`/perfsprint violations.

### 6. `tail/watch: synchronise watcher goroutine cleanup with Stop`

Running the stress test from §3 on Linux under `GOMAXPROCS=1` exposed a second, distinct race in the same subsystem: `TestInotifyTrackerNoDeadlockOnRepeatedRotation` reliably failed around cycle 130 with `received 0 events`. The new tail saw none of the writes its subprocess produced.

The race is a lifecycle mismatch between `tail.Stop` and the goroutine `ChangeEvents` spawns to forward fsnotify events:

1. `tail.Stop` only waits on the tomb. `tomb.Done()` fires from `tailFileSync` as soon as its loop exits.
2. The `ChangeEvents` goroutine owns the `track`/`untrack` pair on the shared `InotifyTracker`. It notices `t.Dying()` on its next `select` iteration and only then issues the synchronous `untrack` RPC into `run()`.
3. Between `tail.Stop` returning and that `untrack` executing, the tracker still reports `chans[path] != nil` and `watchNums[path] == 1` for the stopped tail.

If the next rotation cycle calls `track(path)` inside that window, `run()` may service the new watch before the stale remove. `addWatch` sees a stale `chans[path]` entry and reuses it; `removeWatch` fires next and nulls it out. The new tail's events goroutine then reads a nil channel and blocks forever, and `sendEvent` silently drops every event because `chans[path] == nil`. The inotify watch is still live in the kernel; every event vanishes on the way to the tail.

Fix: give `Tail` a `sync.WaitGroup` that tracks the `ChangeEvents` goroutine. `FileWatcher.ChangeEvents` takes `*sync.WaitGroup`, increments it before spawning, and defers `Done` in the goroutine body. `tailFileSync` defers `watchers.Wait` between `close()` and `Done()`, so `tail.Wait` (and therefore `tail.Stop`) returns only after every goroutine has finished its `untrack`. Subsequent `track()` calls see a clean tracker.

The polling watcher takes the same signature for interface symmetry — it has no `track`/`untrack` but still spawns a goroutine tied to the tail lifetime, so waiting on it is the right semantics.

`TestInotifyTrackerNoDeadlockOnRepeatedRotation` now passes reliably on Linux (`GOMAXPROCS=1`, 422 s, all 200 cycles) as well as Windows, and covers both the deadlock and the lifecycle race.

## Why not fix it upstream first

The fix is worth upstreaming, but:

- `github.com/nxadm/tail` hasn't had a release since 2019 and has open PRs years old; waiting on it would block CI indefinitely.
- `github.com/fsnotify/fsnotify`'s unbuffered `Errors` channel is the underlying hazard and should also be patched upstream — again, not on a schedule we control.

Our fork lets us unblock CI now. Sharing the fixed code with Lima is out of scope for this PR; once Lima converges on a patched `nxadm/tail` or we coordinate a joint fork, we can consolidate.

## Resource impact

| Platform | Before        | After         | Notes |
|----------|---------------|---------------|-------|
| Linux    | 1 inotify instance per process | 1 inotify instance per process | `fs.inotify.max_user_instances` unchanged |
| macOS    | 1 kqueue fd + N file fds     | same          | —     |
| Windows  | 1 IOCP thread | 1 IOCP thread + 2 tiny drainer goroutines | goroutines are free |

## Test plan

- [ ] `go build ./...` passes.
- [ ] `go vet ./...` passes.
- [ ] `go tool golangci-lint run ./...` reports 0 issues.
- [ ] `go test ./pkg/util/tail/... ./pkg/hostagent/... ./pkg/controllers/lima/...` passes with `-short` (fast unit tests).
- [ ] `go test -run TestInotifyTrackerNoDeadlockOnRepeatedRotation -timeout 1800s ./pkg/util/tail/` passes (~7 minutes) on both Linux and Windows. Locally verified pre-fix red on Windows (fails cycle 2–30 on the deadlock) and on Linux with `GOMAXPROCS=1` (fails ~cycle 130 on the lifecycle race), and post-fix green on both across repeated runs.
- [ ] CI: `BATS (windows-latest, bats-lima)` job passes without `33-lima-controllers` watcher timeouts across multiple runs.
- [ ] `rdd service log -f`, `rdd limavm log -f` still work (exercise `tail.Stream`).

## Out of scope

- Upstreaming the fix to `github.com/nxadm/tail` or `github.com/fsnotify/fsnotify`.
- Converting Lima's `hostagent/events.Watch` to use the fixed code (we just bypass it).


jandubois:
Closing: the head ref points at jandubois/rancher-desktop-daemon-old after a visibility flip severed the cross-repo fork link. Reopening from the new fork next.

PR #328, opened 2026-04-18
engine: port docker_context to docker/cli ContextStore

jandubois:
Replace the hand-rolled `~/.docker/contexts` writers with the upstream `github.com/docker/cli` `ContextStore` API and `cli/config.ConfigFile`. The `contexts/meta` layout, atomic writes, and on-disk schema now track Docker CLI's definitions, and the port drops ~145 lines across the file and its tests.

Saving through docker/cli's typed `ConfigFile` drops unknown top-level keys from `config.json` and injects empty `auths` and `credsStore` fields. `docker context use` does the same today, so users of Docker CLI already see this behavior.

Lima's `cilium/ebpf` transitive dep already pulled `docker/cli` into `go.mod` as indirect; promoting it to direct adds 4 new indirect entries and ~161 KiB to the stripped `rdd` binary. The store's context-metadata `TypeGetter` takes `nil` and decodes into `map[string]any`, which lets us skip `cli/command` and its metrics + telemetry chain (`otel/sdk/metric`, `go-metrics`, `gorilla/mux`, `backoff/v5`, `morikuni/aec`). `app-controller` stays unchanged because it does not import the `engine` package.


jandubois:
AI review of preview version at https://rancher-sandbox.github.io/rancher-desktop-daemon/20260420-095326-pr-328.html

I've addressed most of the suggestions/testgaps, so the line savings is not that large anymore, but most of the new lines are tests.

Nino-K (approved):
I will modify my PR after this gets merged. I made the probe result for `ssh://`, `npipe://`, and unknown schemes distinguish “unprobeable” from “healthy.” Treating them all as healthy might be problematic.

PR #331, opened 2026-04-21
engine: extract engineWatcher interface from dockerWatcher

jandubois:
Prepares the reconciler for a second watcher implementation (containerd) alongside the Docker one. `EngineReconciler.watcher` and the three `processXFinalizers` helpers now take the interface. A compile-time assertion keeps `*dockerWatcher` honest.

Type-level refactor only; no behavior change.

### DIFF
--- a/pkg/controllers/app/engine/controllers/engine_reconciler.go
+++ b/pkg/controllers/app/engine/controllers/engine_reconciler.go
@@ -53,8 +53,30 @@ const (
 	engineMoby = "moby"
 )
 
+// engineWatcher is the reconciler-facing contract every container-engine
+// watcher must satisfy. dockerWatcher is the only current implementation; a
+// forthcoming containerd watcher will provide a second. Methods on the
+// watcher that the reconciler does not call (event handlers, full-sync
+// internals) stay off the interface.
+type engineWatcher interface {
+	// alive reports whether the watcher goroutine is still running.
+	alive() bool
+	// stop cancels the watcher goroutine and waits for it to finish.
+	stop()
+	// processContainerAction dispatches the action annotation on a single
+	// Container mirror and records the outcome in status.lastAction.
+	processContainerAction(ctx context.Context, c *containersv1alpha1.Container) error
+	// deleteContainer removes a container from the engine by ID.
+	deleteContainer(ctx context.Context, id string) error
+	// deleteImage removes an image from the engine.
+	deleteImage(ctx context.Context, img *containersv1alpha1.Image) error
+	// deleteVolume removes a volume from the engine by name. Engines
+	// without a native volume concept return nil.
+	deleteVolume(ctx context.Context, name string) error
+}
+
 // EngineReconciler watches the App resource for the Running condition and
-// manages a Docker watcher goroutine that mirrors engine state into K8s.
+// manages an engine watcher goroutine that mirrors engine state into K8s.
 //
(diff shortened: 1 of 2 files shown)


pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -53,8 +53,30 @@ const (
 	engineMoby = "moby"
 )
 
+// engineWatcher is the reconciler-facing contract every container-engine
+// watcher must satisfy. dockerWatcher is the only current implementation; a
+// forthcoming containerd watcher will provide a second. Methods on the
+// watcher that the reconciler does not call (event handlers, full-sync
+// internals) stay off the interface.
```
mook-as: For logs, I will need inspect, start streaming logs (to a writer), and eventually start exec (to a ReadWriter) in the future; but we can think about that at that point.

But that is basically, should this be the engine watcher, or the engine abstraction? Don't actually change the implementation in this PR, though.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -53,8 +53,30 @@ const (
...
+type engineWatcher interface {
+	// alive reports whether the watcher goroutine is still running.
```
mook-as:
```suggestion
	// alive reports whether the watcher is still running.
```
The fact that there's a goroutine involved isn't really relevant to the API consumers.  Same for `stop()`.

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -53,8 +53,30 @@ const (
...
+	alive() bool
+	// stop cancels the watcher goroutine and waits for it to finish.
+	stop()
+	// processContainerAction dispatches the action annotation on a single
+	// Container mirror and records the outcome in status.lastAction.
```
mook-as: Somehow I had to read that three times to understand it.  Consider:

> processContainerAction dispatches the action on the [containersv1alpha1.AnnotationAction] annotation on a Container, and records the outcome in status.lastAction.

I'm not actually sure how much better that is…

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -53,8 +53,30 @@ const (
...
+	processContainerAction(ctx context.Context, c *containersv1alpha1.Container) error
+	// deleteContainer removes a container from the engine by ID.
+	deleteContainer(ctx context.Context, id string) error
```
mook-as: Any reason this doesn't take a `*containersv1alpha1.Container` instead of a string? Either is acceptable, just curious.

mook-as (dismissed):
As expected, this is basically fine.

pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: You are right, the interface will eventually cover more than the watcher. I've renamed it to `engine`.

However I left all the `watcher` fields and variables as-is for.

pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: Because we only need the name?But Image needs `ID` and `RepoTag`, so I changed every signature to pass the structure, for symmetry.

mook-as (approved):

PR #333, opened 2026-04-21
Docker Engine: improve Docker context probe safety and isolation

Nino-K:
This PR extends the use of docker/cli `ContextStore` to probing, replace `getDockerContextHost` + `probeDockerContext` with `probeNamedDockerContext`, leveraging `WithTLSData` and `ep.ClientOpts()` to correctly handle TCP+TLS, mTLS, and `SkipTLSVerify` contexts; treat SSH/npipe as healthy to avoid false overrides.

To elaborate further, `probeDockerContext` only calls `dockerclient.New(dockerclient.WithHost(host))`. That means:

   - TCP+TLS context: host is `tcp://192.168.1.1:2376`, no TLS certs → TLS handshake fails → probe returns false → context falsely treated as unhealthy → overridden with ours
   - mTLS context: same, no client cert → server rejects connection → false override
   - SkipTLSVerify=true: context works fine for the user, but probe fails without the right client config → false override
   - SSH context: `ssh://user@host` passed to WithHost → dial fails → false override

### DIFF
--- a/pkg/controllers/app/engine/controllers/docker_context.go
+++ b/pkg/controllers/app/engine/controllers/docker_context.go
@@ -8,13 +8,16 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/containerd/errdefs"
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/context/docker"
 	"github.com/docker/cli/cli/context/store"
-	mobyclient "github.com/moby/moby/client"
+	dockerclient "github.com/moby/moby/client"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // dockerContextProbeTimeout is the maximum time allowed to ping a Docker
@@ -46,7 +49,8 @@ func newContextStore() (store.Store, error) {
 	// Pass nil for the context-metadata TypeGetter; the store decodes a nil
 	// type into a map[string]any, which is all we need since we never inspect
 	// the Metadata field. This avoids importing cli/command, which would pull
-	// in the docker/cli metrics + telemetry chain (otel/sdk/metric, etc.).
+	// in the docker/cli metrics + telemetry chain (otel/sdk/metric, go-metrics,
+	// gorilla/mux, backoff/v5, morikuni/aec).
 	cfg := store.NewConfig(
 		nil,
 		store.EndpointTypeGetter(docker.DockerEndpoint, func() any { return &docker.EndpointMeta{} }),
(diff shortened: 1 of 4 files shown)


pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -8,13 +8,16 @@ import (
...
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/containerd/errdefs"
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/context/docker"
 	"github.com/docker/cli/cli/context/store"
-	mobyclient "github.com/moby/moby/client"
+	dockerclient "github.com/moby/moby/client"
```
mook-as, line 18: Please rebase and correct the merge; this will have spell check issues (unless spell check broke again).

pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -358,20 +362,29 @@ func (r *EngineReconciler) manageDockerContext(socketPath string) {
...
+			// would overwrite a file we cannot safely round-trip.
 			log.Error(err, "Failed to read current Docker context")
+			return
+		}
+
+		// Docker CLI resolves context env-first: DOCKER_HOST overrides,
+		// then DOCKER_CONTEXT, then config.json (currentContext).
+		// getCurrentDockerContext only reads config.json, so check
+		// DOCKER_CONTEXT to avoid probing the wrong endpoint and
+		// overwriting a user-selected context.
```
mook-as: Would it make more sense to never set current context in the JSON file if `DOCKER_CONTEXT` is set, then? I'm think about the case where:
- `DOCKER_CONTEXT` is only set in the shell that `rdd` started in, not globally (at a user level)
- `DOCKER_CONTEXT` is `foo`
- JSON has `currentContext: bar`
- `foo` is broken, `bar` is fine.

The current flow would set `current` to `foo`, see that it's broken, and update the JSON file to `rancher-desktop`, but `bar` would have worked… Basically, I'm not sure `DOCKER_CONTEXT` is relevant for the code where we update the JSON.

It's possible that I'm missing the context that made this block of code make more sense.

pkg/controllers/app/engine/controllers/docker_context.go
```diff
@@ -149,16 +129,73 @@ func clearCurrentDockerContext(name string) error {
...
+// is unreadable, the endpoint metadata is missing or malformed, the scheme is
+// not tcp/unix (e.g. SSH), or the TLS config cannot be loaded — we return true
+// (assume healthy). This conservatism prevents RDD from clobbering a context
+// that is in the middle of being edited or migrated.
+func probeNamedDockerContext(ctx context.Context, name string) bool {
+	log := logf.FromContext(ctx).WithName("docker-context").WithValues("context", name)
+
+	s, err := newContextStore()
+	if err != nil {
+		log.V(1).Error(err, "Cannot open context store; assuming context healthy")
```
mook-as: Per the [documentation](https://pkg.go.dev/github.com/go-logr/logr#Logger.Error), `V(1)` here is pointless:

> The log message will always be emitted, regardless of verbosity level. 

(I think that's bad API, but we work with what we have.)

mook-as (changes requested):

mook-as (approved):
The spellcheck seems to pass. I don't understand why, given that it complained so hard before…

jandubois:
AI review at https://symmetrical-adventure-qjle1je.pages.github.io/20260424-130819-pr-333.html claims one missed `DOCKER_CONTEXT` override in (I1).

Nino-K:
> AI review at https://symmetrical-adventure-qjle1je.pages.github.io/20260424-130819-pr-333.html claims one missed `DOCKER_CONTEXT` override in (I1).

It's already been addressed in a better way. The current fix that I applied is more conservative in the sense that if `DOCKER_HOST` or `DOCKER_CONTEXT` is set in the daemon's env, we skip the probe and set entirely. This closes the exact false override path that I1 describes (the user never gets their config.json overwritten) without the
  risk of treating a local shell env var as representative of the user's global preference.

PR #334, opened 2026-04-21
bats: bound per-attempt timeouts and capture guest logs on Windows hangs

jandubois:
Two independent diagnostics/reliability fixes, motivated by the Windows BATS hang on `fix-lima-ssh-port` (run 24707925880 / job 72265349580) where the app controllers suite wedged for 25+ minutes at the \"wait for dockerd to be ready\" test.

- **`try --per-try-timeout DUR`** wraps each attempt in `timeout(1)` with a 5s SIGTERM->SIGKILL grace window. Stops a single hung ssh session from burning the full retry budget when the remote command blocks on an unresponsive dockerd socket (ssh keepalives hide the hang from `ConnectTimeout` and `ServerAliveInterval`). Applied to the \"wait for dockerd to be ready\" test in `bats/tests/32-app-controllers/app.bats`, the first confirmed failure mode.

- **`dump_windows_vm_logs`** in `scripts/bats-with-timeout.sh` collects, for every running `lima-*` WSL distro: `systemctl status rancher-desktop.target` as a compact summary, `ps auxf` as a process snapshot, and the full boot journal via `journalctl -b --no-pager` (no line cap, matching the macOS/Linux serial.log capture). `wsl.exe` talks to the WSL service directly, so the dump survives a wedged rdd. Every invocation runs under `timeout(1)`; the function returns early on non-Windows hosts.

### DIFF
--- a/bats/helpers/utils.bash
+++ b/bats/helpers/utils.bash
@@ -272,6 +272,10 @@ trace() {
 # try runs the specified command until it either succeeds, or --max attempts
 # have been made (with a --delay seconds sleep in between).
 # With --until-fail, waits until the command fails instead of succeeds.
+# With --per-try-timeout DUR, each attempt is wrapped in `timeout(1)` so a
+# single hung call cannot stall the whole retry budget; DUR accepts any
+# timeout(1) suffix (s, m, h). The attempt exits non-zero (124 on SIGTERM,
+# 137 on SIGKILL) and counts like any other failure toward --max.
 #
 # Right now the command is **always** run with --separate-stderr, and stderr
 # is output after all of stdout. This is subject to change, if we can figure
@@ -279,6 +283,7 @@ trace() {
 try() {
     local max=24
     local delay=5
+    local per_try_timeout=""
     local until_fail=0
 
     while [[ $# -gt 0 ]] && [[ $1 == -* ]]; do
@@ -289,6 +294,10 @@ try() {
             ;;
         --delay)
             delay=$2
+            shift
+            ;;
+        --per-try-timeout)
+            per_try_timeout=$2
             shift
             ;;
         --until-fail)
@@ -308,7 +317,14 @@ try() {
 
     local count=0
     while true; do
-        run_e "$@"
+        if [[ -n "${per_try_timeout}" ]]; then
+            # --kill-after=5s promotes SIGTERM to SIGKILL if the command
+            # ignores the graceful signal; the grace window is short so
+            # pathological cases still cannot stall the retry loop.
+            run_e timeout --kill-after=5s "${per_try_timeout}" "$@"
+        else
+            run_e "$@"
+        fi
         local success
         if ((until_fail)); then
             success=$((status != 0))
--- a/bats/helpers/utils.bash
+++ b/bats/helpers/utils.bash
@@ -318,10 +318,12 @@ try() {
     local count=0
     while true; do
         if [[ -n "${per_try_timeout}" ]]; then
-            # --kill-after=5s promotes SIGTERM to SIGKILL if the command
-            # ignores the graceful signal; the grace window is short so
-            # pathological cases still cannot stall the retry loop.
-            run_e timeout --kill-after=5s "${per_try_timeout}" "$@"
+            # --verbose logs "sending signal TERM" to stderr so the try
+            # log distinguishes a timeout kill from a plain non-zero
+            # exit. --kill-after=5s escalates SIGTERM to SIGKILL when
+            # the command ignores the graceful signal, so a hung retry
+            # cannot stall the loop.
+            run_e timeout --verbose --kill-after=5s "${per_try_timeout}" "$@"
         else
             run_e "$@"
         fi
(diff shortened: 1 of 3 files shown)


bats/helpers/utils.bash
```diff
@@ -308,7 +317,14 @@ try() {
 
     local count=0
     while true; do
-        run_e "$@"
+        if [[ -n "${per_try_timeout}" ]]; then
+            # --kill-after=5s promotes SIGTERM to SIGKILL if the command
+            # ignores the graceful signal; the grace window is short so
+            # pathological cases still cannot stall the retry loop.
+            run_e timeout --kill-after=5s "${per_try_timeout}" "$@"
```
mook-as: Do we want `--verbose` here too, to have something in the log that we're timing out?

```
$ timeout 1s sleep inf ; echo $? ; timeout --verbose 1s sleep inf ; echo $?
124
timeout: sending signal TERM to command ‘sleep’
124
```

scripts/bats-with-timeout.sh
```diff
@@ -200,6 +200,58 @@ dump_memory_pressure() {
...
+# only that ssh is stuck; the actual cause lives in the VM's journal.
+# wsl.exe talks to the WSL service directly and does not depend on the
+# rdd daemon, so this dump still works after rdd is wedged or killed.
+# Every invocation is behind `timeout` so a hung guest cannot block the
+# bundle capture itself.
+dump_windows_vm_logs() {
+    case "$(uname -s)" in
+        MINGW*|MSYS*|CYGWIN*) ;;
+        *) return ;;
+    esac
```
mook-as: Do we need this check, if we check that `wsl.exe` is available right after?

scripts/bats-with-timeout.sh
```diff
@@ -200,6 +200,58 @@ dump_memory_pressure() {
...
+    command -v wsl.exe >/dev/null 2>&1 || return
+
+    # `--quiet --list --running` prints one distro name per line, but WSL
+    # emits UTF-16LE (with a BOM) on some Windows builds. Strip NULs and
```
mook-as: Just export `WSL_UTF8=1`; that was introduced in [2022](https://github.com/microsoft/WSL/releases/tag/0.64.0).  We're not going to support anything that old.

scripts/bats-with-timeout.sh
```diff
@@ -200,6 +200,58 @@ dump_memory_pressure() {
...
+
+    local distro
+    for distro in $distros; do
+        echo
+        echo "=== VM: ${distro} ==="
+
+        echo
+        echo "--- ${distro}: systemctl status rancher-desktop.target ---"
+        timeout --kill-after=1 10 \
+            wsl.exe -d "$distro" -- systemctl status rancher-desktop.target --no-pager 2>&1 || true
```
mook-as: It'll probably be easier to read if we do the line break after the `--` (so the second line is the `systemctl …` command).

scripts/bats-with-timeout.sh
```diff
@@ -200,6 +200,58 @@ dump_memory_pressure() {
...
+
+        echo
+        echo "--- ${distro}: ps auxf ---"
+        timeout --kill-after=1 10 \
+            wsl.exe -d "$distro" -- ps auxf 2>&1 || true
+
+        # Dump the full journal for this boot. Lima VMs live only as
+        # long as the bats target, so the journal is bounded by the
+        # run's wall time; no caps to match the macOS/Linux serial.log
```
mook-as: > … no caps to match …

What's _caps_ in this context? Capabilities?

mook-as (changes requested):

scripts/bats-with-timeout.sh
jandubois: It was meant to be "limits", but I tightened the whole comment and got rid of it.

mook-as (approved):

scripts/bats-with-timeout.sh
```diff
@@ -200,6 +200,54 @@ dump_memory_pressure() {
...
+# On Windows, Lima runs each VM as a WSL2 distro named `lima-<vmname>`.
+# When a bats target hangs waiting on the guest (e.g. `docker info`
+# blocked on an unresponsive dockerd socket), the host-side logs show
+# only that ssh is stuck; the actual cause lives in the VM's journal.
+# wsl.exe talks to the WSL service directly and does not depend on the
+# rdd daemon, so this dump still works after rdd is wedged or killed.
+# Every invocation is behind `timeout` so a hung guest cannot block the
+# bundle capture itself.
+dump_windows_vm_logs() {
+    command -v wsl.exe >/dev/null 2>&1 || return
```
mook-as:
```suggestion
    command -v wsl.exe >/dev/null 2>&1 || return 0
```

That is causing the test failures.

mook-as (changes requested):
CI fails on not-Windows.

scripts/bats-with-timeout.sh
jandubois: Thanks, amended!

mook-as (approved):

PR #336, opened 2026-04-22
cli: standardize --wait and --timeout across svc and limavm

### LINKED ISSUE #196: `rdd lima start/stop/delete` commands all should have a `--wait=false` option

jandubois:
Every wait-capable subcommand under `rdd svc` and `rdd limavm` now defaults to `--wait=true` with a finite `--timeout`, and exits with code 4 on deadline expiry to match `rdd set`.

Previously the defaults were inconsistent: `rdd limavm delete` shipped with `--wait=false`; `rdd svc start/stop` exposed `--wait` but no `--timeout`, with 90s / 60s hardcoded in the service package; and `rdd svc delete` exposed neither flag. On timeout, every command exited 1. Users who scripted around the CLI had no way to distinguish "did not settle in time" from a real error.

## Default timeouts

- **90s:** `svc start` (CRD establishment + controller-manager registration)
- **5m:**  `svc stop`, `svc delete`, and every `limavm create/start/stop/restart/delete`

`--timeout=0` on any command waits indefinitely.

`svc delete` drops `--wait=false` entirely: removing `instance.Dir()` under a live process corrupts the directory on Windows and breaks PID-file mutual exclusion on Unix. `--timeout` still bounds the embedded shutdown wait. `Delete` also re-checks `Running()` on a stop failure so a process that exits in the race window between the outer and inner check no longer blocks deletion.

`ensureServiceRunning` routes `context.DeadlineExceeded` through `timeoutError`, so the exit-code-4 contract holds for every command that autostarts the service (`rdd set`, `rdd limavm *`, `rdd kubectl`, `rdd service config`). `serviceStartAction`'s already-running branch now waits for the discovery ConfigMap's `ReadyAnnotation`, matching the cold-start path.

## Library surface

- `service.StopWithWait(wait, timeout)` replaces the single-arg version and drives the wait off a `context.WithTimeout`.
- `service.Delete(timeout)` replaces `Delete()` and always waits for the process to exit before removing state.
- `service.Stop()` and `service.WaitWithTimeout()` are gone; the CLI was their only caller and now builds its own timed context.

`cmd/rdd/*` routes `context.DeadlineExceeded` (including wrapped forms) through a small `timeoutError` helper that tags the error with `cliexit.Timeout` at the CLI boundary.

## Documentation

- `docs/design/cmd_service.md` documents `--wait=false` and `--timeout`, names SIGINT as the shutdown signal, explains the safety reason for removing `--wait=false` on `svc delete`, and cross-references `svc stop`.
- `docs/design/cmd_lima.md` matches the new 5m defaults and the `--wait=true` default on `limavm delete`.
- `docs/design/api_service.md` shutdown signal description matches the implementation (SIGINT on Unix, `CTRL_BREAK_EVENT` on Windows).

## Tests

- `bats/tests/33-lima-controllers/limavm-cli.bats` adds `lima create --start --timeout` and `lima restart --timeout` exit-code-4 assertions.
- `bats/tests/20-service/6-timeout.bats` asserts `svc stop --timeout=1ms` exits with code 4 and emits `context deadline exceeded`.


jandubois:
This is round 8 of AI reviews now: https://rancher-sandbox.github.io/rancher-desktop-daemon/20260422-170229-pr-336.html

I don't want to do any more rounds. I will deal with the main issues in a follow-up PR that will move the Discovery ConfigMap cleanup from the CLI code into the daemon itself, which will hopefully get rid of some of the race scenarios.

PR #338, opened 2026-04-22
Check-spelling: Update to v0.0.26

mook-as:
This fixes CI to actually fail on spelling errors; the second commit then fixes all of the cases we have.


jandubois (approved):
Thanks, LGTM

Impressed that you actually changed the sources to make them more readable instead of just adding all the words to the exception list...

PR #339, opened 2026-04-23
api: containers: make ContainerPort.Bindings optional

jandubois:
- Flip `ContainerPort.Bindings` from `+required` to `+optional` so the Container mirror can surface exposed but unpublished ports (e.g. nginx's `EXPOSE 80` without `-p` at runtime). Docker returns these as `NetworkSettings.Ports={"80/tcp":null}`; the engine emits a `ContainerPort` with an empty `bindings` slice, which the prior schema rejected under SSA, leaving the mirror frozen at `status=created`.

- The `Ports` field already documents itself as "exposed ports of the container", so requiring bindings on each entry contradicted the documented intent. Clients wanting only published ports can filter by `len(bindings) > 0`.

- Regenerate CRD yaml + apply configuration.

### DIFF
--- a/bats/tests/32-app-controllers/engine-docker.bats
+++ b/bats/tests/32-app-controllers/engine-docker.bats
@@ -170,6 +170,44 @@ do_websocket() { # endpoint
         container/"${cid}" --timeout=30s
 }
 
+@test "ports mirror with and without host bindings" {
+    # nginx has EXPOSE 80 in its Dockerfile. Running it without -p
+    # yields NetworkSettings.Ports={"80/tcp":null} — an exposed but
+    # unpublished port. Running it with -p 80 adds IPv4 and IPv6 host
+    # bindings. The mirror must surface both cases: the port name
+    # always appears, and bindings appear only when the port is
+    # published.
+    run_e -0 docker run -d --name test-exposed nginx
+    exposed_cid=${output}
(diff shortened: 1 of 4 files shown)


mook-as (dismissed):
How does the user connect to that port then?

Of course, if the container is created but not running, not  having a port makes sense, so we should still allow for it either way.

Approving, but there's a merge conflict.

jandubois:
> How does the user connect to that port then?

They obviously can't. But the field description said: "Ports describes the exposed ports of the container." Which includes ports that are exposed but not bound/published.

It is easy to filter out unbound ports from the list. Otherwise if we ever wanted to mirror the `ExposedPorts` list, we would have to add another field.

It is not clear to me how much info from `docker inspect` we want to eventually mirror.

> Of course, if the container is created but not running, not having a port makes sense, so we should still allow for it either way.

It will be interesting when we ever want to create/modify a container from the GUI. But in that case we would probably fetch the list of exposed ports from the image.

> Approving, but there's a merge conflict.

Rebased on latest `main`.

mook-as (approved):

jandubois:
So the only use-case I can come up with for listing unbound ports is when you are debugging, and you look at the container, and you see: "hmm, it exposes port 4711, I wonder what that is for, because I'm not using it at all".

PR #341, opened 2026-04-23
Windows Docker Socket Forwarding via Socket Bridge

### LINKED ISSUE #157: Figure out docker socket forwarding on Windows
Currently we don't have any way to do the docker socket forwarding (to `npipe://./docker_engine`) in Lima.
We should figure something out and upstream it if possible.

Nino-K:
On Windows (WSL2/Hyper-V), there was no path for the Docker CLI to reach dockerd inside the Lima VM. The Docker CLI on Windows expects `npipe://./pipe/docker_engine`, but nothing was listening on that pipe, and no forwarding existed between the host and the VM. macOS/Linux already works via Lima's built-in SSH portForwards.

To tackle this, the following solution is done as part of this PR:

A thin, raw TCP socket bridge composed of two parts:

  Host side (this PR) — a new HostBridge that runs inside the host-switch goroutine (which already has the Hyper-V VM GUID from the vsock handshake):

   - Listens on `\\.\pipe\docker_engine`
   - For each connection, dials vsock port 6660 to the VM
   - Proxies raw bytes bidirectionally with half-close semantics

  Guest side — a new rdd-guest binary (built and deployed via rancher-desktop-opensuse):

   - Listens on `AF_VSOCK` port 6660 inside the VM
   - Forwards each connection to `/var/run/docker.sock`
   - Started by systemd as part of rancher-desktop.target

Because the bridge operates at the raw byte level (not HTTP), `docker exec -it` and `docker logs -f` work correctly without any HTTP 101 upgrade handling, the stream is transparent.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/157

### DIFF
--- a/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
+++ b/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
@@ -24,6 +24,8 @@ import (
 	"github.com/linuxkit/virtsock/pkg/hvsock"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/windows/registry"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/socketbridge"
 )
 
 // hostSwitchPlatform holds the host-switch state for Windows. On non-Windows
@@ -160,7 +162,7 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
 		return
 	}
 
-	ln, err := vsockHandshake(ctx, logger)
+	ln, vmGUID, err := vsockHandshake(ctx, logger)
 	if err != nil {
 		logger.Error(err, "Vsock handshake failed")
 		return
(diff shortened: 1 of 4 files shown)


jandubois:
I assume this is waiting for the guest agent changes to land, so we can enable the docker tests on Windows to cover the functionality of this PR.

pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -190,6 +192,18 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
...
 	g, ctx := errgroup.WithContext(ctx)
 
+	// Start the host-side socket bridge now that we have the VM GUID.
+	// It listens on the Docker named pipe and forwards each connection to
+	// rdd-guest inside the VM via vsock port 6660.  rdd-guest is baked into
+	// the VM image (via rancher-desktop-opensuse) and started by systemd.
+	g.Go(func() error {
+		bridge := socketbridge.NewDockerHostBridge(vmGUID, logger)
+		if err := bridge.Run(ctx); err != nil {
+			logger.Error(err, "Socket bridge exited with error")
```
mook-as, line 202: Do we ever need to restart the bridge? I guess we don't because this is post-handshake?

pkg/socketbridge/host_windows.go
```diff
@@ -0,0 +1,119 @@
...
+	defer pipeConn.Close()
+
+	vsockConn, err := dialVsock(b.vmGUID, b.vsockPort)
+	if err != nil {
+		b.log.Error(err, "Failed to dial vsock", "port", b.vsockPort)
+		return
+	}
+
+	// Attempt half-close path first; fall back to plain bidirectional copy if
+	// either side does not implement CloseWrite (e.g. named-pipe connections).
```
mook-as: The parenthetical part makes it sound like named-pipe connections do _not_ implement `CloseWrite`, which sounds wrong given the comments elsewhere.

pkg/socketbridge/proxy.go
```diff
@@ -0,0 +1,63 @@
...
+			mu.Unlock()
+		}
+	}
+
+	forward := func(dst, src HalfCloser) {
+		defer wg.Done()
+		_, err := io.Copy(dst, src)
+		record(err)
+		// Half-close the write side of dst so the remote peer sees EOF.
+		_ = dst.CloseWrite()
```
mook-as: Shouldn't this be `record(dst.CloseWrite())` (in case it's the `Close` that triggers an unexpected error)?

mook-as (dismissed):
Basically fine, up to you if the comments needs addressing.

mook-as (approved):

PR #348, opened 2026-05-03
rdd kubectl: embed kuberlr-style version resolver

jandubois:
When `rdd kubectl` runs against an external cluster more than +/-1 minor out of skew with the embedded kubectl, the resolver fetches a version-matched kubectl from dl.k8s.io into a shared cache, sha512-verifies it, and execs it in place. Within skew, or when the cluster is unreachable, control falls through to the embedded kubectl. Modeled on github.com/flavio/kuberlr; everything functional there sits under internal/ (only build-time version constants are exported), so Go forbids importing it and this commit reimplements the resolver inline in pkg/kuberlr/.

`rdd ctl` always targets the embedded apiserver. The embedded apiserver and the embedded kubectl come from the same k8s.io/kubernetes module entry, so they cannot drift. `ctlAction` calls `kuberlr.SkipResolver()` before `kubectlAction` so the probe never runs.

The cache lives under os.UserCacheDir() and is shared across instances:

```
  macOS:   ~/Library/Caches/rancher-desktop/kubectl/<os>-<arch>/
  Linux:   ~/.cache/rancher-desktop/kubectl/<os>-<arch>/
  Windows: %LocalAppData%\rancher-desktop\kubectl\<os>-<arch>\
```

`rdd svc paths kubectl_cache` reports the path. Two env vars override the resolver's I/O for tests and offline operation:

```
  RDD_CACHE_DIR        rdd-wide cache root; kuberlr appends
                       kubectl/<os>-<arch>/ inside it. Anticipates a
                       shared rdd cache that future consumers (k3s
                       images, lima templates) can join.
  RDD_KUBECTL_MIRROR   alternate release mirror (default
                       https://dl.k8s.io). Used by the BATS tests below.
```

Error policy: download failures (mirror 404, sha512 mismatch, network) surface loudly with non-zero exit and a "resolving kubectl version: ..." message. Falling back to a version-mismatched embedded kubectl would hide the failure behind weird kubectl errors. Apiserver-probe failures (unreachable cluster, missing or malformed kubeconfig, unparseable server version) fall through silently: serverVersion returns ok=false with nil error, Resolve returns "", and the embedded kubectl handles client-only commands. The three routine paths log at debug; the unparseable-version path logs at warn because an apiserver answering /version with garbage is a surprise worth surfacing.

Binary delta: +16 KB stripped. The only new direct dep is github.com/blang/semver/v4, already a transitive dep.

A TODO marks an offline mode (suppress downloads, fall back to closest cached binary) layered on top of the same env vars.

Coverage:
- pkg/kuberlr/resolver_test.go: skew rule (TestCompatible, with EKS/GKE/AKS gitVersion suffixes), kubectl connection-override flag binding (TestParseKubeConfigFlags, full genericclioptions.ConfigFlags surface), client-only detection (TestIsClientOnly), and cache-path layout (TestCachedPath).
- bats/tests/10-cli/7-kubectl-cache.bats with sibling fake-kube/: a standalone Go module providing a fake kubectl (prints args) and a fake server playing both apiserver /version and release mirror /release/* roles. Five tests cover download-on-miss, skip-on-hit, mirror-404, sha-mismatch, and apiserver-500 silent fall-through. The server reads /version overrides from per-test files so tests steer behavior without restarting it.
- bats/tests/10-cli/5-paths.bats: kubectl_cache added to ALL_KEYS.

### DIFF
--- a/cmd/rdd/service_paths.go
+++ b/cmd/rdd/service_paths.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/kuberlr"
 )
 
 func instancePaths() map[string]string {
@@ -28,6 +29,8 @@ func instancePaths() map[string]string {
 		"pid_file":      instance.PIDFile(),
 		"args_file":     instance.ArgsFile(),
 		"docker_socket": instance.DockerSocket(),
+		"cache_dir":     kuberlr.CacheRoot(),
+		"kubectl_cache": kuberlr.CacheDir(),
 	}
 }
 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,9 @@ updates:
   cooldown:
     default-days: 7
 - package-ecosystem: gomod
-  directory: /
+  directories:
+  - /
+  - /bats/tests/10-cli/fake-kube
   schedule:
     interval: daily
   cooldown:
--- a/bats/helpers/os.bash
+++ b/bats/helpers/os.bash
@@ -96,11 +96,26 @@ sudo_needs_password() {
 # Cross-platform process management.
 # MSYS2's kill cannot reach Win32 processes, so use taskkill.exe on Windows.
 
+# Translate an MSYS PID to its Win32 PID. taskkill and tasklist filter
+# on the Win32 PID, but $! in MSYS bash returns the MSYS PID, and the
+# two differ for native Win32 children. /proc/<pid>/winpid holds the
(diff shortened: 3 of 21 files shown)


cmd/rdd/service_paths.go
```diff
@@ -28,6 +29,7 @@ func instancePaths() map[string]string {
 		"pid_file":      instance.PIDFile(),
 		"args_file":     instance.ArgsFile(),
 		"docker_socket": instance.DockerSocket(),
+		"kubectl_cache": kuberlr.CacheDir(),
```
mook-as, line 33: Would it be possible to expose `RDD_CACHE_DIR` here too?

jandubois:
Closing: the head ref points at jandubois/rancher-desktop-daemon-old after a visibility flip severed the cross-repo fork link. Reopening from the new fork next.

Issue #349, opened 2026-05-04
Handle client version skew


mook-as:
We need to figure out the policy when the `rdd` client is a different version than the server (i.e. when the user accidentally uses a client that isn't the same executable that is running the actual daemon).

We should probably start extremely restrictive, so we can relax the requirements down the road if that seems useful.

Initial thoughts:

- We could check `/version` API endpoint and abort if `.gitVersion` does not match
- We could disable this check when developer mode is enabled (so testing changes locally does not require restarting the daemon)
- This probably means we probably need to replace the bits in `~/.rd2/bin` on daemon start.

jandubois:
We may do something more elaborate in the future, but for now fixed by #351

PR #351, opened 2026-05-07
cmd: Abort on server version mismatch

mook-as:
When waiting for the server to be ready, abort if the server version does not match the client version exactly.  This prevents potential issues when the user has two versions of Rancher Desktop Daemon installed, and used the wrong one to interact with a running server.

We skip the check completely in developer mode, to make it easier to work on the client without restarting the server unnecessarily.

See: #349


jandubois (approved):
Thanks, LGTM

I wish we had a test for this, but I can't think of an easy way to test this.

PR #353, opened 2026-05-13
rdd: Support emitting JSON logs

mook-as:
This adds a command line option to cause the logs to be emitted in JSON; this is intended to make it machine-readable.

This is to be used by the frontend, so when issues happen such as #351 triggering, the UI can display the error better.

### DIFF
--- a/cmd/rdd/main.go
+++ b/cmd/rdd/main.go
@@ -73,8 +73,18 @@ func parseInstanceFlag() {
 	}
 }
 
-// setLogLevel sets the log level from command flags or RDD_LOG_LEVEL.
-func setLogLevel(cmd *cobra.Command, _ []string) error {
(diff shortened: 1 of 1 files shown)


jandubois (commented):
The PR looks fine, but I would prefer `--log-format=json` to allow adding additional formatters later. It is also what is used by `kube-apiserver` / `kubelet` / `k3s` (`--logging-format=text|json`) and `containerd` (`--log-format`).

mook-as:
Switched to `--log-format=json` (or `=text`, or `--log-format=` empty string which defaults to text).

This can be tested via `rdd service status --log-format=json`, for example.

jandubois (approved):
Thanks, LGTM

Could be added to a BATS test, but also fine to merge as-is.

PR #355, opened 2026-05-14
K3s: Use a dynamic port

### LINKED ISSUE #327: Use a unique port for Kubernetes
Right now we are using `6443`, which does not allow side-by-side installation with Rancher Desktop 1.

Also does not allow to run multiple RDD instances at the same time, for BATS testing etc.

I think we should use something to that the default RDD_instance maps to `7443` (or some other `X443`), so something like `7441 + index`.

mook-as:
This configures K3s to use a port based on the instance index, but only whatever is free.

Fixes #327 (excluding the bits about traefik)

### DIFF
--- a/pkg/controllers/app/app/lima-template.yaml
+++ b/pkg/controllers/app/app/lima-template.yaml
@@ -28,22 +28,26 @@ portForwards:
   - 0
   hostSocket: "{{.Param.HOST_DOCKER_SOCKET}}"
 provision:
+# Write runtime parameters as a systemd EnvironmentFile. Using mode:data lets
+# Lima's Go template engine handle quoting via printf "%q", which produces
+# double-quoted strings compatible with systemd's EnvironmentFile parser.
+# Lima's boot script creates parent directories automatically.
+- mode: data
+  path: /etc/sysconfig/rancher-desktop
+  content: |
+    CONTAINER_ENGINE={{printf "%q" .Param.CONTAINER_ENGINE}}
+    HOST_DOCKER_SOCKET={{printf "%q" .Param.HOST_DOCKER_SOCKET}}
+    HOST_HOME_GUEST={{printf "%q" .Param.HOST_HOME_GUEST}}
+    KUBERNETES_ENABLED={{.Param.KUBERNETES_ENABLED}}
+    KUBERNETES_VERSION={{.Param.KUBERNETES_VERSION}}
+    KUBERNETES_PORT={{.Param.KUBERNETES_PORT}}
 - mode: system
   script: |
     #!/bin/bash
 
-    # All param need to be mentioned for Lima validation.
-    # $PARAM_KUBERNETES_PORT
-
     set -o errexit -o nounset -o pipefail -o xtrace
 
     systemctl stop rancher-desktop.target
-
-    # Write runtime parameters; service ExecCondition checks read these from this file.
-    mkdir -p /etc/sysconfig
-    for p in "${!PARAM_@}"; do
-      printf '%s=%q\n' "${p#PARAM_}" "${!p}"
-    done > /etc/sysconfig/rancher-desktop
 
     # Enable the requested container engine and disable the other.
     # containerd.service has a Conflicts=docker, so we must not start both.
--- a/pkg/controllers/app/app/lima-template.yaml
+++ b/pkg/controllers/app/app/lima-template.yaml
@@ -33,6 +33,7 @@ provision:
     #!/bin/bash
 
     # All param need to be mentioned for Lima validation.
+    # $PARAM_KUBERNETES_PORT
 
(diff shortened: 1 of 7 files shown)


pkg/controllers/app/app/lima-template.yaml
```diff
@@ -33,16 +33,17 @@ provision:
...
 
     systemctl stop rancher-desktop.target
 
     # Write runtime parameters; service ExecCondition checks read these from this file.
     mkdir -p /etc/sysconfig
-    printf 'CONTAINER_ENGINE=%s\nKUBERNETES_ENABLED=%s\nKUBERNETES_PORT=6443\n' \
-        "${PARAM_CONTAINER_ENGINE}" "${PARAM_KUBERNETES_ENABLED}" \
-        >/etc/sysconfig/rancher-desktop
+    for p in "${!PARAM_@}"; do
+      printf '%s="%s"\n' "${p#PARAM_}" "${!p}"
```
Nino-K: Is this going to be a concern for `HOST_HOME_GUEST` if it contains quotes, since it's the user's home path? If so, would it make sense to use `%q` instead of `%s`.

```
 printf '%s=%q\n' "${p#PARAM_}" "${!p}"
```

mook-as: Good idea, this is bash so that's available. Amended.

mook-as:
I don't understand why BATS is failing on Windows, that looks unrelated.  But it's passing fine on `main`, so it's _probably_ to do with the PR?

jandubois:
AI review: https://symmetrical-adventure-qjle1je.pages.github.io/20260519-103443-pr-355.html

Nino-K:
@jandubois can you review please? I was the original reviewer, but I don't want to self-approve. I've addressed the AI review comments.

pkg/controllers/app/app/lima-template.yaml
jandubois: This seems wrong, the file is read via `EnvironmentFile=/etc/sysconfig/rancher-desktop` and not sourced via bash, so bash syntax is not valid:

> The parser strips leading and trailing whitespace from the values of assignments, unless you use double quotes ("), in which case the whitespace inside the quotes is preserved. Inside the double quotes, \\", \\', \\s, \\t, \\n and \\\\ may be used for escaping the listed characters. Other backslash escape sequences are reserved for future expansion and currently produce an undefined value.

So a filename with a space would break:

```console
$ printf '%q\n' '/Users/jan with space/foo'
/Users/jan\ with\ space/foo
```

It needs to be `/Users/jan\swith\sspace/foo`. Maybe escaped spaces work (undefined value), but they are not in the spec, and I haven't tested it.

jandubois (approved):

pkg/controllers/app/app/lima-template.yaml
```diff
@@ -32,17 +32,26 @@ provision:
...
-        "${PARAM_CONTAINER_ENGINE}" "${PARAM_KUBERNETES_ENABLED}" \
-        >/etc/sysconfig/rancher-desktop
+    cat > /etc/sysconfig/rancher-desktop <<EOF
+CONTAINER_ENGINE=${PARAM_CONTAINER_ENGINE}
+HOST_DOCKER_SOCKET="${PARAM_HOST_DOCKER_SOCKET}"
+HOST_HOME_GUEST="${PARAM_HOST_HOME_GUEST}"
+KUBERNETES_ENABLED=${PARAM_KUBERNETES_ENABLED}
+KUBERNETES_VERSION=${PARAM_KUBERNETES_VERSION}
+KUBERNETES_PORT=${PARAM_KUBERNETES_PORT}
+EOF
```
jandubois: Invalid YAML format; all lines must be indented:

```suggestion
    cat > /etc/sysconfig/rancher-desktop <<EOF
    CONTAINER_ENGINE=${PARAM_CONTAINER_ENGINE}
    HOST_DOCKER_SOCKET="${PARAM_HOST_DOCKER_SOCKET}"
    HOST_HOME_GUEST="${PARAM_HOST_HOME_GUEST}"
    KUBERNETES_ENABLED=${PARAM_KUBERNETES_ENABLED}
    KUBERNETES_VERSION=${PARAM_KUBERNETES_VERSION}
    KUBERNETES_PORT=${PARAM_KUBERNETES_PORT}
    EOF
```

But it would probably be better to use a `data` provisioning "script". And using Go text template syntax allows us to get the right kind of quoting:

```yaml
provision:
- mode: data
  path: /etc/sysconfig/rancher-desktop
  content: |
    CONTAINER_ENGINE={{printf "%q" .Param.CONTAINER_ENGINE}}
    ...
```

jandubois (changes requested):

jandubois (approved):
Thanks, LGTM

Nino-K (approved):

PR #356, opened 2026-05-15
Add Kubernetes context controller

### LINKED ISSUE #325: Create Kubernetes context
Currently we have code in the provisioning script, that is copied from TwoCows and tries to do that:

- it uses `rancher-desktop` instead of `rancher-desktop-$RDD_INSTANCE` for the context name
- merging with an existing `rancher-desktop` fails

The whole logic should be done by the App controller, similar to how we manage docker contexts. We'll need to setup a watcher to be able to set the `KubernetesReady` condition, and then we can manage the kube context whenever that condition changes.

We already include the whole `kubectl` implementation in `rdd`, so we should only use upstream library functions to manipulate the `~/.kube/config` file.

Nino-K:
- Add KubernetesReconciler that probes k3s `/healthz` and manages the rancher-desktop instance context in `~/.kube/config`
- Add KubernetesReady condition and related constants to app_types.go
- Gate `computeSettledCondition` on KubernetesReady when k8s is enabled
- Pass `HOST_INSTANCE_CONFIG` param to Lima so the probe script writes the raw k3s kubeconfig to the host instance config path
- Replace old kubeconfig merge logic in lima-template.yaml with a simple copy, context merging is now handled by the controller

Fixes: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/325

### DIFF
--- a/bats/Makefile
+++ b/bats/Makefile
@@ -9,7 +9,7 @@ BATS_CORE := ./lib/bats-core/bin/bats
 # BATS_TIMEOUT=0 disables the wrapper entirely, which also disables
 # support-bundle capture; for long local runs, set a large value
 # (e.g. 86400) instead of 0.
-BATS_TIMEOUT ?= 1800
+BATS_TIMEOUT ?= 2700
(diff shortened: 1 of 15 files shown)


jandubois (approved):
Thanks, LGTM

Will file a follow-up PR with some stuff I noticed that doesn't strictly has to go into this PR.

PR #358, opened 2026-05-20
debug: surface probe failures and capture rdd goroutines on bats timeout

jandubois:
## Purpose

Investigation-only PR layered on top of #356 to diagnose two distinct failures in test 43 ("start VM with Kubernetes enabled"):

- **Ubuntu**: the host-side k3s probe (`kube_reconciler.go:88`) returns "not yet healthy" every ~5s for 26+ minutes. The original code collapsed connection errors and non-200 responses into a single `(false, nil)` return, so the underlying cause never reaches the log.
- **macOS Intel**: after a benign 409 conflict at 22:11:42, the `app` and `kubernetes-reconciler` workers fire zero reconciles for the remaining 28 minutes of the run. No goroutine dump exists in `rdd.stderr.log` because `matches_our_instance` does not match the rdd service (its cmdline carries no instance suffix), so the existing SIGQUIT-on-leak path skips it.

## Changes

- `probeK3sAPI` now returns a single error explaining each failure mode (kubeconfig unreadable, connection error with URL, non-200 with status code and body snippet).
- `bats-with-timeout.sh` reads the rdd service pidfile and SIGQUITs the service before SIGTERM-ing the bats target, so goroutine stacks land in `rdd.stderr.log`.

## Test plan

- [ ] Trigger a fresh BATS run and confirm the new probe error format on Ubuntu (`bats-app` job).
- [ ] Confirm a goroutine dump appears in `rdd-logs-macos-15-intel-bats-app/.../rdd.stderr.log` when the suite times out.

## Do not merge

Diagnostic patch for one CI run. After we have the new logs, the real fix goes to #356.

### DIFF
--- a/bats/tests/32-app-controllers/kube-context.bats
+++ b/bats/tests/32-app-controllers/kube-context.bats
@@ -34,22 +34,27 @@ kube_current_context() {
         return 0
     }
 
-    kubectl config current-context 2>/dev/null || echo ""
+    rdd kubectl config current-context 2>/dev/null || echo ""
 }
 
 kube_context_exists() { # <context-name>
     [[ -f "${KUBECONFIG}" ]] || return 1
-    kubectl config get-contexts -o name 2>/dev/null | grep -qx "$1"
+    rdd kubectl config get-contexts -o name 2>/dev/null | grep -qx "$1"
 }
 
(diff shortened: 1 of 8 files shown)


jandubois:
Superseded by #359 and #360 (and some followup once both are merged).

PR #359, opened 2026-05-21
fix(create-k8s-context): make k3s probe pass and unwedge kube-context.bats

jandubois:
## Summary

Four-part fix series sitting on top of #356. Each commit addresses a distinct bug the create-k8s-context tests hit after the earlier ones surface. Together they take Ubuntu and macOS `bats-app` from red to green.

## Commits

- **`fix: present client cert to k3s healthz probe`**
  `probeK3sAPI` and `probeCurrentKubeContext` set up TLS with the CA bundle but never loaded the client cert from `cfg.CertData`/`cfg.KeyData`. The k3s default kubeconfig uses client-cert auth (no bearer token), so `/healthz` rejected every probe with HTTP 401. `KubernetesReady` stayed at `Probing` forever and `rdd set running=true` timed out — the original failure on test 43.

- **`fix: split rdd apiserver kubeconfig from k3s mirror`**
  `instance.Config()` was overloaded: it both pointed at the rdd apiserver's kubeconfig (used by `rdd ctl` pass-throughs as `KUBECONFIG`) and was the destination the Lima probe copied the in-VM k3s kubeconfig to. The k3s overwrite clobbered the rdd apiserver kubeconfig, breaking `rdd ctl wait limavm/rd ...` in test 43's second command. Introduces `instance.K3sConfig()` returning `${Dir}/k3s.yaml` for the mirror, exposes it via `rdd svc paths k3s_config`, and documents `RDD_K3S_CONFIG`.

- **`fix: drop redundant KUBECONFIG-set guard in manageKubeContext`**
  `manageKubeContext` skipped `setCurrentKubeContext` whenever `$KUBECONFIG` was set, treating it as "user is curating manually." That conflicted with kube-context.bats setting `KUBECONFIG` for test isolation. `probeCurrentKubeContext` already checks whether the existing current-context is healthy and leaves it alone if so — the `KUBECONFIG`-existence check was redundant safety on top of that. Removing it restores both production safety and test isolation.

- **`fix(test): use rdd kubectl via helper functions in kube-context.bats`**
  kube-context.bats invoked bare `kubectl` for kubeconfig manipulation, which required system `kubectl` in PATH. Ubuntu CI has it; macOS doesn't, so tests 117 and 118 failed with "kubectl: command not found". Switching to bundled `rdd kubectl` (preserves caller's `KUBECONFIG`, unlike `rdd ctl`) removes the runner dependency. Also replaces `bash -c "kubectl ... | grep -qx"` in `try`-loops with a `kube_current_context_is` helper that uses `assert_output` per `bats-style.md`.

## Test plan

- [ ] CI passes on Ubuntu `bats-app` (probe path + kubeconfig handling)
- [ ] CI passes on macOS `bats-app` (test 43 onwards including kube-context tests)
- [ ] No regression in Windows `bats-app` (Kubernetes tests already skipped there)

## Notes

The series came out of a longer investigation tracked in #358. The diagnostic instrumentation that uncovered these bugs stays in that branch and will be extracted as a separate "permanent investigation infrastructure" PR.

## Do not merge separately

Layered on top of #356. Squash into #356 or merge in order after #356 lands.

### DIFF
--- a/bats/tests/32-app-controllers/kube-context.bats
+++ b/bats/tests/32-app-controllers/kube-context.bats
@@ -34,22 +34,27 @@ kube_current_context() {
         return 0
     }
 
-    kubectl config current-context 2>/dev/null || echo ""
+    rdd kubectl config current-context 2>/dev/null || echo ""
 }
 
 kube_context_exists() { # <context-name>
     [[ -f "${KUBECONFIG}" ]] || return 1
-    kubectl config get-contexts -o name 2>/dev/null | grep -qx "$1"
+    rdd kubectl config get-contexts -o name 2>/dev/null | grep -qx "$1"
 }
 
 kube_cluster_exists() { # <cluster-name>
     [[ -f "${KUBECONFIG}" ]] || return 1
-    kubectl config get-clusters 2>/dev/null | tail -n +2 | grep -qx "$1"
+    rdd kubectl config get-clusters 2>/dev/null | tail -n +2 | grep -qx "$1"
 }
 
 kube_user_exists() { # <user-name>
     [[ -f "${KUBECONFIG}" ]] || return 1
-    kubectl config get-users 2>/dev/null | tail -n +2 | grep -qx "$1"
+    rdd kubectl config get-users 2>/dev/null | tail -n +2 | grep -qx "$1"
+}
+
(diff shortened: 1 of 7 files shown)


jandubois:
Changes have been merged into #356

PR #361, opened 2026-05-21
kubernetes-reconciler: add V(1) tracing and merge-failure handling

jandubois:
## Summary

- Promote `watchEventLogger` to a shared `pkg/controllers/app/predicates` package so the App and kubernetes reconcilers share one implementation.
- Add a V(1) `reconcile entered` log and the watch-event predicate to the kubernetes reconciler, and log the six return-true paths in `probeCurrentKubeContext` (the function still returns true to preserve the user's current-context).
- Set `KubernetesReady=False` with reason `MergeFailed` when `manageKubeContext` cannot merge the instance kubeconfig. Previously the controller silently logged the error and set True, contradicting the condition's doc contract.

## Test plan

- [x] `go build ./...`, `go vet ./...`, unit tests under `pkg/controllers/app/...` and `pkg/apis/...` pass locally.
- [x] Smoke-tested via `rdd svc start --wait=false` then `rdd set running=false`; observed `FLAG: --v="4"` and 17 V(1) `reconcile entered` lines in stderr, plus watch events under `kubernetes-reconciler.watch`.
- [x] CI green.

Follow-up to #358.

### DIFF
--- a/cmd/rdd/service.go
+++ b/cmd/rdd/service.go
@@ -72,6 +72,9 @@ func newServiceConfigCommand() *cobra.Command {
 
 func ensureServiceRunning(ctx context.Context) error {
 	if !service.Exists() {
+		// Persist no extra args; the per-start klogArgs below override
+		// verbosity on every start, so baking the caller's logrus level
+		// into args.json would only mislead future inspectors of the file.
 		if err := service.Create(nil); err != nil {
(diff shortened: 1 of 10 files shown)


Nino-K (approved):

Nino-K:
LGTM

PR #370, opened 2026-05-27
Stop KubernetesReady from flapping on transient probe failures

jandubois:
## Summary

The kubernetes-reconciler treated every healthz probe failure as "k3s is not ready," so a single 3s `context deadline exceeded` flipped Ready→Probing even when the next probe would have succeeded. The `bats-app` `kube-context` tests 112 and 113 straddled that flip in run #26480815405 on `macos-15-intel`: 112 read `.status` (saw True), the
reconciler's own-write rebound ran another probe that timed out, and 113 read `.reason` (saw "Probing").

This PR classifies probe failures into three kinds and smooths only the genuinely ambiguous ones:

- **Unreachable** (ECONNREFUSED, EOF, ECONNRESET, unrecognized transport error) — k3s is gone. Flip immediately.

- **Unhealthy** (HTTP non-200) — k3s explicitly said "not ready." Flip immediately.

- **Ambiguous** (request did not complete within `kubeProbeTimeout`) — could be load, could be wedged, can't tell from one sample. Only this category is smoothed: while Ready, the reconciler tolerates up to `probeFailureThreshold=3` consecutive ambiguous failures (~15s total) before flipping to Probing. Any success or any decisive failure resets the counter. During startup the condition is not yet True, so the first Ambiguous failure flips immediately.

The shape mirrors the kubelet readiness-probe model (single success flips Ready, N failures flip NotReady) but adds error classification so deliberate k3s stops surface immediately rather than being smoothed for 15s. Rancher Desktop is interactive — a user who just stopped k3s shouldn't sit through a smoothing window when the controller can tell the difference from the failure mode itself.

Also adds an explicit `RequeueAfter=kubeHealthyRequeue=15s` on the healthy path. Without it, `setKubeCondition` is a no-op when Ready→Ready and nothing else would trigger a re-probe until the controller-runtime default resync (~10 min), so a later k3s death would go unnoticed.

### DIFF
--- a/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
+++ b/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
@@ -88,12 +88,12 @@ type KubernetesReconciler struct {
 	// contextProbeWg lets removeKubeContext wait for any probe to finish.
 	contextProbeWg sync.WaitGroup
 
-	// probeMu protects consecutiveProbeFailures.
-	probeMu sync.Mutex
 	// consecutiveProbeFailures counts ambiguous probe failures while the
-	// KubernetesReady condition was True. Resets on any success or any
-	// decisive failure (Unreachable, Unhealthy). When the count reaches
-	// probeFailureThreshold, the reconciler flips Ready to Probing.
+	// KubernetesReady condition was True. Resets on any success or
+	// decisive failure (Unreachable, Unhealthy). Crossing
+	// probeFailureThreshold flips Ready to Probing and resets the counter.
+	// Reconcile runs single-threaded (MaxConcurrentReconciles=1), so the
+	// counter needs no synchronization.
 	consecutiveProbeFailures int
 
 	// probeFn lets tests inject probe results without making real HTTP calls.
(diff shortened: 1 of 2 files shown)


pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -122,6 +179,35 @@ func (r *KubernetesReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (c
...
+		// condition is not yet Ready, so report Probing immediately.
+		failures := r.incrementProbeFailures()
+		currentlyReady := apimeta.IsStatusConditionTrue(
+			app.Status.Conditions, appv1alpha1.AppConditionKubernetesReady)
+		if currentlyReady && failures < probeFailureThreshold {
+			log.V(1).Info("ambiguous probe failure tolerated",
+				"failures", failures, "threshold", probeFailureThreshold)
+			return ctrl.Result{RequeueAfter: kubeProbeRequeue}, nil
+		}
+		log.V(1).Info("ambiguous probe failures crossed threshold",
```
Nino-K: This could be a little ambiguous since during the startup (before `KubernetesReady=true`), the first failure could land here with `failures=1` and `threshold=3` and the message would say "crossed threshold", but the real failed reason is "not currently ready".  It's not really major, but I found it a little confusing.

pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -58,6 +87,26 @@ type KubernetesReconciler struct {
 	contextProbeGen int
 	// contextProbeWg lets removeKubeContext wait for any probe to finish.
 	contextProbeWg sync.WaitGroup
+
+	// probeMu protects consecutiveProbeFailures.
+	probeMu sync.Mutex
```
Nino-K: Do we need the mutex? It looks like `Recncile` is never called concurrently.

pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -122,6 +179,35 @@ func (r *KubernetesReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (c
...
+		currentlyReady := apimeta.IsStatusConditionTrue(
+			app.Status.Conditions, appv1alpha1.AppConditionKubernetesReady)
+		if currentlyReady && failures < probeFailureThreshold {
+			log.V(1).Info("ambiguous probe failure tolerated",
+				"failures", failures, "threshold", probeFailureThreshold)
+			return ctrl.Result{RequeueAfter: kubeProbeRequeue}, nil
+		}
+		log.V(1).Info("ambiguous probe failures crossed threshold",
+			"failures", failures, "threshold", probeFailureThreshold)
+		if condErr := r.setKubeCondition(ctx, &app,
```
Nino-K, line 205: How are we resetting the `consecutiveProbeFailures` after the condition is set to `probing`.

Nino-K (approved):

PR #373, opened 2026-05-28
Add Stringer to `probeResult` to improve output

Nino-K:
Add `String()` method to `probeResult`, making the logs more readable by implementing the stringer interface.
Before: 
```
"result": 1
```    
After: 
```
"result": "Unreachable"
```


jandubois (approved):
Thanks, LGTM

PR #375, opened 2026-05-28
Add rdd start and rdd stop subcommands

jandubois:
- Add `rdd start` and `rdd stop` as thin wrappers around `setAction`. `rdd start` patches `spec.running=true` (auto-creating the App); `rdd stop` patches `spec.running=false` and short-circuits when the App is absent rather than creating it.

- Drop the empty `rdd create` stub from `docs/design/cmd_app.md`; `setAction` already creates the App on first call.

- Cover the unique branches in `bats/tests/32-app-controllers/start-stop.bats`.

### DIFF
--- a/cmd/rdd/app.go
+++ b/cmd/rdd/app.go
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/cli/help"
+)
+
+func newStartCommand() *cobra.Command {
+	var (
+		wait    bool
+		timeout time.Duration
+	)
+	command := &cobra.Command{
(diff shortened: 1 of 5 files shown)


docs/design/cmd_app.md
```diff
@@ -41,9 +41,6 @@ rdd set --wait=false running=true
 
 `2` is reserved for cobra usage errors. Other rdd commands will adopt the same scheme as they grow `--wait` semantics; the codes are defined in `pkg/cli/exit`.
 
-## `rdd create`
```
mook-as, line 44: Should also update https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/1f72c68510da1a48647a29ffd1d982d1aa80c01c/docs/design/api_app.md#L11

cmd/rdd/app.go
```diff
@@ -0,0 +1,94 @@
...
+			and, when Kubernetes is enabled, a Kubernetes cluster with
+			its context merged into ~/.kube/config.
+
+			By default, rdd start waits until Rancher Desktop is ready
+			to use. Pass --wait=false to return as soon as the request
+			is queued, or --timeout=0 to wait indefinitely.
+		`),
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return setAction(cmd.Context(), []string{"running=true"}, false, wait, timeout)
```
Nino-K, line 41: I noticed the `dryRun` was hardcoded to false, I guess we are not supporting dry run for these commands?

jandubois: No, just like we don't support setting any other properties with `rdd start`.

The `--dry-run` support is mostly useful to check if the new settings will validate without actually changing the object.

Think about the settings dialog that shows validation errors in the bottom pane. Yo can only "Apply" if `--dry-run` returns no errors. And if you "Cancel", then the object remains untouched. Of course the GUI will not run `rdd set --dry-run` but use the API directly. But this is the reason we support --dry-run to begin with.

Nino-K (approved):

PR #376, opened 2026-05-29
Show progress during control-plane startup and rdd set

jandubois:
## Summary

`rdd svc start` and `rdd set` waited silently through their slow paths and printed only a final line. Both now report progress as it happens.

```console
❯ rdd set running=true
INFO[0000] starting "rancher-desktop-2" control plane
INFO[0000] waiting for API server
INFO[0001] waiting for CRDs to be established
INFO[0008] App created
INFO[0008] Waiting for App to settle
INFO[0012] Created: Lima instance created successfully (Created)
INFO[0012] Settled: LimaVM has not yet reached Started (Reconciling)
INFO[0012] Running: Verifying instance state after controller restart (Reconciling)
INFO[0012] Settled: LimaVM has not yet reached Started (Starting)
INFO[0012] Running: Lima instance is starting (Starting)
INFO[0029] Settled: Container engine stopped (Stopped)
INFO[0029] Running: Lima instance is running (Started)
INFO[0029] ContainerEngineReady: Container engine synced (Connected)
INFO[0029] Settled: App has reached the desired state (Settled)

❯ rdd set kubernetes.enabled=true kubernetes.version=1.32.3
INFO[0000] App updated
INFO[0000] Waiting for App to settle
INFO[0000] ContainerEngineReady: Container engine synced (Connected)
INFO[0000] KubernetesReady: Waiting for k3s API server (Probing)
INFO[0000] Created: Lima instance created successfully (Created)
INFO[0000] Settled: Waiting for k3s API server (Probing)
INFO[0000] Running: Lima instance is running (Started)
INFO[0000] ContainerEngineReady: failed to ping Docker: failed to connect to the docker API at unix:///Users/jan/.rd2/docker.sock; check if the path is correct and if the daemon is running: dial unix /Users/jan/.rd2/docker.sock: connect: no such file or directory (ConnectFailed)
INFO[0000] Settled: failed to ping Docker: failed to connect to the docker API at unix:///Users/jan/.rd2/docker.sock; check if the path is correct and if the daemon is running: dial unix /Users/jan/.rd2/docker.sock: connect: no such file or directory (ConnectFailed)
INFO[0002] Settled: LimaVM has not yet reached Started (Stopped)
INFO[0002] Running: Stopped for restart (Stopped)
INFO[0002] KubernetesReady: VM is not running (NotRunning)
INFO[0002] Settled: LimaVM has not yet reached Started (Reconciling)
INFO[0002] Running: Verifying instance state after controller restart (Reconciling)
INFO[0002] Settled: LimaVM has not yet reached Started (Starting)
INFO[0002] Running: Lima instance is starting (Starting)
INFO[0002] ContainerEngineReady: Container engine stopped (Stopped)
INFO[0024] Settled: Container engine stopped (Stopped)
INFO[0024] Running: Lima instance is running (Started)
INFO[0024] KubernetesReady: Kubernetes API server is ready (Ready)
INFO[0024] ContainerEngineReady: Container engine synced (Connected)
INFO[0024] Settled: App has reached the desired state (Settled)
```

Log messages are at INFO level, so don't show unless developer mode is true, or the log level has explicitly been set to INFO or lower.

We can change the formatting later, and maybe filter some updates, but for now I like to see the individual transitions.

### DIFF
--- a/pkg/service/cmd/service.go
+++ b/pkg/service/cmd/service.go
@@ -289,15 +289,42 @@ func Start(ctx context.Context, args []string) error {
 	return cmd.Start()
 }
 
-// checkReadiness performs a single readiness check.
-func checkReadiness(ctx context.Context) error {
+// startupPhase identifies the control-plane startup stage the readiness
+// check is waiting on, so callers can report progress as the phase advances.
+type startupPhase int
+
+const (
+	phaseAPIServer startupPhase = iota
+	phaseControllers
+	phaseCRDs
+)
+
+// String returns the user-facing progress message for the phase.
+func (p startupPhase) String() string {
+	switch p {
+	case phaseAPIServer:
+		return "waiting for API server"
+	case phaseControllers:
+		return "waiting for controllers to register"
+	case phaseCRDs:
+		return "waiting for CRDs to be established"
+	default:
+		return ""
+	}
+}
+
+// checkReadiness performs a single readiness check, reporting via onPhase the
+// phase it ends up waiting on.
+func checkReadiness(ctx context.Context, onPhase func(startupPhase)) error {
 	config, err := GetKubeRestConfig()
 	if err != nil {
+		onPhase(phaseAPIServer)
 		klog.V(2).Infof("Waiting for kubeconfig to be available: %v", err)
 		return err
 	}
 
 	if err := checkSupportedVersion(config); err != nil {
+		onPhase(phaseAPIServer)
 		klog.V(2).Infof("%v", err)
 		return err
 	}
@@ -305,6 +332,7 @@ func checkReadiness(ctx context.Context) error {
 	// Wait for the controller manager to register with the actual running controllers
 	runtimeControllers, err := getRuntimeControllersFromCluster(ctx, config)
 	if err != nil {
+		onPhase(phaseControllers)
 		klog.V(2).Infof("getRuntimeControllersFromCluster: %v", err)
 		return err
 	}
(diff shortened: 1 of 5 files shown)


mook-as:
Should these also show up somewhere on the `app` object? This is the exact sort of thing that would be useful in the GUI to show, so if it's attached to the `app` we can just update the UI when that changes.  (Which also means this can be done by watching the app object, too…)

jandubois:
Everything after "Waiting for App to settle" actually comes from the `app.status.conditions`. It already watches the App object to determine when it is settled; now it also dumps out any condition changes to the log. So you can already do the same thing from the GUI.

And part of the reason for doing this was to see if this could be useful for the GUI status bar. Not sure yet about that.

But I noticed that changing k8s versions doesn't actually work properly, so this has already been useful.

pkg/service/cmd/service.go
```diff
@@ -289,33 +289,63 @@ func Start(ctx context.Context, args []string) error {
...
+type startupPhase int
+
+const (
+	phaseAPIServer startupPhase = iota
+	phaseControllers
+	phaseCRDs
+)
+
+// String returns the user-facing progress message for the phase.
+func (p startupPhase) String() string {
```
mook-as, line 303: It would be nice if these were available on the `app` instead of just here, so that the UI can pick these up instead of just using the status condition message directly.

mook-as (approved):
Looks fine, it's up to you if you want to make the suggested change.

pkg/service/cmd/service.go
jandubois: I don't see how this can work. These phases ("waiting for API server" etc) are before the app object exists, so can't be exposed through the app object. Once the app object is created, all further progress is tracked via the Condition changes.

PR #377, opened 2026-05-29
Verify process identity before trusting a recycled PID on Windows

jandubois:
On Windows, graceful hostagent shutdown signals the hostagent's PID with `GenerateConsoleCtrlEvent(CTRL_BREAK_EVENT)`. Windows recycles PIDs quickly, so a stale PID — especially one a later service reads from on-disk Lima state — can resolve to an unrelated process. When it does, the `CTRL_BREAK` escapes the intended process group and reaches every process sharing the console, killing the rdd service and the controlling terminal. In CI this aborted a bats run with `STATUS_CONTROL_C_EXIT` right after `killOrphanedHostagent` signalled a recycled PID.

The forced-stop path already sidesteps this by using `taskkill` instead of `CTRL_BREAK`; the graceful paths did not.

Add `process.IsOurProcess`, which confirms a PID still runs our own executable with the expected command line, and apply it where a PID comes from on-disk state with no open handle: `killOrphanedHostagent` and the force fallbacks. The in-memory paths that signal a live `*exec.Cmd` (`signalHostagent`, the `shutdownAllHostagents` loop) instead rely on the held process handle, which keeps Windows from recycling the PID while it is open. A failed check falls through to the reuse-safe forced stop, so an unverifiable PID is never signalled.

The service's own PID file carries the same on-disk hazard. `PID()` confirmed only that some process held the recorded PID, so after a crash left the file behind, a recycled PID made `Running()` report a dead control plane as running: `rdd svc start` waited on a control plane that never became ready and timed out, and `StopWithWait` could signal an unrelated process. Guard `PID()` with `IsOurProcess` too, so a recycled PID reads as not-running and the stale file is removed.

`IsOurProcess` compares the target's image path against `os.Executable()`. Because a user can launch rdd through an 8.3 short path (`C:\PROGRA~1\...`), it normalizes both paths to their long form before comparing, so a short-path launch never false-negatives a live process into an orphan.

On Unix the check is a no-op: `SIGINT` to a stale PID is harmless and PIDs are not recycled within a daemon's lifetime.

### DIFF
--- a/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
+++ b/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
@@ -200,13 +200,24 @@ func (r *LimaVMReconciler) waitForProcessExit(ctx context.Context, name string) 
 // signalHostagent sends a graceful shutdown signal to the hostagent process.
 // Uses process.Interrupt which sends SIGINT on Unix and CTRL_BREAK on Windows
 // (targeted at the hostagent's process group, not the parent).
-// Returns false if no watcher exists or the signal could not be delivered.
+// Returns false if no watcher exists, the process has already been reaped (its
+// PID may be reused on Windows), or the signal could not be delivered.
 func (r *LimaVMReconciler) signalHostagent(name string) bool {
 	r.instanceStatesMu.RLock()
 	state, ok := r.instanceStates[name]
 	r.instanceStatesMu.RUnlock()
 	if !ok || state.cmd == nil || state.cmd.Process == nil {
 		return false
+	}
+	// Once procExited is closed, cmd.Wait() has released the OS process handle
+	// and the PID may be reused; signalling it could deliver CTRL_BREAK to an
+	// unrelated console process on Windows. While it is open the handle is
+	// normally still held; cmd.Wait() releases it just before closing the
+	// channel, leaving a brief window.
+	select {
+	case <-state.procExited:
+		return false
+	default:
 	}
 	return process.Interrupt(state.cmd.Process.Pid) == nil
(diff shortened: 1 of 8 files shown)


mook-as:
I think I'm missing something: `state.cmd.Process` is a [`os.Process`](https://pkg.go.dev/os@go1.26.3#Process), which has an [open handle](https://github.com/golang/go/blob/2dc996f71b0ebafb77e64433e58333e049488a3c/src/os/exec.go#L55-L61) on the process; so, the PID can't be reused, because that process is still "alive" (not in the sense of running, but in the sense of has an active handle).  It's actually better to use the handle instead of the PID anyway…

This does not apply for `inst.HostAgentPID` because that's not an `os.Process`.

jandubois:
> I think I'm missing something:

Yes, when the pid files from the previous run have not been cleaned up properly, e.g. when the service crashed/terminated without stopping all instances. On a restart we need to decide if we can adopt the running instance, or need to restart them. But if the instance is gone and only the pid files survive, then the PIDs can already be reused.

The fact that PID files are left behind in BATS is yet another bug in rdd (graceful shutdown doesn't delete the files; only forceful shutdown does). I now have 2 separate stacks of Yaks, that refuse to collapse again...

jandubois:
> the PID can't be reused, because that process is still "alive" (not in the sense of running, but in the sense of has an active handle)

The `killOrphanedHostagent` case still needs the additional checks, but I've reverted the change on the cases where we still have an open handle. I think this should be good now.

jandubois:
Superseded by #383, which combines this with #379 — the two interact in shutdownAllHostagents.

PR #379, opened 2026-05-30
Always reclaim hostagent resources during graceful shutdown

jandubois:
`shutdownAllHostagents` removed the PID file, socket, and tmp files and terminated the WSL2 distro only when a hostagent failed to exit within `gracefulShutdownTimeout`. On Windows that cleanup was usually skipped: `exec.CommandContext` force-kills the hostagent (`TerminateProcess`) when the manager context is cancelled, so `procExited` closes promptly and the exit looks graceful — but the force-killed hostagent never ran its own teardown. It left its PID file and WSL2 distro behind, and the next `rdd svc start` mistook the stale PID file for an orphaned hostagent.

Run `stopInstanceForcibly` after waiting for the process to exit, regardless of how it exited. It is a no-op when the hostagent already cleaned up after itself — the PID file is gone, so `Inspect` reports no PID to kill and the distro is already stopped — and it reclaims the leftovers when the hostagent was force-killed.

The new bats assertion that the PID file is gone after `rdd svc stop` fails on Windows without this change.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -63,6 +63,15 @@ lima_instance_running() {
     assert_file_exists "${RDD_LIMA_HOME}/${name}/ha.pid"
 }
 
+# True while the instance's WSL2 distro is still running. wsl.exe prints UTF-16,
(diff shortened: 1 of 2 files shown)


jandubois:
Superseded by #383, which combines this with #377 — the two interact in shutdownAllHostagents.

PR #380, opened 2026-05-30
Hold Settled False until the LimaVM applies a template change

### LINKED ISSUE #378: `rdd set` returns `Settled` before a restart-requiring change is applied
## Summary

`rdd set kubernetes.version=X` (and any spec change that requires a VM restart) returns `Settled: App has reached the desired state` **before** the change is applied. The version does eventually take effect — the LimaVM restarts and the guest reinstalls k3s — but `rdd set --wait` has already returned, so the command reports success while the cluster still runs the old version.

## Reproduction

On a running instance with Kubernetes enabled:

```
rdd set kubernetes.version=<a different valid version>
```

`rdd set` prints `Settled: App has reached the desired state` and exits 0 within a second, while the VM has not yet restarted into the new k3s.

## Root cause

The change propagates correctly: the App controller rewrites the LimaVM template ConfigMap, the LimaVM controller sets `restartNeeded`, the guest reboots, and the provision script reinstalls k3s. That path works.

The bug is a race between when `Settled` is computed and when the restart starts:

1. The patch bumps `app.metadata.generation` to N.
2. The App reconciler re-stamps every feeding condition at generation N while the VM is still up and the old k3s is still reachable: `Running=Started`, `KubernetesReady=Ready`, `ContainerEngineReady=Connected`. `computeSettledCondition` (`pkg/controllers/app/app/controllers/app_controller.go`) therefore returns `Settled=True` with `observedGeneration=N`.
3. `waitForDesiredState` (`cmd/rdd/set.go`) returns as soon as it sees `Settled=True` at `observedGeneration >= N`.
4. Only afterward does the LimaVM controller set `restartNeeded`, stop the VM, and reboot into the new k3s.

So a multi-second window exists where `Settled` is True at the new generation but the change has not been applied, and `rdd set` returns inside that window.

## Why KubernetesReady does not catch it

`KubernetesReady` is version-blind. The kubernetes reconciler (`pkg/controllers/app/kubernetes/controllers/kube_reconciler.go`) reads `spec.kubernetes.enabled` but never `spec.kubernetes.version`; readiness is "VM running and `/healthz` returns 200." `/healthz` reveals nothing about the running version, so the old k3s answers 200 and the condition reports `Ready` at the new generation. Comparing the running k3s server version to `spec.kubernetes.version` would hold the condition not-ready until the reinstall finished.

## Evidence (instance "2", local repro)

Service log — the template update and restart fire, in order:

```
14:38:30.510 app_controller     Updated template ConfigMap   kubernetesVersion="1.32.1"
14:38:30.511 limavm_controller  Instance running with stale template, requesting restart
14:38:33     limavm_lifecycle   No watcher ... resetting Running condition to Unknown
14:38:33     limavm_lifecycle   Starting Lima instance
```

Guest serial log — the reinstall actually happens (on an earlier set, version 1.32.3):

```
21:30:24 boot.sh  Downloading binary .../v1.32.3+k3s1/k3s-arm64
21:30:26 boot.sh  Installing k3s to /usr/local/bin/k3s
21:30:28 k3s      Starting k3s v1.32.3+k3s1
```

The App reported `Settled` at the new generation before the `requesting restart` and reboot shown above.

## Impact

- `rdd set kubernetes.version=X` exits 0 "settled" while the cluster still runs the old version and is about to restart.
- A script that waits on `rdd set` proceeds too early.
- The same applies to other restart-requiring changes (for example `containerEngine.name`).

## Possible directions

1. Make `KubernetesReady` version-aware: compare the running k3s server version to `spec.kubernetes.version` and hold the condition not-ready until they match. Because `Settled` gates on `KubernetesReady`, this closes the window.
2. Gate `Settled` on the LimaVM having applied the template for generation N (the restart for that generation having completed), so a not-yet-restarted VM cannot satisfy the wait.

## Notes / out of scope

- The restart-and-reinstall path itself works; this is purely about premature readiness.
- Separately, rapid k3s **downgrades** appear to leave k3s unable to start (k3s does not support a datastore downgrade). That is a different issue.

jandubois:
Fixes #378.

## Problem

`rdd set kubernetes.version=X` (and any spec change that rewrites the LimaVM template, such as `containerEngine.name`) returns `Settled: App has reached the desired state` **before** the change is applied. For a few seconds after the patch, the App reconciler re-stamps every feeding condition at the new generation while the old VM is still up — `Running=Started`, `KubernetesReady=Ready`, `ContainerEngineReady=Connected` — so `Settled` briefly goes True at the new generation. `rdd set`'s wait returns inside that window, then the LimaVM restarts and reinstalls.

## Fix

Gate `computeSettledCondition` on whether the LimaVM is running the current template, comparing the template ConfigMap's `resourceVersion` against the LimaVM's `ObservedTemplateResourceVersion`. The LimaVM defers that field until the restart completes, so a match means the running instance reflects the current template. While they differ, `Settled` is False with reason `ApplyingTemplate`, regardless of the feeding conditions.

This is deliberately generic: it unsettles on any template write without inspecting what changed, and it makes a version-aware `KubernetesReady` probe unnecessary — once the restart has forced `Settled` False, `KubernetesReady` can only report `Ready` against the rebooted instance.

## Verification

**Unit test** — added a case proving the gate forces `Settled=False` even when every feeding condition is ready at the current generation. Confirmed it fails with the gate disabled and passes with it.

**End to end** (throwaway VM) — `rdd set containerEngine.name=containerd`, polling the `Settled`/`Running` conditions:

```
True/Settled/obsGen=1           | Running=True/Started   <- initial
False/ApplyingTemplate/obsGen=2 | Running=True/Started   <- gate holds it False on the not-yet-restarted VM
False/Starting/obsGen=2         | Running=False/Starting <- restart
True/Settled/obsGen=2           | Running=True/Started   <- settled after restart
```

The command blocked 18.65s (the real restart) instead of returning in ~1s. The `ApplyingTemplate` line is the exact window that previously reported `Settled=True` and let `rdd set` return early.


docs/design/api_app.md
```diff
@@ -125,11 +125,12 @@ status:
...
...conciler's read and write, the stamp reflects the write-time generation rather than the observed one. The App reconciler computes `Settled` from the mirrored `Running`, the engine-written `ContainerEngineReady`, and the Kubernetes-written `KubernetesReady` (the last only when `spec.kubernetes.enabled` is true), and stamps its own `observedGeneration` with the `metadata.generation` observed when it computed the condition, not the generation at write time. A spec change that rewrites the LimaVM template ConfigMap holds `Settled=False` (reason `ApplyingTemplate`) until the LimaVM restarts into the new template: LimaVM defers `status.observedTemplateResourceVersion` until the restart completes, and the App reconciler gates `Settled` on that field matching the template ConfigMap's `resourceVersion`. The retry-on-conflict loop re-reads on 409, so a successful write always carries the current generation. `rdd set` waits for `Settled.status=True` with `observedGeneration >= post-patch metadata.generation`: the generation filter rejects a stale snapshot from a prior reconcile, and the template gate rejects a fresh snapshot taken against a VM that has not yet restarted into the new template.
```
mook-as: This super-long line is becoming unreadable; please consider reorganizing this.  (While I prefer hard line wraps, that's not what I mean here.)

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -449,7 +456,11 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 				LastTransitionTime: cond.LastTransitionTime,
 			}) || changed
 		}
-		settled := computeSettledCondition(latest, engineEnabled, kubernetesEnabled)
+		// templateUpToDate, like engineEnabled and kubernetesEnabled, is computed
```
mook-as: The `templateUpToDate` comment here seems misplaced; if it's here, so should comments about `engineEnabled` and `kubernetesEnabled`.

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -494,6 +505,17 @@ func computeSettledCondition(app *v1alpha1.App, engineEnabled, kubernetesEnabled
 		ObservedGeneration: app.Generation,
 	}
 
+	// A template change triggers a VM restart; until the LimaVM is running the
+	// current template, the spec change has not taken effect. Hold Settled
+	// False even when the feeding conditions are momentarily all True at the
```
mook-as: It probably reads nicer as `… Hold Settled to False…`?  (Largely because that's three words in a row in title case.)

pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -483,7 +494,7 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 // kubernetesEnabled is false when no controller writes KubernetesReady;
 // it also gates on spec.kubernetes.enabled so a stopped cluster does not
 // hold Settled pending.
-func computeSettledCondition(app *v1alpha1.App, engineEnabled, kubernetesEnabled bool) metav1.Condition {
+func computeSettledCondition(app *v1alpha1.App, engineEnabled, kubernetesEnabled, templateUpToDate bool) metav1.Condition {
```
mook-as: Passing in three `bool` conditions in a row seems error-prone (it's hard to see if we're passing them in the correct order).  Fine for now, but something to consider in the future.

pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -325,13 +326,28 @@ func Test_computeSettledCondition(t *testing.T) {
...
+			wantMessage:       settledMessageApplyingTemplate,
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			got := computeSettledCondition(tt.app, tt.engineEnabled, tt.kubernetesEnabled)
+			got := computeSettledCondition(tt.app, tt.engineEnabled, tt.kubernetesEnabled, !tt.templateOutOfDate)
```
mook-as: Negating that third argument is hard to read (probably because it happend to line wrap right _after_ the `!` for me LOL).  But I guess you have it that way to avoid changing all the other test cases?

bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -772,6 +772,13 @@ docker_context_dir() {
     echo "${DOCKER_CONFIG}/contexts/meta/${hash}"
 }
 
+# assert_docker_context checks config.json's currentContext against the
+# expected value; suitable for polling with try.
+assert_docker_context() { # <expected-context>
+    run jq -r '.currentContext' "${DOCKER_CONFIG}/config.json"
```
mook-as: Given the other changes in this PR, `jq_raw`?

bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -42,17 +42,20 @@ kube_current_context() {
 
 kube_context_exists() { # <context-name>
     [[ -f "${KUBECONFIG}" ]] || return 1
-    rdd kubectl config get-contexts -o name 2>/dev/null | grep -qx "$1"
+    run -0 rdd kubectl config get-contexts -o name
+    grep --quiet --line-regexp "$1" <<<"${output}"
```
mook-as, line 46: Should that be using `assert_output --regexp`?  Same for the two functions below.

mook-as (changes requested):
I mostly just want the docs clarified; that's kind of hard to read right now.

bats/tests/32-app-controllers/engine-docker.bats
jandubois: I guess, for consistency. I didn't do it because here we read from a file and not a variable, but we can use `$(<config.json)`. Will make the change.

bats/tests/32-app-controllers/kube-context.bats
jandubois: I actually did this at first, but Claude talked me out of it because the functions are (also) used like this:

```bash
run -1 kube_context_exists "${context_name}"
```

For this to work properly we would need a version with `refute_output --regexp` as well. Which is why I reverted back to this. I'm going to leave this for another day.

docs/design/api_app.md
```diff
@@ -125,11 +125,18 @@ status:
...
+  How a consumer can trust that these conditions reflect the current spec:
+
+  - **Mirrored-condition timestamps** — `Created` and `Running` are mirrored from LimaVM, so their `lastTransitionTime` reflects the LimaVM transition, not the copy. That makes the timestamp meaningful for staleness checks.
```
mook-as: What copy is this talking about? The mirroring? I think I might rephrase this as:
> - **Mirrored-condition timestamps** — `Created` and `Running` are mirrored from LimaVM, including the `lastTransitionTime` timestamps.

(_That makes X meaningful for Y_ is awkward phrasing, please avoid.)

docs/design/api_app.md
```diff
@@ -125,11 +125,18 @@ status:
...
+  - **`ContainerEngineReady.observedGeneration`** — the engine reconciler stamps it with the App's `metadata.generation` read at write time. If the App's generation advances between the reconciler's read and write, the stamp carries the write-time generation, not the one it observed.
```
mook-as: FWIW, this still seems wrong for me: it's being stamped with the generation that _wasn't_ observed, contrary to the comment. But that's probably a different issue.

docs/design/api_app.md
```diff
@@ -125,11 +125,18 @@ status:
...
 
-  The `Created` and `Running` conditions are mirrored from LimaVM, so their `lastTransitionTime` reflects the LimaVM transition rather than the copy — the timestamp is meaningful for staleness checks. The engine reconciler stamps `ContainerEngineReady.observedGeneration` with the App's `metadata.generation` at the time it writes the condition; if the App's generation advances between the reconciler's read and write, the stamp reflects the write-time generation rather than the observed one. The App reconciler computes `Settled` from the mirrored `Running`, the engine-written `ContainerEngineReady`, and the Kubernetes-written `KubernetesReady` (the last only when `spec.kubernetes.enabled` is true), and stamps its own `observedGeneration` with the `metadata.generation` observed when it computed the condition, not the generation at write time. The retry-on-conflict loop re-reads on 409, so a successful write always carries the current generation. `rdd set` waits for `Settled.status=True` with `observedGeneration >= post-patch metadata.generation`, so stale snapshots cannot prematurely satisfy the wait.
+  How a consumer can trust that these conditions reflect the current spec:
```
mook-as: I think this might be better written as:

> In general things which care about waiting for the control plane to be ready should wait for `Settled=True`, because reaching that requires all of the following prerequisites:
> - The LimaVM instance must be up-to-date with its ConfigMap, otherwise `Settled` is `False/ApplyingTemplate`.
> - The LimaVM instance must be ready, based on the mirrored `Created` and `Running` status conditions, including the `lastTransitionTime` timestamps.
> - The `ContainerEngineReady` status condition is set by the engine reconciler; the `observedGeneration` is the App's `metadata.generation` when the status condition is set.  This ensures that the status is up to date with the App.
>
> The `Settled` status condition's `observedGeneration` is the App's generation when the conditions were observed.  This means, for example, `rdd set` can wait for `Settled=True` where `Settled.observedGeneration` is is at least the post-patch app's `metadata.generation`; this would ensure that it's waiting for the control plane to settle after it has received the changes.

The important things are:
- The status conditions are laid out in the order that they become true.
- The stuff _not_ about status conditions are not in the list, but appended as extra text.

Please double-check my text to ensure that it's accurate, and not missing important information. (Sorry, AI prompting leaked in here…)

mook-as (changes requested):
Looks good, just the markdown needs some work.

jandubois (commented):
@mook-as PTAL

I've tried to reword it, but not sure if this is fine now. It is a bit complicated because the way to Settled(Started) is not quite symmetric to Settled(Stopped).

mook-as (approved):
Good enough; we can refine in the future as it becomes necessary.

PR #382, opened 2026-05-30
Keep the container engine up when a k3s install fails

jandubois:
## Problem

A Kubernetes version change rewrites the LimaVM template and restarts the VM, and the provision script reinstalls k3s on reboot. But the script stops `rancher-desktop.target` at the top and only restarts it on the last line, all under `set -o errexit`. So when the k3s download fails — common right after a restart, when the interface has an address but the host-switch network is not yet routable — the script aborts before the restart, leaving the **container engine** (which has nothing to do with Kubernetes) stopped. `docker.sock` goes unreachable and the App never settles.

## Fix

- **Restart `rancher-desktop.target` from an EXIT trap**, so the engine comes back regardless of the k3s install outcome.
- **Retry the k3s download** in a bounded loop, so a transient early-boot failure resolves on its own instead of aborting the script.
- **`curl -fsSL`** — `-S` surfaces the real error on failure, while `-s` still keeps the progress meter out of the logs.

## Verification

**Deterministic mock** (engine survives a k3s-install failure): with `curl` forced to fail, the new script restarts the target via the trap; the old script exits 6 and never restarts it.

```
new template:  systemctl start --no-block rancher-desktop.target   -> engine recovers
old template:  (never reached)                                     -> engine stays down
```

**Live VM**: initial k8s 1.32.2 install settled in 34s; a 1.32.2 -> 1.32.3 change went `ConnectFailed -> Starting -> Probing -> Settled`, ending with docker active, k3s active, k3s v1.32.3, `ContainerEngineReady=Connected`.

`bash -n` clean; app-controller Go tests (incl. `Test_applySpecToTemplate`) pass.

## Notes

The provision script is intentionally left without a unit test. A mocked harness would stub out `systemctl` and `curl`, so it would only exercise `trap ... EXIT` (basic bash), not this change's behavior. Inducing a real k3s-download failure is impractical — the App webhook rejects unknown versions, and the trigger is a nondeterministic early-boot network race — so the behavior is verified manually as above.

### DIFF
--- a/pkg/controllers/app/app/lima-template.yaml
+++ b/pkg/controllers/app/app/lima-template.yaml
@@ -50,6 +50,12 @@ provision:
 
     set -o errexit -o nounset -o pipefail -o xtrace
 
+    # Restart rancher-desktop.target on exit, even if a step below fails. The
+    # container engine belongs to this target but does not depend on
+    # Kubernetes, so a failed k3s install must not leave it stopped. --no-block
+    # returns immediately; the kubeconfig probe below absorbs k3s startup time.
+    trap 'systemctl start --no-block rancher-desktop.target' EXIT
+
     systemctl stop rancher-desktop.target
 
     # Enable the requested container engine and disable the other.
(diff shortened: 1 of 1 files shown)


mook-as:
That sounds like k3s install should be a systemd unit that runs after `network-online.target` instead of being a provisioning script?

jandubois:
> That sounds like k3s install should be a systemd unit that runs after `network-online.target` instead of being a provisioning script?

Yes, I think so. The template/provisioning scripts should just do the configuration, but not any actual installation, or management of services.

For now I just wanted to make some minimal changes to make it more robust. (Also didn't know about `network-online.target` before).

Nino-K (approved):

PR #383, opened 2026-06-01
Harden Windows shutdown against recycled PIDs and leftover hostagent state

jandubois:
Windows recycles PIDs quickly, so a PID read from on-disk state — the service PID file, or the hostagent's Lima state — can resolve to an unrelated process long after the original exited. rdd then signals or force-kills that bystander.

This PR adds `process.IsOurProcess`, which confirms a stored PID still runs our own executable, and applies it to every path that acts on such a PID.

1. **Verify hostagent identity before sending CTRL_BREAK on Windows** — adds `process.IsOurProcess` and guards the signal and force-stop paths that trust an on-disk PID, so `GenerateConsoleCtrlEvent(CTRL_BREAK)` cannot escape to the console and kill the rdd service or controlling terminal.

2. **Reject a recycled service PID on Windows** — guards `service.PID()` so a recycled PID left in the PID file after a crash reads as not-running, instead of making `Running()` report a dead control plane as alive.

3. **Always reclaim hostagent resources during graceful shutdown** — `shutdownAllHostagents` force-stops after every wait, not only on timeout, and screens the on-disk PID with `IsOurProcess` before taskkill.

### DIFF
--- a/bats/tests/33-lima-controllers/limavm-running.bats
+++ b/bats/tests/33-lima-controllers/limavm-running.bats
@@ -58,9 +58,17 @@ assert_stdout_logs_contain() {
     assert_output --partial "${expected}"
 }
 
-lima_instance_running() {
+assert_instance_running() {
     local name=$1
     assert_file_exists "${RDD_LIMA_HOME}/${name}/ha.pid"
+}
+
+# Asserts the instance's WSL2 distro is still running. WSL_UTF8 makes
+# wsl.exe emit UTF-8 instead of its default UTF-16.
+assert_distro_running() {
+    local name=$1
+    run -0 env MSYS_NO_PATHCONV=1 WSL_UTF8=1 wsl.exe --list --running
+    assert_output --partial "lima-${name}"
 }
 
 get_restart_count() {
@@ -128,7 +136,7 @@ editor_cmd() {
 }
 
 @test "verify hostagent PID file exists" {
-    lima_instance_running "${VM_NAME}"
+    assert_instance_running "${VM_NAME}"
 }
 
 @test "logs shows hostagent stderr" {
@@ -187,7 +195,7 @@ EOF
 }
 
 @test "verify hostagent PID file is removed" {
-    try --max 30 --delay 1 --until-fail -- lima_instance_running "${VM_NAME}"
+    try --max 30 --delay 1 --until-fail -- assert_instance_running "${VM_NAME}"
 }
 
 # Crash recovery tests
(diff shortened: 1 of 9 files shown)


pkg/util/process/process_windows.go
```diff
@@ -33,6 +38,220 @@ func Interrupt(pid int) error {
...
+func samePath(image, self string) bool {
+	return strings.EqualFold(longPath(image), longPath(self))
+}
+
+// longPath returns p in its long, non-8.3 form. If the long form cannot be
+// resolved (for example the file no longer exists), it falls back to
+// filepath.Clean(p) so a transient resolution failure degrades to a lexical
+// comparison rather than a guaranteed mismatch — for the service PID a false
+// "not ours" would orphan a live control plane.
+func longPath(p string) string {
```
Nino-K, line 151: I think this could treat an unknown process as ours, kind of a false positive. If `GetLongPathName` fails for the process that is checking, let's say the file was moved or deleted after the process started, the fallback compares raw cleaned paths. If both paths fail to resolve, they'll likely compare equal and return `true`. Could that be a concern?

bats/tests/33-lima-controllers/limavm-running.bats
```diff
@@ -332,6 +340,21 @@ EOF
...
+    # Windows, where exec.CommandContext terminates it when the manager context
+    # is cancelled). A leftover PID file makes the next "rdd svc start" mistake
+    # it for an orphan. This assertion is cross-platform: it fails on Windows
+    # before the shutdownAllHostagents cleanup fix.
+    try --max 30 --delay 1 --until-fail -- assert_instance_running "${VM_NAME}"
+
+    # On Windows the hostagent runs inside a WSL2 distro that a force-kill leaves
+    # running — the keepAlive process (nohup sleep) outlives the hostagent — so
+    # graceful shutdown must terminate the distro explicitly.
+    if is_msys; then
```
Nino-K: Do we need to check for `WSL2` too?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -638,6 +617,63 @@ func waitForInstanceStopped(ctx context.Context, name string) bool {
...
+	return filepath.Join(inst.Dir, filenames.HostAgentPID)
+}
+
+// clearRecycledHostAgentPID zeroes inst.HostAgentPID unless it still names our
+// live hostagent, so a force-stop's taskkill cannot reach an unrelated process
+// that recycled the PID. The stored PID comes from on-disk state a previous
+// service wrote, which Windows may have reassigned after the hostagent exited;
+// on other platforms IsOurProcess is a no-op, so the PID is kept. DriverPID is
+// not screened here — IsOurProcess matches the rdd image, not the qemu/wsl
+// driver — so a recycled DriverPID is still taskkilled by the caller.
```
Nino-K, line 635: Should we have a `TODO` referencing Job Objects as the proper fix for the DriverPID case?

bats/tests/33-lima-controllers/limavm-running.bats
jandubois: Yes, this should be `is_windows`.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: This is already mentioned earlier in the file, but not with an explicit TODO marker. Maybe I'll add the marker...

pkg/util/process/process_windows.go
jandubois: `filepath.Clean()` for 2 different files will never return strings that can compare as equal.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
Nino-K: Oops, I didn't see that.

Nino-K (approved):
LGTM

PR #389, opened 2026-06-04
Add support for `stable` and `latest` Kubernetes versions

### LINKED ISSUE #387: Add support for `stable` and `latest` Kubernetes versions
This has to be implemented in a mutating admission webhook. We do have the `stable` and `latest` mappings in `k3s-version.json`, so we can replace these strings with the actual versions at creation time.

The validating admission webhook runs afterward and will only accept real versions that it knows about.

jandubois:
A mutating admission webhook resolves channel aliases in `spec.kubernetes.version` to concrete versions before the validating webhook runs. The aliases come from the bundled `k3s-versions.json`, which maps `stable`, `latest`, and each minor version (`v1.32`, `v1.33`, ...) to a concrete release.

When Kubernetes is enabled without a version, the webhook fills in the `stable` channel. It strips a leading "v" so both "v1.32" and "1.32" resolve, and leaves an unknown alias unchanged for the validating webhook to reject as unsupported.

Enabling Kubernetes with an empty version now succeeds, so the dry-run test that asserted rejection asserts acceptance.

Closes #387

### DIFF
--- a/bats/tests/32-app-controllers/app.bats
+++ b/bats/tests/32-app-controllers/app.bats
@@ -334,6 +334,26 @@ EOF
     assert_output --partial "not found"
 }
 
+@test "mutating webhook resolves Kubernetes version channels and defaults to stable" {
+    delete_app
+    # --wait=false: the webhook resolves the version at admission, so the test reads it
+    # without waiting for the App to settle. The LimaVM is still created in the background.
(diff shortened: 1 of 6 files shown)


mook-as (approved):
… no comments. Huh.

PR #390, opened 2026-06-04
Add the `rdd run` command

### LINKED ISSUE #386: Create `rdd run` command
This command will add `~/.rd2/bin` to the front of the `PATH` and select `rancher-desktop-2` as the docker and kube context before executing the command, e.g.

```shell
rdd run docker run --rm hello-world
```

This will be implemented in Go as part of `rdd`, but work similar to:

```shell
rdd set running=true
export PATH=~/.rd2/bin:$PATH
unset DOCKER_HOST
export DOCKER_CONTEXT=rancher-desktop-2
# setup kube context via a temporary KUBECONFIG file with a copy of rancher-desktop-2 set as current
# similar to `kubectx`
eval "$@"
```

jandubois:
`rdd run COMMAND [ARGS...]` runs a command wired to this instance's Docker and Kubernetes contexts, so users need not edit their shell profile or switch their global contexts.

rdd run first ensures Rancher Desktop is running: it starts the App, and when the App is new and the command is kubectl or helm it also enables Kubernetes so the command has a cluster to talk to. It then wires the child process's environment — prepending `~/.rd<instance>/bin` to `PATH`, setting `DOCKER_CONTEXT` to `rancher-desktop-<instance>` while clearing `DOCKER_HOST`, and pointing `KUBECONFIG` at `~/.rd<instance>/kube.config`. The user's selected Docker context and the current context in `~/.kube/config` stay unchanged; starting the App still merges the `rancher-desktop-<instance>` entry into `~/.kube/config` as usual.

To back that `KUBECONFIG`, the Kubernetes controller publishes a standalone kubeconfig at `~/.rd<instance>/kube.config` holding only the `rancher-desktop-<instance>` context. It is written alongside the `~/.kube/config` merge once k3s is healthy and removed on teardown, so `rdd run` points at this file instead of copying credentials into a per-invocation temp file.

`rdd` runs the command as a child process rather than replacing itself, so it propagates the child's exit status uniformly across platforms. A help flag prints usage before the App starts, and a non-zero child exit sets `rdd`'s status without emitting a spurious error line.

Unit tests cover the kube-command classification, the standalone kubeconfig writer and remover, the environment wiring (`PATH`, Docker context, `DOCKER_HOST`, `KUBECONFIG`), and the help short-circuit. Integration tests in `engine-docker.bats` and `kube-context.bats` exercise the Docker context, Kubernetes context, `PATH` wiring, and exit-status propagation against a running instance.

Closes #386

### DIFF
--- a/bats/tests/32-app-controllers/kube-context.bats
+++ b/bats/tests/32-app-controllers/kube-context.bats
@@ -12,6 +12,7 @@ load '../../helpers/load'
 APP_NAME="app"
 VM_NAME="rd"
 K3S_VERSION="1.32.0"
+CONTEXT_NAME="rancher-desktop-${RDD_INSTANCE}"
 
 local_setup_file() {
     # Isolate ~/.kube/config so tests never touch the developer's real kubeconfig.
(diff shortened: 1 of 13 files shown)


pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -511,6 +522,9 @@ func (r *KubernetesReconciler) removeKubeContext(ctx context.Context) {
 	if err := deleteKubeContext(contextName); err != nil {
 		log.Error(err, "Failed to delete Kubernetes context", "context", contextName)
 	}
+	if err := removeInstanceKubeConfig(r.InstanceKubeConfigPath); err != nil {
```
Nino-K, line 525: It looks like `removeKubeContext` deletes `~/.rd2/kube.config` synchronously when k8s is disabled or the VM stops. Let's say a `rdd run kubeclt ...` process is in flight during that time and it has already read `KUBECONFIG=~/.rd2/kube.config`. But kubectl re-reads the file for each operation, and it would fail mid-operation. The previous temp file approach avoided this because each `rdd run` call got its own copy that lived for the duration of the child. I'm not saying this approach is going to be problematic, just thinking out loud here.

jandubois: I think this is fine. If you stop Kubernetes while you have a `kubectl` in flight, you get to keep all the pieces that fall down. Even even we keep the file, it will still not work because the apiserver is gone.

bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -23,9 +24,14 @@ local_setup_file() {
     # from the drive root (C:\tmp\...) rather than MSYS2's /tmp which maps to a
     # different location. cygpath -m produces a mixed-format path (C:/msys64/...)
     # that both native Windows processes and MSYS2 agree on.
-    if is_windows; then
+    if is_msys; then
```
Nino-K, line 27: I guess we should always make sure the `is_msys` is the first condition to avoid sending `MSYS2` into the wslpath branch since it satisfies both conditions.

Nino-K (approved):

bats/tests/32-app-controllers/kube-context.bats
mook-as: Isn't the plan to drop support for running tests under WSL anyway? So we should just delete the WSL path where that causes issues (and at some point just delete all of it).

jandubois: Yes, but can we leave this for another day?

It might still be useful to run BATS under WSL using Linux binaries (for testing `rdd` instead of `rdd.exe`). It obviously doesn't work yet, and I don't want to think about this right now.

mook-as: Ah, yeah, good point; either way, not something to actively worry about right now.

PR #392, opened 2026-06-04
Enable Kubernetes context tests on Windows

Nino-K:
This PR enables the previously skipped kube-context.bats tests on Windows and fixes two root causes that prevented them from passing.

- Enable kube-context tests on Windows. Removed the `skip_on_windows` guard from kube-context.bats.
- Bump opensuse distro to v0.2.4.
- Fix stale Lima instance recovery and `KUBECONFIG` path

Two bugs prevented the tests from passing on Windows:

1. Stale Lima instance on restart when RDD was force-killed, Lima's WSL2 hostagent could survive as an orphan holding `ext4.vhdx` open. On the next start, `limainstance.Create` would fail with "already exists". The fix changes this in the LimaVM reconciler and calls
  `wsl.exe --unregister` (not just --terminate) to fully remove the WSL2 registration and VHD before retrying. Using `--terminate` alone was insufficient: it left the WSL2 registration intact, causing Lima's `Prepare()` to skip `wsl.exe --import`, which meant no new VHD was created and the hostagent would crash-loop.
2. Wrong KUBECONFIG path on Windows — `BATS_FILE_TMPDIR` is a POSIX path (`/tmp/...`). The rdd.exe binary interprets this as `C:\tmp\... `(drive root) rather than MSYS2's `C:\msys64\tmp\...`, so the kubernetes controller was writing the kubeconfig to the wrong location. Fixed by using `cygpath -m` to convert the path to a mixed format that both MSYS2 and native Windows processes agree on, similar to what was already done in engine-docker.bats.

Related: https://github.com/rancher-sandbox/rancher-desktop-daemon/issues/391

### DIFF
--- a/bats/tests/32-app-controllers/kube-context.bats
+++ b/bats/tests/32-app-controllers/kube-context.bats
@@ -18,7 +18,18 @@ local_setup_file() {
     # Set KUBECONFIG before starting the service so the controller process inherits
     # it; service.Start uses exec.Command without an explicit Env, so it inherits
     # the caller's environment.
-    export KUBECONFIG="${BATS_FILE_TMPDIR}/kube/config"
+    #
+    # On Windows, rdd.exe is a native binary: it interprets /tmp/... as a path
+    # from the drive root (C:\tmp\...) rather than MSYS2's /tmp which maps to a
+    # different location. cygpath -m produces a mixed-format path (C:/msys64/...)
+    # that both native Windows processes and MSYS2 agree on.
+    if is_windows; then
+        run -0 cygpath -m "${BATS_FILE_TMPDIR}/kube/config"
+        KUBECONFIG="${output}"
+    else
+        KUBECONFIG="${BATS_FILE_TMPDIR}/kube/config"
+    fi
+    export KUBECONFIG
     mkdir -p "$(dirname "${KUBECONFIG}")"
 
     rdd svc delete
--- a/bats/tests/32-app-controllers/kube-context.bats
+++ b/bats/tests/32-app-controllers/kube-context.bats
@@ -14,8 +14,6 @@ VM_NAME="rd"
 K3S_VERSION="1.32.0"
 
 local_setup_file() {
-    skip_on_windows "Kubernetes context tests require Lima"
-
     # Isolate ~/.kube/config so tests never touch the developer's real kubeconfig.
     # Set KUBECONFIG before starting the service so the controller process inherits
     # it; service.Start uses exec.Command without an explicit Env, so it inherits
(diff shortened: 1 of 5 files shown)


bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -14,13 +14,22 @@ VM_NAME="rd"
...
     # it; service.Start uses exec.Command without an explicit Env, so it inherits
     # the caller's environment.
-    export KUBECONFIG="${BATS_FILE_TMPDIR}/kube/config"
+    #
+    # On Windows, rdd.exe is a native binary: it interprets /tmp/... as a path
+    # from the drive root (C:\tmp\...) rather than MSYS2's /tmp which maps to a
+    # different location. cygpath -m produces a mixed-format path (C:/msys64/...)
+    # that both native Windows processes and MSYS2 agree on.
+    if is_windows; then
+        run -0 cygpath -m "${BATS_FILE_TMPDIR}/kube/config"
```
mook-as, line 27: I'm surprised this is the only place we need this. Yay?

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -718,6 +718,20 @@ func terminateWSL2Distro(ctx context.Context, logger logr.Logger, instName strin
 	}
 }
 
+// unregisterWSL2Distro sends `wsl.exe --unregister` for the Lima distro with
+// the given instance name. This removes the WSL2 registration and deletes the
+// distro's ext4.vhdx, allowing Lima to import a fresh distro on the next
+// Prepare. Best-effort with a 30-second timeout (unregister is slower than
+// terminate and can hang if WSL2 is degraded).
+func unregisterWSL2Distro(ctx context.Context, logger logr.Logger, instName string) {
```
mook-as, line 729: Shouldn't this be skipped on non-Windows? On Linux, for example, this would always report an error because `wsl.exe` doesn't exist. (I was expecting a `runtime.GOOS` check somewhere, either in this function or in its caller.)

Would it make sense to move the `os.RemoveAll` logic in here too, as a general "remove existing dangling Lima directory" function?

mook-as (changes requested):
Just the changes around `unregisterWSL2Distro`.

bats/tests/32-app-controllers/kube-context.bats
Nino-K: I copied this over from Docker context bats.

pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
Nino-K: I'll change that for both `terminateWSL2Distro` and `unregisterWSL2Distro`.

mook-as (approved):
Now we get the OS check in the caller _and_ the callee, super sure we're doing the right thing! (I expect the compiler to optimize it all out because `runtime.GOOS` is a constant.)

jandubois (approved):
Thanks, LGTM

I think #383 has a more comprehensive solution but let's get this one merged to unblock Kubernetes on Windows.

PR #395, opened 2026-06-06
Test reaching a container from the host

### LINKED ISSUE #393: Create BATS tests to run a container and connect to it from the host
End-to-end test that you can actually run and use a containerized application.

Should be 2 tests, one for `docker` and one for `kubectl`.

jandubois:
Two end-to-end bats tests for #393: run a containerized nginx and reach it from the host, one per container path:

- `engine-docker.bats` publishes a Docker port (`-p 18080:80`) and curls `localhost:18080`.
- `kube-context.bats` exposes a Kubernetes NodePort and curls `localhost:<nodePort>`.

Both ride the VM their suite already boots, so they add no VM startup.

Closes #393

### DIFF
--- a/rdd/bats/tests/32-app-controllers/engine-docker.bats
+++ b/rdd/bats/tests/32-app-controllers/engine-docker.bats
@@ -245,6 +245,32 @@ assert_single_mirror() { # <image-id> <expected-tag>
     docker rm --force test-exposed test-published
 }
 
+# --- Host connectivity ---
+
+@test "published container port is reachable from the host" {
+    # Run nginx with a published port and reach it from the host. The
+    # container listens on 0.0.0.0 inside the VM; the Lima template forwards
+    # each guest 0.0.0.0 port to the host, so localhost:<port> reaches the
+    # containerized application.
+    #
+    # Pull nginx up front (quietly), mirroring the Kubernetes NodePort test;
+    # docker run would otherwise pull it inline.
+    docker pull --quiet nginx
+    host_port=18080
+    run_e -0 docker run --detach --name test-connect --publish "${host_port}:80" nginx
+    cid=${output}
+
+    rdd ctl wait --for=jsonpath='{.status.status}'=running \
+        --namespace="${RDD_NAMESPACE}" container/"${cid}" --timeout=30s
+
+    # The port forward is established asynchronously once the container
+    # starts listening, so poll until nginx answers.
+    try --max 30 --delay 2 -- \
+        assert_http_body "http://localhost:${host_port}" --partial "Welcome to nginx"
+
+    docker rm --force test-connect
+}
+
 # --- Container logs ---
 
 @test "docker logs for container with tty" {
--- a/rdd/bats/tests/32-app-controllers/kube-context.bats
+++ b/rdd/bats/tests/32-app-controllers/kube-context.bats
@@ -10,7 +10,6 @@ load '../../helpers/load'
 # KubernetesReady condition on the App resource.
 
 APP_NAME="app"
-VM_NAME="rd"
 K3S_VERSION="1.32.0"
 CONTEXT_NAME="rancher-desktop-${RDD_INSTANCE}"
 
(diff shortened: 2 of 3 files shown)


rdd/bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -155,6 +155,39 @@ kube_current_context_is() { # <expected-context>
...
+    # opens the NodePort on the node's 0.0.0.0; the Lima template forwards
+    # that port to the host, so localhost:<nodePort> reaches the application.
+
+    # Warm the image cache before deploying so the cold nginx pull and its layer
+    # extraction stay off the rollout's clock; on a slow CI runner the freshly
+    # booted node is also importing the k3s system images, and that extraction
+    # alone can outlast the rollout timeout. With the default moby backend, k3s
+    # runs through cri-dockerd, so the kubelet pulls from dockerd; warm it with a
+    # docker pull in the guest. --quiet keeps the log readable if a later step fails.
+    rdd limavm shell "${VM_NAME}" sudo docker pull --quiet nginx
```
mook-as: Wouldn't just `run -0 docker pull …` be good enough? Do we need a `limavm shell` here?

Also, for the comment, `… warm it with a docker pull in the guest` is confusing, in that _guest_ is unexpected here (though I assume it's from the Lima terminology); saying `… in the VM` would be clearer.

mook-as (changes requested):

rdd/bats/tests/32-app-controllers/kube-context.bats
jandubois: We would need to set the docker context first, but now we have `rdd run`, so we can use that.

rdd/bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -245,6 +245,32 @@ assert_single_mirror() { # <image-id> <expected-tag>
...
+
+@test "published container port is reachable from the host" {
+    # Run nginx with a published port and reach it from the host. The
+    # container listens on 0.0.0.0 inside the VM; the Lima template forwards
+    # each guest 0.0.0.0 port to the host, so localhost:<port> reaches the
+    # containerized application.
+    #
+    # Pull nginx up front (quietly), mirroring the Kubernetes NodePort test;
+    # docker run would otherwise pull it inline.
+    docker pull --quiet nginx
```
mook-as, line 258: Oh, we don't use the GHCR mirrors here. Not in scope for this PR, though.

rdd/bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -245,6 +245,32 @@ assert_single_mirror() { # <image-id> <expected-tag>
...
+
+    rdd ctl wait --for=jsonpath='{.status.status}'=running \
+        --namespace="${RDD_NAMESPACE}" container/"${cid}" --timeout=30s
+
+    # The port forward is established asynchronously once the container
+    # starts listening, so poll until nginx answers.
+    try --max 30 --delay 2 -- \
+        assert_http_body "http://localhost:${host_port}" --partial "Welcome to nginx"
+
+    docker rm --force test-connect
```
mook-as, line 271: This would never run if the test fails (e.g. if `rdd ctl wait` returns an error)?

Any reason we're not using `cid`  here? We probably don't even need to name the container.

rdd/bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -245,6 +245,32 @@ assert_single_mirror() { # <image-id> <expected-tag>
...
+    # Run nginx with a published port and reach it from the host. The
+    # container listens on 0.0.0.0 inside the VM; the Lima template forwards
+    # each guest 0.0.0.0 port to the host, so localhost:<port> reaches the
+    # containerized application.
+    #
+    # Pull nginx up front (quietly), mirroring the Kubernetes NodePort test;
+    # docker run would otherwise pull it inline.
+    docker pull --quiet nginx
+    host_port=18080
+    run_e -0 docker run --detach --name test-connect --publish "${host_port}:80" nginx
```
mook-as, line 260: We may need to force delete any containers named `test-connect` first.

rdd/bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -155,6 +154,37 @@ kube_current_context_is() { # <expected-context>
...
 
+# --- Host connectivity ---
+
+@test "NodePort service is reachable from the host" {
+    # Deploy nginx behind a NodePort service and reach it from the host. k3s
+    # opens the NodePort on the node's 0.0.0.0; the Lima template forwards
+    # that port to the host, so localhost:<nodePort> reaches the application.
+
+    # Warm the image cache so the cold nginx pull stays off the rollout's clock
+    # on a slow CI runner. With the default moby backend k3s pulls through
```
mook-as: I think I'd phrase the first sentence as:
> Manually pull the image ahead of time so the time spent pulling the image is not part of the timeout rolling out the deployment; this is especially helpful on slow CI runners.

rdd/bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -155,6 +154,37 @@ kube_current_context_is() { # <expected-context>
...
+    # cri-dockerd from dockerd, so warm it with rdd run, which points docker at
+    # the VM's engine.
```
mook-as: Suggestion:
> With the default moby backend k3s pulls through cri-dockerd from dockerd; use `rdd run docker` to ensure we are using the correct docker context.

rdd/bats/tests/32-app-controllers/kube-context.bats
```diff
@@ -155,6 +154,37 @@ kube_current_context_is() { # <expected-context>
...
+    rdd run docker pull --quiet nginx
+
+    rdd kubectl --context "${CONTEXT_NAME}" create deployment test-connect --image=nginx
```
mook-as, line 185: Hmm, `rdd kubectl` doesn't set up the correct context automatically? Is there a `rdd run kubectl` that would?

Either way, that's out of scope for this PR, probably, unless the correct command already exists and you feel like using it.

mook-as (changes requested):
Some stuff that would be out of scope for this PR, plus some phrasing.

rdd/bats/tests/32-app-controllers/engine-docker.bats
jandubois: Yeah, we need to add this from RD1. Later.

rdd/bats/tests/32-app-controllers/engine-docker.bats
jandubois: We always start the test file with a fresh VM (because we run `rdd svc delete`).

That has always been our strategy with BATS: we make sure we start from a fresh slate. We don't clean up at the end of the test. This help to manually debug failing tests because the broken instance is still there.

rdd/bats/tests/32-app-controllers/engine-docker.bats
jandubois: See above, we should not clean up here. `rdd svc delete` will clean up everything when we rerun the test.

rdd/bats/tests/32-app-controllers/kube-context.bats
jandubois: This is a good point. `rdd run` sets `KUBECONFIG` to `~/.rd2/kube.config`, which only contains the `rancher-desktop-2` context (and is the current context).

I added a test for that as well, but kept the one with the explicit `--context` using `~/.kube/context` to confirm that both locations are setup correctly.

rdd/bats/tests/32-app-controllers/kube-context.bats
jandubois: I don't see how this is clearer; it just uses more words. 😄

mook-as (approved):

PR #396, opened 2026-06-06
Add docker engine command BATS suite

jandubois:
Adds `tests/40-docker`, a BATS suite that exercises the docker command surface against the rdd moby engine: container lifecycle, images, volumes, networks, build, compose, and system. Run via `make bats-docker`.

`engine-docker.bats` covers the controller that mirrors docker objects into `Container`/`Image`/`Volume` resources, not the docker command surface itself. This suite exercises the commands users run.

### DIFF
--- a/rdd/bats/tests/40-docker/compose.bats
+++ b/rdd/bats/tests/40-docker/compose.bats
@@ -1,0 +1,50 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+load '../../helpers/load'
+
+# Compose commands against the rdd Docker engine: up, ps, logs, down. Compose is
+# a docker CLI plugin; tests skip when it is absent.
+
+local_setup_file() {
+    start_docker_engine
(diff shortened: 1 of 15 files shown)


rdd/bats/tests/40-docker/compose.bats
```diff
@@ -0,0 +1,50 @@
...
+@test "compose logs returns service output" {
+    cat >"${BATS_TEST_TMPDIR}/compose.yaml" <<EOF
+services:
+  app:
+    image: ${IMAGE_BUSYBOX}
+    command: echo composed
+EOF
+    run -0 host_path "${BATS_TEST_TMPDIR}/compose.yaml"
+    compose_file=${output}
+    docker compose --file "${compose_file}" --project-name rdd-compose-logs up --detach
```
Nino-K: This might end up giving us some headache in CI (just thinking out loud), because the container above runs `echo composed` and exists immediately. `up --detached` returns as soon as the service container is created, not after it exists. On a loaded CI runner, this container might not have flushed all its logs yet when `compose logs` is called.

It looks like `container.bats "logs"` test avoids this by running in the foreground (no `--detach`) so the container exits before `docker logs` is called.

Nino-K (approved):

PR #397, opened 2026-06-06
Recover and surface a failed host-switch

jandubois:
The Windows host-switch goroutine carries the WSL2 guest's networking, but an exit on error rather than cancellation went unnoticed: the guest silently lost DHCP/DNS/NAT while the hostagent reported healthy. Mark the bridge failed on unexpected exit and enqueue a reconcile; handleWatchedState restarts it, rate-limited to one per hostSwitchRetryInterval, with a Warning Event. Also surface a stuck vsock handshake, previously silent for up to five minutes.

### DIFF
--- a/rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
@@ -18,6 +18,7 @@ import (
 	"net/http/httptest"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/Microsoft/go-winio"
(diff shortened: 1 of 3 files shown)


rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -148,12 +212,10 @@ func (r *LimaVMReconciler) stopHostSwitch(name string) {
...
-// If the goroutine exits due to an error (not context cancellation), the
-// controller is not notified: the guest loses DHCP/DNS/NAT and must be
-// restarted manually. Integrating host-switch health into the controller's
-// state machine (enqueue a reconcile on unexpected exit) would allow
-// automatic recovery but requires non-trivial plumbing.
-func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger, done chan struct{}) {
+// On an unexpected exit (an error rather than cancellation) the guest loses
+// DHCP/DNS/NAT, so the goroutine marks the bridge failed and enqueues a
+// reconcile; handleWatchedState then restarts it via restartHostSwitch.
+func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger, name string, state *hostSwitchState, done chan struct{}) {
```
Nino-K, line 232: Four early returns in this function might also require setting the `state.failed` or calling the `enqueuReconcile`, because, if the handshake fails, the reconciler sees the bridge as healthy and might never retry?

jandubois: Yes, you are right. 3 of them should retry. `validateSubnet` validates a constant, so there is no point retrying that one.

Nino-K (dismissed):

Nino-K (approved):

PR #398, opened 2026-06-07
Move the daemon into an rdd/ subdirectory

jandubois:
## Why

We are merging rancher-desktop-app into this repository (to become
rancher-desktop-2). The app will live at the repository root, so this PR moves
the daemon's code into `rdd/` first, clearing the root of the directories that
would otherwise collide with the app (`pkg/`, `bats/`, `scripts/`, `docs/`).

This is step one. The app import and the reconciliation of the root-anchored
files come later.

## What moved

Everything daemon-specific moved under `rdd/`: `cmd/ pkg/ config/ bats/ docs/
scripts/ go.mod go.sum Makefile macos-entitlements.plist .golangci.yml .ltag/
NOTICE README.md`. The module path is unchanged, so imports are untouched.

## What stays at the root, and why

- `.github/` and `.gitmodules` are anchored there by GitHub and git.
- `LICENSE` covers the whole repository.
- A root `Makefile` runs the repo-wide checks. Spelling covers the entire tree
  (its two driver scripts stay in the root `scripts/`); the spelling config and
  the check-spelling submodule stay under `.github/`.
- `ltag` stays scoped to `rdd/`. rancher-desktop-app checks license headers
  with its own tooling and carries no SPDX headers, so a repo-wide ltag would
  flag every app file. It widens to the whole repository later.

## Commits

1. `Move daemon code into the rdd/ subdirectory` — pure `git mv` (306 files at
   100% rename similarity), kept separate so open PRs rebase cleanly.
2. `Run spelling from a root Makefile; keep ltag scoped to rdd/`.
3. `Point CI workflows at the rdd/ subdirectory`.

## Verified locally

`build-rdd`, `test`, `lint-rdd` (0 issues), `lint-bats`, and `check-ltag` all
pass from `rdd/`; the spelling generator walks the whole tree from the root.
The BATS suite and end-to-end spelling run in CI.

## Note for reviewers with open PRs

You will need to rebase. Edits to moved files follow the rename, but new files
land at their old paths and need a manual `git mv`, and the `bats.yaml` ->
`rdd-bats.yaml` rename may surface.


mook-as (approved):
LGTM, we can sort out the fallout later.

Note that I think we should delete BATS from the RDA repo before merging, to reduce confusion.

PR #405, opened 2026-06-08
Merge in rancher-desktop-app

mook-as:
This merges in the `rancher-desktop-app` repository.

Since the future of RDD is going to be `make` (using `msys` on Windows), some of the other bits are also done similarly.


jandubois (approved):
Thanks, LGTM

PR #409, opened 2026-06-09
Decompress distro image in-process when no system xz is found

jandubois:
## Problem

Lima's `Prepare` decompresses the distro image by shelling out to `xz -d`. Base macOS and a clean Windows ship no `xz`, so provisioning there fails unless the user installs one. rdd should depend only on what the OS ships.

## Change

When no `xz` is on `PATH`, the LimaVM controller decompresses the image in-process with a pure-Go decoder (`github.com/ulikunitz/xz`):

- Download and verify the image through Lima's own downloader, reusing its digest check and cache.
- Decompress to `<instance>/image`. `Prepare` then finds the ready image and skips its `xz` shell-out, because Lima skips a download whose destination already exists.
- When `xz` is present, rdd leaves the image alone so Lima keeps using it; its threaded decode is faster.

One integration point covers both the macOS/Linux `.raw.xz` and the Windows/WSL2 `.tar.xz` paths, which share `Prepare`.

## Why pure-Go (no CGO/liblzma)

`ulikunitz/xz` is already in the module graph (via `diskfs/go-diskfs`) and the decoder is pure Go, so this adds no build dependency and leaves `CGO_ENABLED` unchanged on every platform — no liblzma, no mingw, no Makefile changes.

The decoder buffers both sides. Unbuffered, `ulikunitz/xz` issues tiny reads and spends most of its time in syscalls. On the real 1.4 GB arm64 distro image (8-core M-series), byte-exact decode times:

| Decoder | Time |
|---|---|
| system `xz --threads=0` (threaded) | 5.6s |
| system `xz --threads=1` | 10.2s |
| `ulikunitz/xz`, 4 MB buffered | 18.9s |
| `ulikunitz/xz`, unbuffered | 95.3s |

The fallback is single-threaded, so a slow host without `xz` pays more — a one-time cost behind an already-multi-minute download. `DecompressFile` writes to a temp file and renames it into place, so an interrupted decode leaves no partial image.

## CI

The new `pkg/xz` test compares decoded bytes against an LF fixture, which exposed two Windows CI gaps:

- `test-rdd` and `test-wix-helper` never set `core.autocrlf=false` and `core.eol=lf` before checkout, unlike the lint jobs, so Windows checked the fixture out as CRLF and the byte comparison failed. Both jobs now run the same line-ending step.
- The Windows long-path setup used `core.longpath` (singular), which git silently ignores. The lint and bats workflows now use the real key, `core.longpaths`.

## Testing

- `pkg/xz` unit tests: decode an `xz`-CLI-produced fixture (verifies real-xz compatibility), a larger-than-buffer round-trip, a truncated-input error, and temp-file cleanup on both success and failure.
- `make build-rdd`, `go vet`, `golangci-lint` (0 issues), and the existing `limavm` controller tests pass.
- Manual end-to-end with `xz` removed from `PATH` is not yet run — flagging for review.

## Follow-up

This is shaped to lift into upstream Lima: the decoder is standalone, and the fallback mirrors what Lima's `decompressLocal` could do itself (`exec.LookPath("xz")`, else decode in-process). The rdd hook retires if Lima adopts it.

### DIFF
--- a/rdd/pkg/controllers/lima/limavm/controllers/image_decompress_test.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/image_decompress_test.go
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package controllers
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+func TestLinkWSL2BaseDisk(t *testing.T) {
+	writeImage := func(t *testing.T, dir string) string {
+		t.Helper()
+		imagePath := filepath.Join(dir, filenames.Image)
+		assert.NilError(t, os.WriteFile(imagePath, []byte("decompressed image"), 0o644))
+		return imagePath
+	}
+
+	t.Run("WSL2 hardlinks basedisk to the image", func(t *testing.T) {
+		dir := t.TempDir()
+		imagePath := writeImage(t, dir)
+		inst := &limatype.Instance{Dir: dir, VMType: limatype.WSL2}
+
+		assert.NilError(t, linkWSL2BaseDisk(inst, imagePath))
+
+		baseDisk := filepath.Join(dir, filenames.BaseDiskLegacy)
+		got, err := os.ReadFile(baseDisk)
+		assert.NilError(t, err)
+		assert.Equal(t, string(got), "decompressed image")
+		assert.Assert(t, sameFile(t, imagePath, baseDisk), "basedisk must be a hardlink to image")
+	})
+
+	t.Run("non-WSL2 leaves basedisk absent", func(t *testing.T) {
+		dir := t.TempDir()
+		imagePath := writeImage(t, dir)
+		inst := &limatype.Instance{Dir: dir, VMType: limatype.VZ}
+
+		assert.NilError(t, linkWSL2BaseDisk(inst, imagePath))
+
+		_, err := os.Stat(filepath.Join(dir, filenames.BaseDiskLegacy))
+		assert.Assert(t, cmp.ErrorIs(err, os.ErrNotExist), "basedisk must not exist for vz")
+	})
+
+	t.Run("existing basedisk is left untouched", func(t *testing.T) {
+		dir := t.TempDir()
+		imagePath := writeImage(t, dir)
+		baseDisk := filepath.Join(dir, filenames.BaseDiskLegacy)
+		assert.NilError(t, os.WriteFile(baseDisk, []byte("preexisting"), 0o644))
+		inst := &limatype.Instance{Dir: dir, VMType: limatype.WSL2}
+
+		assert.NilError(t, linkWSL2BaseDisk(inst, imagePath))
+
+		got, err := os.ReadFile(baseDisk)
+		assert.NilError(t, err)
+		assert.Equal(t, string(got), "preexisting")
+	})
+
+	t.Run("missing image is a no-op", func(t *testing.T) {
+		dir := t.TempDir()
+		inst := &limatype.Instance{Dir: dir, VMType: limatype.WSL2}
+
+		assert.NilError(t, linkWSL2BaseDisk(inst, filepath.Join(dir, filenames.Image)))
+
+		_, err := os.Stat(filepath.Join(dir, filenames.BaseDiskLegacy))
+		assert.Assert(t, cmp.ErrorIs(err, os.ErrNotExist), "basedisk must not exist when image is absent")
+	})
+}
+
+func sameFile(t *testing.T, a, b string) bool {
+	t.Helper()
+	ai, err := os.Stat(a)
+	assert.NilError(t, err)
+	bi, err := os.Stat(b)
+	assert.NilError(t, err)
+	return os.SameFile(ai, bi)
+}
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
@@ -10,12 +10,15 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/lima-vm/lima/v2/pkg/fileutils"
 	limainstance "github.com/lima-vm/lima/v2/pkg/instance"
+	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/store"
 
@@ -38,6 +41,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/util/process"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/xz"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -378,6 +396,85 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+// that case with its own decompressor.
+func decompressImage(ctx context.Context, inst *limatype.Instance, imagePath string) error {
+	logger := log.FromContext(ctx)
+	arch := *inst.Config.Arch
+	for _, img := range inst.Config.Images {
+		if img.File.Arch != arch {
+			continue
+		}
+		// rdd's templates point at remote .xz images. A non-.xz or local
+		// location falls through to Prepare, which runs its own decompressor.
```
mook-as: Hmm, I don't see how a local location would fall through, but maybe I just don't understand `img.File.Location` correctly.

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -378,6 +396,85 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		return nil
+	}
+	if _, err := os.Stat(imagePath); err != nil {
+		return nil
+	}
+	baseDiskPath := filepath.Join(inst.Dir, filenames.BaseDiskLegacy)
+	if _, err := os.Stat(baseDiskPath); err == nil {
+		return nil
+	}
+	if err := os.Link(imagePath, baseDiskPath); err != nil {
```
mook-as, line 472: Would that work everywhere (specifically, user is not an admin, and the various developer mode switches are not set)?

rdd/pkg/xz/xz_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"testing"
+
+	ulikunitz "github.com/ulikunitz/xz"
+	"gotest.tools/v3/assert"
+)
+
+// sample.txt.xz is produced by the xz CLI (xz -9), so this verifies
+// compatibility with real xz output rather than only our own round-trip.
+func TestDecompressFixture(t *testing.T) {
+	compressed := readFixture(t, "sample.txt.xz")
```
mook-as: I would feel a lot better if we shelled out to `xz` to create a test archive on the fly. That is, after all, how xz (the project) got into trouble last time :)

rdd/pkg/xz/testdata/.gitattributes
mook-as, line 1: Should this be in the top level? I can't see a `*.xz` that's ever not binary data.

rdd/pkg/xz/xz_test.go
```diff
@@ -0,0 +1,125 @@
...
+	assert.NilError(t, Decompress(context.Background(), bytes.NewReader(compressed), &got))
+	assert.DeepEqual(t, got.Bytes(), fixturePlaintext())
+}
+
+func TestDecompressRoundTrip(t *testing.T) {
+	// Larger than bufSize so the buffered input path is exercised.
+	want := bytes.Repeat([]byte("rancher-desktop xz round-trip payload\n"), 200_000)
+
+	var compressed bytes.Buffer
+	w, err := ulikunitz.NewWriter(&compressed)
```
mook-as: I don't see the point of testing if `ulikunitz/xz` round trips successfully; we don't expose an compression function.

rdd/pkg/xz/xz_test.go
```diff
@@ -0,0 +1,125 @@
...
+	assert.NilError(t, Decompress(context.Background(), bytes.NewReader(compressed.Bytes()), &got))
+	assert.Equal(t, got.Len(), len(want))
+	assert.Assert(t, bytes.Equal(got.Bytes(), want))
+}
+
+func TestDecompressTruncated(t *testing.T) {
+	compressed := readFixture(t, "sample.txt.xz")
+
+	var got bytes.Buffer
+	err := Decompress(context.Background(), bytes.NewReader(compressed[:len(compressed)/2]), &got)
```
mook-as: There's [`io.LimitReader`](https://pkg.go.dev/io@go1.26.4#LimitedReader) too.

Also, both here and elsewhere, should use [`t.Context()`](https://pkg.go.dev/testing@go1.26.4#T.Context) instead of `context.Background()`. (If the test times out, `t.Context()` should cancel when `context.Background()` doesn't.)

rdd/pkg/xz/xz_test.go
```diff
@@ -0,0 +1,125 @@
...
+	dst := filepath.Join(dir, "image")
+	err := DecompressFile(ctx, src, dst)
+	assert.Assert(t, errors.Is(err, context.Canceled), "want context.Canceled, got %v", err)
+
+	_, statErr := os.Stat(dst)
+	assert.Assert(t, os.IsNotExist(statErr), "dst must not exist after a canceled decode")
+	assertNoTempFiles(t, dir)
+}
+
+func readFixture(t *testing.T, name string) []byte {
```
mook-as: Does testing support `embed`?

rdd/pkg/xz/xz_test.go
```diff
@@ -0,0 +1,125 @@
...
+// out line endings on the test host.
+func fixturePlaintext() []byte {
+	const block = "rancher-desktop-daemon in-process xz decoder fixture.\n" +
+		"The quick brown fox jumps over the lazy dog.\n"
+	return bytes.Repeat([]byte(block), 200)
+}
+
+func assertNoTempFiles(t *testing.T, dir string) {
+	t.Helper()
+	entries, err := os.ReadDir(dir)
```
mook-as: Would [`filepath.Glob`](https://pkg.go.dev/path/filepath@go1.26.4#Glob) be better here?

rdd/pkg/xz/xz.go
```diff
@@ -0,0 +1,92 @@
...
+		return fmt.Errorf("decompressing xz stream: %w", err)
+	}
+	return nil
+}
+
+// ctxReader makes the otherwise-uninterruptible decode cancellable: each Read
+// returns the context error once ctx is cancelled, which unwinds io.Copy. The
+// decode is single-threaded and can run for tens of seconds on a large image,
+// so this is what lets a service shutdown propagate instead of blocking on it.
+type ctxReader struct {
```
mook-as, line 42: It would be nice if this could implement `WriteTo`, but that doesn't take a context so I guess it wouldn't work.

rdd/pkg/controllers/lima/limavm/controllers/image_decompress_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	t.Run("non-WSL2 leaves basedisk absent", func(t *testing.T) {
+		dir := t.TempDir()
+		imagePath := writeImage(t, dir)
+		inst := &limatype.Instance{Dir: dir, VMType: limatype.VZ}
+
+		assert.NilError(t, linkWSL2BaseDisk(inst, imagePath))
+
+		_, err := os.Stat(filepath.Join(dir, filenames.BaseDiskLegacy))
+		assert.Assert(t, os.IsNotExist(err), "basedisk must not exist for vz")
```
mook-as: Should probably use [`cmp.ErrorIs`](https://pkg.go.dev/gotest.tools/v3@v3.5.2/assert/cmp#ErrorIs) here.

mook-as (changes requested):

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yes, only the file extension matters. Will fix the comment.

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: Yes, everyone can create hardlinks. Only symlinks are a problem.

rdd/pkg/xz/xz_test.go
jandubois: Yeah, better not to have binary blobs in the repo

rdd/pkg/xz/xz_test.go
jandubois: I don't see that `LimitReader` buys us much, but will use `t.Context()` everywhere.

rdd/pkg/xz/xz_test.go
jandubois: No longer relevant if we create the fixture by shelling out to `xz`.

rdd/pkg/xz/testdata/.gitattributes
jandubois: Removing the whole file because we don't want binary blobs in the repo.

jandubois (commented):
Made changes based on review in a new commit. Let me know if you want me to squash to remove the binary blob even from history!

mook-as (changes requested):
LGTM, but yes please squash.

mook-as (approved):

PR #411, opened 2026-06-09
RDA: locate RDD

mook-as:
- On postinstall, run `make -C rdd build-rdd` to ensure RDD is built.
- In `yarn dev`, locate the RDD artifact built from the above.
- When packaging, package the same RDD artifact.

This should mean a fresh checkout plus `yarn && yarn dev` should still work fine without needing any manual steps to build RDD.

### DIFF
--- a/.github/actions/yarn-install/action.yaml
+++ b/.github/actions/yarn-install/action.yaml
@@ -33,6 +33,11 @@ runs:
     shell: powershell
     run: .\scripts\windows-setup.ps1 -SkipVisualStudio -SkipTools
 
+  - name: Build RDD
+    shell: bash
+    run: make build-rdd
+    working-directory: rdd
+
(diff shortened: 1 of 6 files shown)


jandubois (changes requested):
When `rdd` is resigned, it needs to get the `rdd/macos-entitlements.plist` entitlements reapplied, and not the defaults.

The entitlements could probably be mirrored in `build/signing-config-mac.yaml`, but add a comment to both files that they need to be kept in sync.

mook-as:
For resigning, are you talking about `yarn package`? That would be part of #415.  For `yarn dev`, this just runs `make -C rdd build-rdd`, so it doesn't involve resigning.

Of course, I may be missing something…

jandubois (approved):
Thanks, LGTM

PR #412, opened 2026-06-09
RDA: Drop `M1` build environment variable

mook-as:
The architecture of the built artifact is now based on the `GOARCH` environment variable, and if not provided, the host architecture.


jandubois (approved):
Thanks, LGTM

Haven't tried cross-signing yet, but looks like it should work.

PR #413, opened 2026-06-09
CI: Skip bats when RDD not modified

mook-as:
This should save some CI time; there is no need to run BATS when its inputs solely depend on the RDD part of the tree.

We couldn't use [paths-ignore](https://github.com/rancher-sandbox/rancher-desktop-daemon/blob/b87f0253e2196a3bf85a059e538c4f30bfbe3adb/.github/workflows/paths-ignore.yaml) because that ignores files, instead of only allowing them.

### DIFF
--- a/.github/workflows/rdd-bats.yaml
+++ b/.github/workflows/rdd-bats.yaml
@@ -9,6 +9,26 @@ permissions:
   contents: read
 
 jobs:
+  check-paths:
+    name: Check for relevant changes
+    # Due to required status checks, these still need to run even if there are
+    # no relevant changes.
+    outputs:
+      should-run: ${{ steps.check.outputs.result }}
+    runs-on: ubuntu-latest
+    steps:
+    - id: check
+      continue-on-error: true
+      run: |
+        set -o errexit -o pipefail -o xtrace
+        if test "${{ github.event_name }}" != 'pull_request'; then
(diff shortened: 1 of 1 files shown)


.github/workflows/rdd-bats.yaml
```diff
@@ -34,8 +47,8 @@ jobs:
         GH_TOKEN: ${{ github.token }}
   bats:
     name: BATS
-    needs: get-pr-number
-    if: ${{ ! needs.get-pr-number.outputs.number }}
+    needs: [get-pr-number, check-paths]
+    if: ${{ needs.check-paths.outputs.should-run && ! needs.get-pr-number.outputs.number }}
```
jandubois: How would this work? The BATS tests are required checks, if you skip them, then GitHub will block merging (I think).

I believe we will need to have the tests return with a 0 exit code instead of running the tests, if we want to skip them, but still having them gate the merge.

mook-as: Ah, I thought having the workflow was enough; but that's why this is draft :) Not to mention I didn't do the step correctly in the first place…

jandubois (approved):
Thanks, LGTM

I thought `needs.check-paths.outputs.should-run` sounds a bit generic because we might eventually have different versions of `check-paths` to skip different jobs. But we can just rename it when we actually have other ones, so good for now.

PR #415, opened 2026-06-10
Set up release packaging

mook-as:
This updates the signing workflow to also produce signed rdd binaries for macOS & Windows (by grabbing them out of `…/resources/…/bin`).  This needs to go on top of #411, as that starts shipping RDD in the installers.


jandubois (approved):
Thanks, LGTM

PR #416, opened 2026-06-10
Poll for dangling-mirror reconciliation instead of racing it

jandubois:
Image tags are updated asynchronously, so retry until the reconciler has done its job.

### DIFF
--- a/rdd/bats/tests/32-app-controllers/engine-docker.bats
+++ b/rdd/bats/tests/32-app-controllers/engine-docker.bats
@@ -101,6 +101,29 @@ do_websocket() { # endpoint
     assert_output
 }
 
+# Assert that exactly one Image mirror exists for the given image ID and
+# that its RepoTag is <expected-tag> (empty = the dangling mirror). All
+# mirrors of an image share status.id, so a check right after an untag or
+# tag races the engine controller's pruning of the stale mirror — calls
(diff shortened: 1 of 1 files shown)


rdd/bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -101,6 +101,29 @@ do_websocket() { # endpoint
     assert_output
 }
 
+# Assert that exactly one Image mirror exists for the given image ID and
+# that its RepoTag is <expected-tag> (empty = the dangling mirror). All
+# mirrors of an image share status.id, so a check right after an untag or
+# tag races the engine controller's pruning of the stale mirror — poll
+# this with `try`.
```
mook-as:
```suggestion
# tag races the engine controller's pruning of the stale mirror — calls
# to `assert_single_mirror` should be wrapped in a `try`.
```
The original wording made me think this implemented the polling; trying to rephrase to make it clearer it's a thing the caller is expected to do.

mook-as (dismissed):
LGTM; I'd like the comment slightly rephrased, though.

rdd/bats/tests/32-app-controllers/engine-docker.bats
jandubois: Applied, please approve again!

mook-as (approved):

PR #417, opened 2026-06-10
CI: Run darwin/x86_64 packaging on arm

mook-as:
Since the arm runners are much faster and available at this point, we can run the package step there even for x86_64 builds, because it does not require running the build artifacts.

~Please test the built result on a real macOS x86_64 machine to make sure it still works fine.  Inspecting the built binaries says all the Mach-O files are x86_64.~ Tested pm a x86_64 mac, it seems to start with no complaints.


jandubois (approved):
Thanks, LGTM

PR #418, opened 2026-06-10
Unexpose all forwarded ports when the host-switch exits

jandubois:
gvisor-tap-vsock never closes the host listener for an exposed port when the virtual network is torn down. On a VM restart a still-exposed port leaks its listener: the next boot can't re-expose it (EADDRINUSE) and connections route into the dead network ("no route to host") until rdd exits. Unexpose every forward when the host-switch exits.

### DIFF
--- a/rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
@@ -7,12 +7,15 @@
 package controllers
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
+	"net/http/httptest"
 	"sync"
 	"time"
 
@@ -175,6 +178,7 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
 		logger.Error(err, "Failed to create virtual network")
 		return
 	}
+	defer unexposeAllForwards(logger, vn)
 
 	// Set up the API listener before starting errgroup goroutines so a
 	// failure here does not leak goroutines.
@@ -286,6 +290,62 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
 		},
 		GatewayVirtualIPs: []string{subnet.StaticDNSHost},
 	}
+}
+
+// unexposeAllForwards closes every host listener the forwarder API opened on
+// this virtual network. gvisor-tap-vsock never closes these listeners when the
+// network is torn down, and the host-switch runs in the long-lived controller,
+// so a port still exposed at teardown keeps its host port bound after the VM
+// is gone: the next boot cannot re-expose it (EADDRINUSE), and connections
+// route into the dead network with no guest attached ("no route to host")
+// until rdd exits.
+func unexposeAllForwards(logger logr.Logger, vn *virtualnetwork.VirtualNetwork) {
+	var forwards []struct {
+		Local    string `json:"local"`
+		Protocol string `json:"protocol"`
+	}
+	all, err := requestVN(vn, http.MethodGet, "/services/forwarder/all", nil)
+	if err != nil {
+		logger.Error(err, "Failed to list exposed forwards for teardown")
+		return
+	}
+	if err := json.Unmarshal(all.Body.Bytes(), &forwards); err != nil {
+		logger.Error(err, "Failed to list exposed forwards for teardown")
+		return
+	}
+	for _, fwd := range forwards {
+		body, err := json.Marshal(fwd)
+		if err != nil {
+			logger.Error(err, "Failed to marshal forward for teardown",
+				"local", fwd.Local, "protocol", fwd.Protocol)
+			continue
+		}
+		rec, err := requestVN(vn, http.MethodPost, "/services/forwarder/unexpose", body)
+		if err != nil {
+			logger.Error(err, "Failed to unexpose forward at teardown",
+				"local", fwd.Local, "protocol", fwd.Protocol)
+			continue
+		}
+		logger.Info("Unexposed forward at teardown",
+			"local", fwd.Local, "protocol", fwd.Protocol, "status", rec.Code)
+	}
+}
+
+// requestVN performs an in-process request against the virtual network's
+// services mux and returns the response recorder. The mux serves the forwarder
+// API under /services/forwarder/. A nil body sends an empty request.
+func requestVN(vn *virtualnetwork.VirtualNetwork, method, path string, body []byte) (*httptest.ResponseRecorder, error) {
+	var r io.Reader = http.NoBody
+	if body != nil {
+		r = bytes.NewReader(body)
+	}
+	req, err := http.NewRequestWithContext(context.Background(), method, path, r)
+	if err != nil {
+		return nil, err
+	}
+	rec := httptest.NewRecorder()
+	vn.ServicesMux().ServeHTTP(rec, req)
+	return rec, nil
 }
 
 // --- Vsock handshake ---


rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -288,6 +294,67 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
...
+		logger.Info("Unexposed forward at teardown",
+			"local", fwd.Local, "protocol", fwd.Protocol, "status", rec.Code)
+	}
+}
+
+// queryVN performs an in-process GET against the virtual network's services
+// mux and returns the response body. The mux serves the forwarder API under
+// /services/forwarder/.
+func queryVN(vn *virtualnetwork.VirtualNetwork, path string) (string, error) {
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, path, http.NoBody)
```
mook-as: Surely we can thread the context from `runHostSwitch` in here?

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -288,6 +294,67 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
...
+	if err != nil {
+		return "", err
+	}
+	rec := httptest.NewRecorder()
```
mook-as: Using a testing facility in production seems weird, but I guess it works…

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -288,6 +294,67 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
...
+	if err != nil {
+		logger.Error(err, "Failed to list exposed forwards for teardown")
+		return
+	}
+	if err := json.Unmarshal([]byte(all), &forwards); err != nil {
+		logger.Error(err, "Failed to list exposed forwards for teardown")
+		return
+	}
+	for _, fwd := range forwards {
+		body, err := json.Marshal(map[string]string{"local": fwd.Local, "protocol": fwd.Protocol})
```
mook-as: Can we marshal `fwd` directly? It has the appropriate tags, but it's an unexported struct, so I'm not sure…

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -288,6 +294,67 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
...
+		if err != nil {
+			continue
```
mook-as, line 321: Should there be some sort of logging? It basically means the port is stuck.

mook-as (dismissed):
Manually unexposing everything feels iffy, but I guess the only other way is to just spawn a child process to handle the host-switch, and that's kind of annoying…

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
jandubois: I actually had a copy of the code at first, but decided that there is nothing wrong with using `httptest` in regular code. See my first forced-push: https://github.com/rancher-sandbox/rancher-desktop-2/compare/ba6851ecf18a58b3b4166cf68e7e9acbbbc0bf85..de3928a680bb8fd976867513d9fb5761f95e614e

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
jandubois: No, we can't. `unexposeAllForwards` is called after `g.Wait()`, which only exits once the context has been cancelled.

jandubois (commented):
> Manually unexposing everything feels iffy, but I guess the only other way is to just spawn a child process to handle the host-switch, and that's kind of annoying…

There is no API to unexpose all remaining forwards.

But you are correct that host-switch should be part of `rdd hostagent` and not run as part of the reconciler main daemon. I will make that change, but I wanted to wrap up the current issues before starting a new construction site. #432

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -217,10 +230,29 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
...
-			} else {
-				logger.Info("Accepted vsock connection")
+			start := time.Now()
+			err = vn.AcceptStdio(ctx, conn)
+			elapsed := time.Since(start)
+			switch {
+			case err == nil:
+				immediateDrops = 0
+				logger.Info("Accepted vsock connection", "duration", elapsed.String())
+			case elapsed < relayMinDuration:
```
mook-as: We're judging no frames by elapsed time; that's a pretty weird metric.  It would be better if the error messages actually just say it's time-based rather than confidently claim no frames.

It would be better to _actually_ track if frames were received, but that's probably not worth the effort.

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -248,9 +280,26 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
 		return nil
 	})
 
+	// Snapshot the CAM table and stack counters once a minute, so a failing run
```
mook-as:
```suggestion
	// Snapshot diagnostics information once a minute, so a failing run
```

That seems like a lot of non-optional debugging information; should that be more like trace level? Compare with https://github.com/rancher-sandbox/rancher-desktop-opensuse/pull/48.

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -248,9 +280,26 @@ func (r *LimaVMReconciler) runHostSwitch(ctx context.Context, logger logr.Logger
...
+				logger.Info("Virtual network diagnostics", vnDiagnostics(vn)...)
+			}
+		}
+	})
+
 	logger.Info("Host-switch running", "subnet", subnet.SubnetCIDR, "gateway", subnet.GatewayIP)
 
-	if err := g.Wait(); err != nil {
+	err = g.Wait()
+	unexposeAllForwards(logger, vn)
```
mook-as: Should this just be called in a `defer` instead? The logging doesn't seem to really need to care about ordering with the cleanup.

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
```diff
@@ -288,6 +337,96 @@ func newVirtualNetworkConfig(subnet hostSwitchSubnet) types.Configuration {
...
+	if err != nil {
+		return "", err
+	}
+	rec := httptest.NewRecorder()
+	vn.ServicesMux().ServeHTTP(rec, req)
+	return strings.TrimSpace(rec.Body.String()), nil
+}
+
+// postVN performs an in-process POST against the virtual network's services mux.
+func postVN(vn *virtualnetwork.VirtualNetwork, path string, body []byte) (*httptest.ResponseRecorder, error) {
```
mook-as: The POST and GET helpers are so close, it makes me itch to have a singular helper instead…

mook-as (changes requested):
The `AcceptStdio` changes don't look like they belong in this PR; they probably help host-switch, but either it needs to be in its own PR, or the PR description should be updated.

jandubois:
I've split out the diagnostics changes and left only the bug fix. Applied the 2 requested changes from code review. PTAL!

mook-as (approved):

rdd/pkg/controllers/lima/limavm/controllers/hostswitch_windows.go
jandubois: The diagnostic part is now in #437.

It logs at `V(1)` now (enabled in CI, disabled for regular user) and has reworded diagnostics.

PR #430, opened 2026-06-10
Bump embedded distro to v0.2.6

jandubois:
The embedded distro was two releases behind at `v0.2.4`. `v0.2.6` adds opt-in vm-switch logging to `network-setup` — the `--trace-packets` and `--vm-switch-logfile-append` flags — and a `NETWORK_SETUP_EXTRA_ARGS` hook on the `systemd` unit, so a drop-in can pass those flags without overriding the unit's ExecStart. All default off, so the image behaves as before until a drop-in opts in.


mook-as (approved):

jandubois:
I believe the CI error has been fixed by #416, so closing and reopening to run against a new merge base.

PR #431, opened 2026-06-11
Capture the guest vm-switch log on CI

jandubois:
Capture vm-switch diagnostics on CI under `RDD_KEEP_LOGS` so WSL2 networking failures stay debuggable after teardown: redirect the vm-switch relay log to the collected log dir (appended across restarts, with optional packet tracing under `RDD_TRACE_PACKETS` via v0.2.6's `NETWORK_SETUP_EXTRA_ARGS` hook), and stream the guest networking journals and `lima-init.log` there too.

### DIFF
--- a/rdd/pkg/controllers/app/app/controllers/app_controller_test.go
+++ b/rdd/pkg/controllers/app/app/controllers/app_controller_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
 	limav1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/lima/v1alpha1"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
 )
 
 // fakeDiscovery satisfies ControllerDiscovery for unit tests.
@@ -627,6 +628,17 @@ func Test_networkSetupExtraArgs(t *testing.T) {
 		t.Setenv("RDD_KEEP_LOGS", "1")
 		t.Setenv("RDD_TRACE_PACKETS", "1")
 		assert.Equal(t, networkSetupExtraArgs(), "--vm-switch-logfile-append --trace-packets")
+	})
+}
+
+func Test_guestLogDir(t *testing.T) {
+	t.Run("unset RDD_KEEP_LOGS yields no log dir", func(t *testing.T) {
+		t.Setenv("RDD_KEEP_LOGS", "")
+		assert.Equal(t, guestLogDir(), "")
+	})
+	t.Run("set RDD_KEEP_LOGS yields the instance log dir", func(t *testing.T) {
+		t.Setenv("RDD_KEEP_LOGS", "1")
+		assert.Equal(t, guestLogDir(), toLinuxPath(instance.LogDir()))
 	})
 }
 
--- a/rdd/pkg/controllers/app/app/controllers/app_controller_test.go
+++ b/rdd/pkg/controllers/app/app/controllers/app_controller_test.go
@@ -619,6 +619,11 @@ func Test_networkSetupExtraArgs(t *testing.T) {
 		t.Setenv("RDD_TRACE_PACKETS", "")
 		assert.Equal(t, networkSetupExtraArgs(), "")
 	})
+	t.Run("RDD_TRACE_PACKETS without RDD_KEEP_LOGS yields no extra args", func(t *testing.T) {
+		t.Setenv("RDD_KEEP_LOGS", "")
+		t.Setenv("RDD_TRACE_PACKETS", "1")
+		assert.Equal(t, networkSetupExtraArgs(), "")
+	})
 	t.Run("RDD_KEEP_LOGS without tracing yields append only", func(t *testing.T) {
 		t.Setenv("RDD_KEEP_LOGS", "1")
 		t.Setenv("RDD_TRACE_PACKETS", "")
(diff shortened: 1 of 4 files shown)


rdd/pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -551,3 +556,82 @@ func Test_Reconcile_clearsKubernetesPortOnDisable(t *testing.T) {
 	assert.Equal(t, result.Status.KubernetesPort, 0,
 		"expected Status.KubernetesPort to be cleared when Kubernetes is disabled")
 }
+
+func Test_networkSetupExtraArgs(t *testing.T) {
```
mook-as, line 616: You're missing `RDD_KEEP_LOGS= RDD_TRACE_PACKETS=1`; it looks like that also outputs nothing, but having a test makes it look more intentional.

I'm not sure why we can't have packet trace without keeping logs… But you're pretty consistent about it in the PR, so that's fine.

rdd/pkg/controllers/app/app/controllers/app_controller_test.go
```diff
@@ -551,3 +556,82 @@ func Test_Reconcile_clearsKubernetesPortOnDisable(t *testing.T) {
...
+	}
+	assert.Assert(t, content != "", "no provision entry writes %s", dropinPath)
+
+	// Empty param renders the distro's own default (an empty value).
+	blank := renderProvisionContent(t, content, map[string]string{"NETWORK_SETUP_EXTRA_ARGS": ""})
+	assert.Assert(t, strings.Contains(blank, `Environment=NETWORK_SETUP_EXTRA_ARGS=""`),
+		"got:\n%s", blank)
+
+	// Quoted as a whole so the space stays one env var; the unit's unquoted
+	// $NETWORK_SETUP_EXTRA_ARGS splits it back into args.
```
mook-as: The key is more that systemd treats `$NETWORK_SETUP_EXTRA_ARGS` (potentially multiple arguments) and `${NETWORK_SETUP_EXTRA_ARGS}` (always one argument, potentially empty) differently; quoting isn't involved.

rdd/pkg/controllers/app/app/lima-template.yaml
```diff
@@ -44,6 +44,43 @@ provision:
...
     KUBERNETES_PORT={{.Param.KUBERNETES_PORT}}
+# Redirect vm-switch's log to a host-mounted file (under the service log dir) so
+# it survives VM teardown and is collected with the other instance logs. The
+# distro's network-setup.service defaults VM_SWITCH_LOG to an in-guest path; this
+# drop-in overrides it and is written before systemd loads the unit.
+- mode: data
+  path: /etc/systemd/system/network-setup.service.d/log-path.conf
+  content: |
+    [Service]
+    Environment=VM_SWITCH_LOG={{printf "%q" .Param.VM_SWITCH_LOG}}
```
mook-as: [`man systemd.exec`](https://manpages.opensuse.org/Tumbleweed/systemd/systemd.exec.5.en.html) actually says:
> If you need to assign a value containing spaces or the equals sign to a variable, put quotes around the whole assignment. 

With an example of:
> `Environment="VAR1=word1 word2" VAR2=word3 "VAR3=$word 5 6"`

I'm also not sure `%q` quoting is correct here.  But we're probably not getting any backslashes here, and that's the first character I can think of that might make a difference…

rdd/pkg/controllers/app/app/lima-template.yaml
```diff
@@ -44,6 +44,43 @@ provision:
     KUBERNETES_ENABLED={{.Param.KUBERNETES_ENABLED}}
     KUBERNETES_VERSION={{.Param.KUBERNETES_VERSION}}
     KUBERNETES_PORT={{.Param.KUBERNETES_PORT}}
+# Redirect vm-switch's log to a host-mounted file (under the service log dir) so
+# it survives VM teardown and is collected with the other instance logs. The
+# distro's network-setup.service defaults VM_SWITCH_LOG to an in-guest path; this
+# drop-in overrides it and is written before systemd loads the unit.
+- mode: data
```
mook-as, line 53: Do we need any comments that this is Windows-only? Not that it really matters; on macOS/Linux the file would just go unused, because we don't run `network-setup.service` via `ConditionVirtualization=wsl`.

For the existing comment, it may be clearer to put that _inside_ the list item:

```yaml
- # Redirect …
  mode: data
```

rdd/pkg/controllers/app/app/lima-template.yaml
```diff
@@ -44,6 +44,43 @@ provision:
...
+  path: /etc/systemd/system/network-setup.service.d/log-path.conf
+  content: |
+    [Service]
+    Environment=VM_SWITCH_LOG={{printf "%q" .Param.VM_SWITCH_LOG}}
+# Diagnostic vm-switch logging flags from networkSetupExtraArgs(); empty
+# (the distro default) outside RDD_KEEP_LOGS.
```
mook-as:
```suggestion
# (the distro default) when RDD_KEEP_LOGS is not set.
```

rdd/pkg/controllers/app/app/lima-template.yaml
```diff
@@ -44,6 +44,43 @@ provision:
...
+  script: |
+    #!/bin/bash
+
+    set -o errexit -o nounset -o pipefail -o xtrace
+
+    if [[ -z $PARAM_GUEST_LOG_DIR ]]; then
+        exit 0
+    fi
+
+    if ! systemctl is-active --quiet rdd-journal-capture.service; then
```
mook-as, line 78: Perhaps these can be generalized to `log-stream-journald@.service` and `log-stream-file@.service`, and then we can just have `lima-init.service` have `Wants=log-stream-journald@lima-init` etc? The implementation can use `ConditionEnvironment=` to only run them when `PARAM_GUEST_LOG_DIR` is set.

In general, I think we should avoid using provisioning scripts when they are better done via unit files.

mook-as (changes requested):

rdd/pkg/controllers/app/app/lima-template.yaml
jandubois: Added a comment, but kept the comments on top, for consistency with the rest of the scripts.

rdd/pkg/controllers/app/app/lima-template.yaml
jandubois: I agree, but let's do this in a separate PR (and later). #445

mook-as (approved):

PR #435, opened 2026-06-11
Gate KubernetesReady until a node is Ready

jandubois:
`KubernetesReady` went `True` as soon as the k3s API server answered `/healthz`. The API server answers before the node registers, so `rdd set` returned `Settled` while a scheduled pod had no node to run on.

Gate the first `Ready` on the node being `Ready`; once `Ready`, API-server liveness alone keeps the condition `True`, so a transient node blip does not flap `Settled`. The new reason `WaitingForNode` reports the blocking gate.

Workload-level readiness (coredns, traefik) is deliberately not gated: it would slow startup for every user, and consumers that need it can wait for those Deployments themselves. The BATS tests will gain explicit coredns/traefik waits separately.

### DIFF
--- a/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
+++ b/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
@@ -17,9 +17,11 @@ import (
 	"syscall"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -120,6 +122,11 @@ type KubernetesReconciler struct {
 	// probeFn lets tests inject probe results without making real HTTP calls.
 	// Production wiring leaves it nil; Reconcile then calls probeK3sAPI.
 	probeFn func(ctx context.Context) (probeResult, error)
+
+	// clientsetFn lets tests inject a fake typed client so the node check
+	// runs against in-memory objects. Production wiring leaves it nil;
+	// checkNodeReady then builds a client from the instance kubeconfig.
+	clientsetFn func() (kubernetes.Interface, error)
 }
 
 // probe dispatches to probeFn if injected (tests) or probeK3sAPI otherwise.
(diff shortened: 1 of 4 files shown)


jandubois:
Quick update: this PR does **not** fix the BATS test in #395.

The change in this code does work though, and I think it is still the way the `KubernetesReady` condition should be reported.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -130,6 +137,68 @@ func (r *KubernetesReconciler) probe(ctx context.Context) (probeResult, error) {
...
+	nodes, err := cs.CoreV1().Nodes().List(checkCtx, metav1.ListOptions{})
+	if err != nil {
+		return false, appv1alpha1.AppKubernetesReasonWaitingForNode,
+			fmt.Sprintf("Cannot list nodes: %v", err)
+	}
+	if len(nodes.Items) == 0 {
+		return false, appv1alpha1.AppKubernetesReasonWaitingForNode, "No node has registered yet"
+	}
+	for i := range nodes.Items {
+		if !nodeReady(&nodes.Items[i]) {
```
mook-as: Do we need to wait for _all_ nodes to be ready? Wouldn't any ready node be good enough?

(I realize that this is moot right now, as there is only one node.)

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -130,6 +137,68 @@ func (r *KubernetesReconciler) probe(ctx context.Context) (probeResult, error) {
...
+				fmt.Sprintf("Node %q is not Ready", nodes.Items[i].Name)
+		}
+	}
+
+	return true, "", ""
+}
+
+// getClientset returns the injected fake client in tests or a real client
+// built from the instance kubeconfig in production.
+func (r *KubernetesReconciler) getClientset() (kubernetes.Interface, error) {
```
mook-as, line 177: This is somewhat odd because `r.Client` already exists.  I guess [`sigs.k8s.io/controller-runtime/pkg/client/fake`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client/fake) doesn't do what you want here? In that case, please leave a note somewhere describing the reason.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -249,14 +318,31 @@ func (r *KubernetesReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (c
 		return ctrl.Result{RequeueAfter: kubeProbeRequeue}, nil
 	}
 
+	// /healthz answers before the node registers, so reporting Ready here
+	// would let `rdd set` return while `kubectl run` still sits Pending.
+	// Gate the first Ready on node Ready. Once Ready, API-server liveness
+	// alone keeps the condition True, so a transient node blip does not
+	// flap Settled.
+	if !apimeta.IsStatusConditionTrue(app.Status.Conditions, appv1alpha1.AppConditionKubernetesReady) {
```
mook-as, line 326: Ah, I had to read the comment quite a few times to understand the comment here; I keep trying to see how it's implement some latching-like behaviour.  I think this can be reworded as:

> /healthz answers before the node registers, so reporting Ready here would let `rdd set` return while `kubectl run` still sits Pending. Gate the first Ready on node Ready. Once Ready, ignore the status of the node, so a transient node blip does not flap Settled.

mook-as (changes requested):

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
jandubois: `r.Client` is for `rdd`, not the `k3s` cluster.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
jandubois: Good catch. It was supposed to check for **any** node, not **all** nodes.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
mook-as: … Err, yeah, now I feel dumb.

mook-as (approved):

PR #436, opened 2026-06-11
Linux packaging: LLM-suggested fixes

mook-as:
This is a few LLM-suggested fixes for Linux packaging.

- appdata.xml: Provide a more proper application ID
- spec: Provide correct qemu dependency in Debian/arm64
- spec: Drop libglib2.0-dev dependency in Debian
- spec: Avoid shipping unneeded `/opt/rancher-desktop/share/`
- spec: Avoid including `chrome-sandbox` twice


jandubois (approved):
Thanks, LGTM (haven't tested it though)

Issue #385, opened 2026-06-03
Create symlinks from the `rdd` directory into the `~/.rd2/bin` directory


jandubois:
This issue is only for macOS and Linux right now.

When `rdd` is installed as part of the Rancher Desktop application, and not just as a CLI tool, then the app controller should make sure all files in the same directory as `rdd` (including itself) are being symlinked into the `shortDir`.

`rdd` will check `strings.HasSuffix(os.Executable(), "/Resources/resources/darwin/bin/rdd")` (or "linux" when running on Linux). If this returns `true`, then it is assumed to be part of the full App and perform the symlinking.

In addition to symlinking all files under identical names, it will also symlink `rdd` to `~/.rd2/bin/kubectl`.

This is subject to change, but to start I think we should always remove the `bin/` directory and recreate all the symlinks on startup, to guarantee they are up-to-date.

This operation only needs to be done once during startup. There is no need to watch the source directory for changes.

mook-as:
Note that as of https://github.com/rancher-sandbox/rancher-desktop-app/pull/64 we no longer have the nested `Resources/resources/…`; it's now just `/Resources/darwin/bin/rdd`.

Unrelated: do we need to think about [`environment.d`](https://manpages.opensuse.org/Tumbleweed/systemd/environment.d.5.en.html#Example) on Linux again to see if that would be useful?  It wouldn't work for non-systemd distributions, but that plus GUI seems… rare these days?

jandubois:
`environment.d` is intriguing, but doesn't work for ssh sessions.

On macOS there is also `/etc/paths.d`, but you must be `root` to write to it.

But note that this issue is strictly about creating the symlinks and not about `PATH` management, which is out-of-scope for the tech preview. On macOS and Linux users will have to edit their shell profile themselves, or use `rdd run docker ...`, which will automatically set the `PATH`, docker and kube context for the command.

jandubois:
> This is subject to change, but to start I think we should always remove the `bin/` directory and recreate all the symlinks on startup, to guarantee they are up-to-date.

I just realized that we should only do this when running from inside the app resources.

Otherwise, if you install both the app and the cli separately, then using the cli version would remove the symlinks to the bundled utilities from the app.

jandubois:
Closed by #438

PR #438, opened 2026-06-12
Link bundled binaries into the instance bin directory

jandubois:
When rdd runs from inside the Rancher Desktop application bundle, publish the bundled binaries into `~/.rd<instance>/bin` as symlinks so a user can put that directory on `PATH`. The new `pkg/binlinks` recreates the directory on each daemon startup and links every file beside the `rdd` executable under the same name, plus a `kubectl` link to `rdd`.

A standalone CLI install makes no change, so it never removes links the application created. Publishing is best-effort: on failure the daemon logs a warning and startup continues.

A BATS test drives a fake app bundle through the daemon to confirm the links are created from the bundle, survive a start from outside it, and update when the bundle changes.

Refs #385

### DIFF
--- a/rdd/bats/tests/20-service/7-binlinks.bats
+++ b/rdd/bats/tests/20-service/7-binlinks.bats
@@ -1,0 +1,72 @@
+load '../../helpers/load'
+
+# Prove the daemon publishes the application's bundled binaries into
+# ~/.rd<instance>/bin only when rdd runs from inside the app bundle. A fake
+# bundle holds a real rdd copy at .../Resources/<os>/bin/rdd next to a couple of
+# stand-in tool files; the daemon links whatever sits beside that rdd.
+
+local_setup_file() {
+    skip_on_windows "Bundled binaries are only published on macOS and Linux."
+    # Clean up from any previous run; svc delete also removes the short dir.
+    rdd svc delete || :
+
+    # A real rdd copy at the bundle path makes the daemon's os.Executable()
+    # match the gate. The siblings are stand-ins; only their names matter.
+    fake_bin="${BATS_FILE_TMPDIR}/Rancher Desktop.app/Contents/Resources/${OS}/bin"
+    mkdir -p "${fake_bin}"
+    cp "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${fake_bin}/rdd${EXE}"
+    echo "stand-in docker" >"${fake_bin}/docker"
+    echo "stand-in helm" >"${fake_bin}/helm"
+
+    FAKE_BIN="${fake_bin}"
+    BIN_DIR="${RDD_SHORT_DIR}/bin"
+    save_var FAKE_BIN BIN_DIR
+}
+
+local_setup() {
+    is_windows && return 0
+    # Each test starts the daemon from a chosen location, so stop any daemon a
+    # previous test left running; an already-running instance rejects the start.
+    rdd svc stop || :
+}
+
+# fake_rdd runs the bundled rdd copy. It closes BATS fds 3 and 4 so the daemon
+# it spawns cannot inherit them and hang bats, matching the rdd() helper.
+fake_rdd() {
+    "${FAKE_BIN}/rdd${EXE}" "$@" 3>&- 4>&-
+}
+
+@test 'publishes bundled binaries when started from the app bundle' {
+    load_var FAKE_BIN BIN_DIR
+    fake_rdd svc start
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/rdd${EXE}"
+    assert_symlink_to "${FAKE_BIN}/docker" "${BIN_DIR}/docker"
+    assert_symlink_to "${FAKE_BIN}/helm" "${BIN_DIR}/helm"
+    # kubectl is not bundled; it links to rdd.
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/kubectl"
+}
+
+@test 'leaves the links untouched when started from outside the bundle' {
+    load_var FAKE_BIN BIN_DIR
+    # The standard rdd is not in a bundle, so its startup must not touch the links.
+    rdd svc start
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/rdd${EXE}"
+    assert_symlink_to "${FAKE_BIN}/docker" "${BIN_DIR}/docker"
+    assert_symlink_to "${FAKE_BIN}/helm" "${BIN_DIR}/helm"
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/kubectl"
+}
+
+@test 'updates the links when the bundle changes and rdd runs from it again' {
+    load_var FAKE_BIN BIN_DIR
+    # Add a tool and drop one, then restart from the bundle.
+    echo "stand-in nerdctl" >"${FAKE_BIN}/nerdctl"
+    rm "${FAKE_BIN}/helm"
+    fake_rdd svc start
+    assert_symlink_to "${FAKE_BIN}/nerdctl" "${BIN_DIR}/nerdctl"
+    # The dropped tool's link is gone, proving the directory was recreated.
+    assert_link_not_exist "${BIN_DIR}/helm"
+    # Unchanged entries are still linked.
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/rdd${EXE}"
+    assert_symlink_to "${FAKE_BIN}/docker" "${BIN_DIR}/docker"
+    assert_symlink_to "${FAKE_BIN}/rdd${EXE}" "${BIN_DIR}/kubectl"
+}
--- a/rdd/bats/tests/20-service/7-binlinks.bats
+++ b/rdd/bats/tests/20-service/7-binlinks.bats
@@ -2,29 +2,31 @@ load '../../helpers/load'
 
 # Prove the daemon publishes the application's bundled binaries into
 # ~/.rd<instance>/bin only when rdd runs from inside the app bundle. A fake
-# bundle holds a real rdd copy at .../Resources/<os>/bin/rdd next to a couple of
-# stand-in tool files; the daemon links whatever sits beside that rdd.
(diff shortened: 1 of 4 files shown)


rdd/pkg/service/cmd/service.go
```diff
@@ -716,6 +717,15 @@ func NewServeCommand(ctx context.Context) *cobra.Command {
 			}
 			cliflag.PrintFlags(fs)
 
+			// Publish the application's bundled binaries to the instance bin
+			// directory. Only macOS and Linux do this. Best-effort: a failure
```
mook-as: Oh, that means I need to revert some installer changes, and make the Windows installer put the bin directories in `$PATH` again.  But probably not the WSL ones, given that other bug in rancher-desktop?

rdd/bats/tests/20-service/7-binlinks.bats
```diff
@@ -0,0 +1,72 @@
...
+# bundle holds a real rdd copy at .../Resources/<os>/bin/rdd next to a couple of
+# stand-in tool files; the daemon links whatever sits beside that rdd.
+
+local_setup_file() {
+    skip_on_windows "Bundled binaries are only published on macOS and Linux."
+    # Clean up from any previous run; svc delete also removes the short dir.
+    rdd svc delete || :
+
+    # A real rdd copy at the bundle path makes the daemon's os.Executable()
+    # match the gate. The siblings are stand-ins; only their names matter.
```
mook-as: Which gate is this? (There's no context here.)

I think a better explanation is something along the lines of:
> Mock up a packaged Rancher Desktop distribution so `rdd` thinks it's bundled.

rdd/bats/tests/20-service/7-binlinks.bats
```diff
@@ -0,0 +1,72 @@
...
+    echo "stand-in docker" >"${fake_bin}/docker"
+    echo "stand-in helm" >"${fake_bin}/helm"
+
+    FAKE_BIN="${fake_bin}"
+    BIN_DIR="${RDD_SHORT_DIR}/bin"
+    save_var FAKE_BIN BIN_DIR
+}
+
+local_setup() {
+    is_windows && return 0
```
mook-as: Does the `skip_on_windows` in `local_setup_file` not do it? Huh.

rdd/bats/tests/20-service/7-binlinks.bats
```diff
@@ -0,0 +1,72 @@
...
+# stand-in tool files; the daemon links whatever sits beside that rdd.
+
+local_setup_file() {
+    skip_on_windows "Bundled binaries are only published on macOS and Linux."
+    # Clean up from any previous run; svc delete also removes the short dir.
+    rdd svc delete || :
+
+    # A real rdd copy at the bundle path makes the daemon's os.Executable()
+    # match the gate. The siblings are stand-ins; only their names matter.
+    fake_bin="${BATS_FILE_TMPDIR}/Rancher Desktop.app/Contents/Resources/${OS}/bin"
```
mook-as: Is that the correct path on Linux?

rdd/pkg/service/cmd/service.go
```diff
@@ -716,6 +717,15 @@ func NewServeCommand(ctx context.Context) *cobra.Command {
...
+			// must not block the control plane.
+			if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
```
mook-as:
```suggestion
			if runtime.GOOS != "windows" {
```
In the theoretical case that we run on FreeBSD…

rdd/pkg/binlinks/binlinks.go
```diff
@@ -0,0 +1,80 @@
...
+	return linkBinaries(execPath, filepath.Join(instance.ShortDir(), "bin"))
+}
+
+// inAppBundle reports whether execPath is the bundled rdd binary for the given
+// OS, as opposed to a standalone CLI install. The application stages its
+// per-platform resources under Resources/<goos>/bin/rdd. The leading separator
+// anchors the match, so an unrelated path ending in "Resources/<goos>/bin/rdd"
+// does not qualify.
+func inAppBundle(execPath, goos string) bool {
+	return strings.HasSuffix(execPath, "/Resources/"+goos+"/bin/rdd")
```
mook-as: Is that correct for Linux? That has lower case `r` in `/opt/rancher-desktop/resources/linux/bin/rdd`, IIRC.

rdd/pkg/binlinks/binlinks.go
```diff
@@ -0,0 +1,80 @@
...
+		name := entry.Name()
+		if err := os.Symlink(filepath.Join(srcDir, name), filepath.Join(binDir, name)); err != nil {
+			return fmt.Errorf("link %q: %w", name, err)
+		}
+	}
+
+	// No separate kubectl binary is bundled; rdd provides it. Link kubectl to
+	// rdd so kubectl on PATH reaches rdd.
+	if err := os.Symlink(execPath, filepath.Join(binDir, "kubectl")); err != nil {
+		return fmt.Errorf("link kubectl: %w", err)
```
mook-as: `kubectl` is not quoted. Not really an issue, just amusing.

rdd/pkg/binlinks/binlinks_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}{
+		{
+			name:     "macOS app bundle",
+			execPath: "/Applications/Rancher Desktop.app/Contents/Resources/darwin/bin/rdd",
+			goos:     "darwin",
+			want:     true,
+		},
+		{
+			name:     "Linux app bundle",
+			execPath: "/opt/rancher-desktop/Resources/linux/bin/rdd",
```
mook-as: My installed package has:
> `/opt/rancher-desktop/resources/linux/bin/rdd`
(From my [temporary branch](https://download.opensuse.org/repositories/home:/mook_work:/branches:/isv:/Rancher:/dev/rpm/home:mook_work:branches:isv:Rancher:dev))

Note that for parallel installation, we may need to change this to `/opt/rancher-desktop-2`? Not sure.

rdd/bats/tests/20-service/7-binlinks.bats
```diff
@@ -0,0 +1,72 @@
...
+    mkdir -p "${fake_bin}"
+    cp "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${fake_bin}/rdd${EXE}"
+    echo "stand-in docker" >"${fake_bin}/docker"
+    echo "stand-in helm" >"${fake_bin}/helm"
+
+    FAKE_BIN="${fake_bin}"
+    BIN_DIR="${RDD_SHORT_DIR}/bin"
```
mook-as: I would recommend calling this `DEST_DIR`, `OUT_DIR`, or something else, there are to many _bin_ directories flying around.

rdd/bats/tests/20-service/7-binlinks.bats
```diff
@@ -0,0 +1,72 @@
...
+local_setup_file() {
+    skip_on_windows "Bundled binaries are only published on macOS and Linux."
+    # Clean up from any previous run; svc delete also removes the short dir.
+    rdd svc delete || :
+
+    # A real rdd copy at the bundle path makes the daemon's os.Executable()
+    # match the gate. The siblings are stand-ins; only their names matter.
+    fake_bin="${BATS_FILE_TMPDIR}/Rancher Desktop.app/Contents/Resources/${OS}/bin"
+    mkdir -p "${fake_bin}"
+    cp "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${fake_bin}/rdd${EXE}"
```
mook-as, line 21: Given that we skip this test on Windows, do we need `${EXE}` at all? I'm totally fine with keeping it.  But we don't put it on `docker` etc. for some reason.

mook-as (dismissed):
The Linux path is definitely wrong, at least.

rdd/pkg/service/cmd/service.go
jandubois: Yes, we postponed the symlinks on Windows because they require developer mode, or otherwise some kind of shim. We still have to figure out exactly how we are going to do it.

And for now we don't do WSL integration, so Linux utilities don't need to be shipped yet.

jandubois:
I just realized that the app should still add the `rdd` and `kubectl` symlinks to itself, even when not running out of the app resources directory. Or at least if the symlinks either don't exist, or are dangling.

jandubois:
I added a partial implementation of #448 to this PR to create symlinks (or hardlinks) on Windows too, but not supporting cross-device links. Without it we would not have a bundled `kubectl` in the tech preview.

Also I want this to work on all platforms (for the tech preview):

```console
❯ wget … rdd
❯ chmod +x ./rdd
❯ ./rdd run kubectl get node
NAME      STATUS   ROLES           AGE   VERSION
lima-rd   Ready    control-plane   2s    v1.34.6+k3s1
```

rdd/pkg/binlinks/binlinks.go
```diff
@@ -0,0 +1,198 @@
...
+
+// link points linkPath at target, preferring a symlink for its self-documenting
+// target and falling back to a hardlink where symlinks need absent privileges
+// (Windows without developer mode). useSymlink false skips straight to a
+// hardlink. A hardlink stays current across app updates because rdd recreates
+// these links on every start; it works only within one volume, so the
+// cross-volume copy fallback is deferred (see #448).
+func link(target, linkPath string, useSymlink bool) error {
+	if useSymlink {
+		if err := os.Symlink(target, linkPath); err == nil {
```
Nino-K: I think we should at least log the symlink error or, even better, fallback only on the specific privilege related errors.

rdd/pkg/binlinks/binlinks.go
```diff
@@ -0,0 +1,198 @@
...
+	entries, err := os.ReadDir(binDir)
+	if err != nil {
+		return fmt.Errorf("read %q: %w", binDir, err)
+	}
+	for _, entry := range entries {
+		if entry.Type()&os.ModeSymlink == 0 {
+			continue
+		}
+		linkPath := filepath.Join(binDir, entry.Name())
+		if _, err := os.Stat(linkPath); err == nil {
```
Nino-K, line 174: If this returns any error other than `ENOENT` (e.g `EACCES` or `ELOOP`), it will return an erorr and stop, leaving any other dangling links unpruned. I can see the function is best-effort cleanup, but would it make sense for non `EnOENT` errors to log and continue rather than abort?

Nino-K (approved):

PR #440, opened 2026-06-12
Turn rdd into a multi-call binary

### LINKED ISSUE #439: Turn `rdd` into a multi-call binary
`rdd` already embeds 2 subcommands: `rdd kubectl` and `rdd yq`.

When `rdd` is invoked via a symlink (or has been renamed), then the `os.Executable()` name should be used to determine the subcommand to be invoked:

* Use the basename of the executable name
* Remove the filename extension (strips `.exe` extension on Windows)
* If the name is `kubectl` or `yq`, invoke the corresponding subcommands (adjust os.Args as necessary)
* Otherwise run `rdd` normally

jandubois:
Dispatch to the embedded `kubectl` and `yq` subcommands when `rdd` is invoked under their name via a symlink or rename. Use `argv[0]`'s basename, not `os.Executable()`, which resolves symlinks back to `rdd` on Linux.

Implements #439.

### DIFF
--- a/rdd/bats/tests/10-cli/7-multicall.bats
+++ b/rdd/bats/tests/10-cli/7-multicall.bats
@@ -1,0 +1,41 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-FileCopyrightText: The Rancher Desktop Authors
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+
+# The symlink mechanism is adapted from
+# https://github.com/lima-vm/lima/blob/master/hack/bats/tests/yq.bats
+
+load '../../helpers/load'
+
+# link_rdd_as symlinks the rdd binary under <name> in the per-test temp dir, so
+# invoking "${BATS_TEST_TMPDIR}/<name>" runs rdd as a multi-call binary.
+link_rdd_as() { # <name>
+    ln -sf "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${BATS_TEST_TMPDIR}/$1"
+}
+
+@test 'a kubectl symlink runs the embedded kubectl' {
+    link_rdd_as "kubectl${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/kubectl${EXE}" version --client
+    assert_line --partial "Client Version"
+}
+
+@test 'a yq symlink runs the embedded yq' {
+    link_rdd_as "yq${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/yq${EXE}" --version
+    assert_output --regexp '^yq .*mikefarah.* version v'
+}
+
+@test 'multi-call detection strips all extensions' {
+    # A name like yq.rdd or yq.rdd.exe still dispatches to yq.
+    link_rdd_as "yq.rdd${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/yq.rdd${EXE}" -n .foo=42
+    assert_output 'foo: 42'
+}
+
+@test 'an unrecognized name runs rdd normally' {
+    link_rdd_as "notacommand${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/notacommand${EXE}" version
+    # Matches a git version tag or a commit hash, like 1-version.bats.
+    assert_output --regexp '^(v[0-9]+\.[0-9]+\.[0-9]+|[a-f0-9]{7,})'
+}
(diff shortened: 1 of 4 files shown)


rdd/cmd/rdd/multicall.go
```diff
@@ -0,0 +1,28 @@
...
+	"strings"
+)
+
+// multiCallArgs lets rdd run as an embedded command (kubectl, yq) when invoked
+// under that name via a symlink or rename, inserting the matching subcommand
+// for Cobra to dispatch. The name is argv[0]'s basename minus extensions;
+// os.Executable() would resolve symlinks back to rdd on Linux.
+func multiCallArgs(args []string) []string {
+	name := filepath.Base(args[0])
+	name, _, _ = strings.Cut(name, ".")
```
mook-as, line 22: That cuts on the first dot. I guess it doesn't matter here, though; we only match against strings without dots. So calling the executable `yq.rdd` would still invoke yq (which is probably fine for now).

Oh, I see that you actually for this in BATS, so it's totally intended.

rdd/cmd/rdd/multicall_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"gotest.tools/v3/assert"
+)
+
+func TestMultiCallArgs(t *testing.T) {
+	cases := []struct {
+		name string
+		args []string
+		want []string
+	}{
+		{"rdd runs normally", []string{"rdd", "start"}, []string{"rdd", "start"}},
```
mook-as, line 19: Needs a test that shows the normal way of invoking works: `[]string{"rdd", "kubectl", "cluster-info"}`.  I think that'll pass, of course.

rdd/bats/tests/10-cli/7-multicall.bats
```diff
@@ -0,0 +1,41 @@
...
+
+# The symlink mechanism is adapted from
+# https://github.com/lima-vm/lima/blob/master/hack/bats/tests/yq.bats
+
+load '../../helpers/load'
+
+# link_rdd_as symlinks the rdd binary under <name> in the per-test temp dir, so
+# invoking "${BATS_TEST_TMPDIR}/<name>" runs rdd as a multi-call binary.
+link_rdd_as() { # <name>
+    ln -sf "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${BATS_TEST_TMPDIR}/$1"
```
mook-as, line 14: Huh, I wonder how this works on msys. But the tests pass, so it's probably fine.

rdd/bats/tests/10-cli/7-multicall.bats
```diff
@@ -0,0 +1,41 @@
...
+    link_rdd_as "yq.rdd${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/yq.rdd${EXE}" -n .foo=42
+    assert_output 'foo: 42'
+}
+
+@test 'an unrecognized name runs rdd normally' {
+    link_rdd_as "notacommand${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/notacommand${EXE}" version
+    # Matches a git version tag or a commit hash, like 1-version.bats.
+    assert_output --regexp '^(v[0-9]+\.[0-9]+\.[0-9]+|[a-f0-9]{7,})'
```
mook-as: Maybe we should run `rdd --help` and match for `RDD manages the Rancher Desktop 2 background services.` instead; that makes it clearer it's the correct executable (mode).

rdd/bats/tests/10-cli/7-multicall.bats
```diff
@@ -0,0 +1,41 @@
...
+# link_rdd_as symlinks the rdd binary under <name> in the per-test temp dir, so
+# invoking "${BATS_TEST_TMPDIR}/<name>" runs rdd as a multi-call binary.
+link_rdd_as() { # <name>
+    ln -sf "${PATH_REPO_ROOT}/bin/rdd${EXE}" "${BATS_TEST_TMPDIR}/$1"
+}
+
+@test 'a kubectl symlink runs the embedded kubectl' {
+    link_rdd_as "kubectl${EXE}"
+    run -0 "${BATS_TEST_TMPDIR}/kubectl${EXE}" version --client
+    assert_line --partial "Client Version"
```
mook-as: That's… probably fine, but it's a bit vague when reading through the test. Matching `kubectl --help` for `kubectl controls the Kubernetes cluster manager.` might be clearer.

mook-as (dismissed):
Basically fine, but I think it would be clearer if we match against the command help's short description instead of the (sometimes ambiguous) version string.

mook-as (approved):

PR #443, opened 2026-06-12
Allow RDD_VM_CPUS to override the VM cpu count

jandubois:
The Lima template ships `cpus: 2` for everyone, but the 4-core CI runners starve the guest: telemetry shows both vCPUs pegged during the k3s bootstrap while container creates queue for minutes. Let CI give its VMs more cores without changing the default users get: 4 on the macOS runners (where the starvation breaks tests), 3 on the Linux runners (their host side idles at the default), and nothing on Windows (Lima's WSL2 driver does not apply `cpus`). A set-but-invalid value is an error, not a fallback, so a typo cannot quietly change what CI measures. An App spec property will replace the variable once VM sizing becomes a real setting.

Validation on the macos-15-intel pool, normalized for runner quality by a single-thread benchmark probe:

- 3 vCPUs: 7/10 NodePort runs passed; both slowness failures hit the slowest runners; pod-Running took 1m44s–3m11s on mid-grade hosts; host idle never reached 0%.
- 4 vCPUs: 4/4 passed, including the slowest runner measured; pod-Running took 55–62s on hosts matched to the 3-vCPU runs (2–3× faster). Host idle hit 0% in a quarter of the samples, but the runner agent stayed scheduled and no run lost its connection in the ~40-run campaign. Drop macOS to 3 if disconnects appear.

Note for sequencing: this branch and #442 both touch `applySpecToTemplate` and `environment.md`; whichever merges second has a trivial conflict to resolve.

### DIFF
--- a/.github/workflows/rdd-bats.yaml
+++ b/.github/workflows/rdd-bats.yaml
@@ -206,6 +206,28 @@ jobs:
         output-dir: rdd/rdd-logs/_diag
       continue-on-error: true
 
+    - name: "Linux: size the app VM for the runner"
+      if: needs.check-paths.outputs.should-run && runner.os == 'Linux'
+      shell: bash
+      run: |
+        # The ubuntu runners also have 4 cores, and the host side idles at
+        # the 2-vCPU default; give the QEMU guest 3 and keep one core for
+        # the runner agent.
+        echo "RDD_VM_CPUS=3" >> "$GITHUB_ENV"
+
+    - name: "macOS: size the app VM for the runner"
+      if: needs.check-paths.outputs.should-run && runner.os == 'macOS'
+      shell: bash
+      run: |
+        # The 2-vCPU default starves the guest on the 4-core macos-15-intel
+        # runners: both vCPUs saturate during the k3s bootstrap and container
+        # creates take minutes. Give the VM all 4 cores: validation showed
+        # 4 vCPUs reaching pod-Running 2-3x faster than 3 and passing even
+        # the slowest runners, while the runner agent stayed scheduled
+        # through 0%-idle stretches. Drop to 3 if runs start losing their
+        # runner connection.
+        echo "RDD_VM_CPUS=4" >> "$GITHUB_ENV"
+
     - name: Configure BATS targets
       if: needs.check-paths.outputs.should-run
       shell: bash
--- a/rdd/docs/design/environment.md
+++ b/rdd/docs/design/environment.md
@@ -13,6 +13,7 @@ These variables control RDD behavior. Set them before running `rdd` commands.
 | `RDD_KEEP_LOGS` | Preserves logs for post-mortem debugging. See [Log Preservation](#log-preservation) for details. | unset |
 | `RDD_LOG_LEVEL` | Sets the log level (`fatal`, `error`, `warn`, `info`, `debug`, `trace`). Overridden by `--log-level` flag. When unset, defaults to `debug` in developer mode, `warn` otherwise. | unset |
 | `RDD_LOG_TITLE` | When set, writes this string as the first line of each new log file. Useful for identifying log files from specific test runs or sessions. | unset |
+| `RDD_VM_CPUS` | Overrides the app VM's CPU count (the Lima template default is 2). Intended for CI, where runner sizing differs from user machines. A set-but-invalid value is an error, not a fallback. To be replaced by an App spec property. | unset |
 
 ### BATS Test Variables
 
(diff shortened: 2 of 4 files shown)


.github/workflows/rdd-bats.yaml
```diff
@@ -206,6 +206,28 @@ jobs:
         output-dir: rdd/rdd-logs/_diag
       continue-on-error: true
 
+    - name: "Linux: size the app VM for the runner"
+      if: needs.check-paths.outputs.should-run && runner.os == 'Linux'
+      shell: bash
```
mook-as, line 211: `bash` is the default on Linux.

.github/workflows/rdd-bats.yaml
```diff
@@ -206,6 +206,28 @@ jobs:
...
+      run: |
+        # The ubuntu runners also have 4 cores, and the host side idles at
+        # the 2-vCPU default; give the QEMU guest 3 and keep one core for
+        # the runner agent.
+        echo "RDD_VM_CPUS=3" >> "$GITHUB_ENV"
+
+    - name: "macOS: size the app VM for the runner"
+      if: needs.check-paths.outputs.should-run && runner.os == 'macOS'
+      shell: bash
```
mook-as, line 220: `bash` (the old GPL2 version, I believe) is the default on macOS.

.github/workflows/rdd-bats.yaml
```diff
@@ -206,6 +206,28 @@ jobs:
         output-dir: rdd/rdd-logs/_diag
       continue-on-error: true
 
+    - name: "Linux: size the app VM for the runner"
+      if: needs.check-paths.outputs.should-run && runner.os == 'Linux'
+      shell: bash
+      run: |
+        # The ubuntu runners also have 4 cores, and the host side idles at
```
mook-as, line 213: _also_ seems inappropriate here: the first case is below this line (?).  Just dropping the word (if you happen to amend) is fine.

.github/workflows/rdd-bats.yaml
```diff
@@ -206,6 +206,28 @@ jobs:
...
+        # the 2-vCPU default; give the QEMU guest 3 and keep one core for
+        # the runner agent.
+        echo "RDD_VM_CPUS=3" >> "$GITHUB_ENV"
```
mook-as, line 216: To make it easier for future runners:

```suggestion
        echo "RDD_VM_CPUS=$(($(nproc) - 1))" >> "$GITHUB_ENV"
```

.github/workflows/rdd-bats.yaml
```diff
@@ -206,6 +206,28 @@ jobs:
...
+      shell: bash
+      run: |
+        # The 2-vCPU default starves the guest on the 4-core macos-15-intel
+        # runners: both vCPUs saturate during the k3s bootstrap and container
+        # creates take minutes. Give the VM all 4 cores: validation showed
+        # 4 vCPUs reaching pod-Running 2-3x faster than 3 and passing even
+        # the slowest runners, while the runner agent stayed scheduled
+        # through 0%-idle stretches. Drop to 3 if runs start losing their
+        # runner connection.
+        echo "RDD_VM_CPUS=4" >> "$GITHUB_ENV"
```
mook-as, line 229: I guess `$(nproc)` would be fine here, but I guess 4 is okay too?

mook-as (approved):
Looks fine, some nits (fixing is optional).

Do we have an issue filed to replace this?

.github/workflows/rdd-bats.yaml
jandubois: I think we always specify the shell explicitly; doesn't hurt.

.github/workflows/rdd-bats.yaml
jandubois: Agree, but I don't want to rerun CI for a trivial change. We are going to touch all this again once we have a `cpus` property on the App object.

.github/workflows/rdd-bats.yaml
jandubois: I would take this if I made another commit, but not really important right now.

.github/workflows/rdd-bats.yaml
jandubois: macOS doesn't have `nproc`, but maybe `sysctl -n hw.ncpu`?

jandubois (commented):
> Looks fine, some nits (fixing is optional).

I think I'll merge as-is because we have a whole pile of PRs waiting.

> Do we have an issue filed to replace this?

Not sure what you mean by "this", but the answer is likely "no". Are you asking about remaining work (making `cpus` an `App` property)?

PR #444, opened 2026-06-12
Allow overriding log directory

mook-as:
This adds the ability to override the log directory (by setting `RDD_LOG_DIR`), analogous to the feature in Rancher Desktop (1).  This is meant to be used for tests to be easier to find logs.

This also adds unit tests for `rdd/pkg/instance`, implemented by spawning subprocesses to reset the various `sync.Once…` things.


jandubois (approved):
Thanks, LGTM

`rdd svc delete` will delete whatever `RDD_LOG_DIR` points to, but I guess that is fine.

PR #447, opened 2026-06-13
Add distro-overlay tool to layer RD assets onto the distro

### LINKED ISSUE #450: Create distro-overlay tool to customize the opensuse distro (tarball and raw image)
Create a `distro-overlay` tool that will take an OS-only tarball or raw image and add additional files, directories, and symlinks based on a manifest (to specify uid/gid/mode/mtime).

The result needs to be a "reproducible build" because we will need to embed the checksum in the `lima.yaml` template. So when we rebuild the final release bits the digest must not change.

jandubois:
distro-overlay merges Rancher Desktop assets into the pristine openSUSE distro at build time -- into both the WSL tarball and the Lima ext4 image -- from a YAML manifest carrying each entry's destination, ownership, and mode.

See `cmd/distro-overlay/manifest.example.yaml` for the schema and defaults.

### DIFF
--- a/rdd/cmd/distro-overlay/main.go
+++ b/rdd/cmd/distro-overlay/main.go
@@ -1,0 +1,138 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Command distro-overlay merges Rancher Desktop assets into a pristine openSUSE
+// distro image or tarball, taking destination paths and ownership from a manifest.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/overlay"
+)
+
+func main() {
+	manifest := flag.String("manifest", "", "overlay manifest (YAML)")
+	source := flag.String("source", "", "directory holding file sources (default: manifest directory)")
+	format := flag.String("format", "auto", "distro format: auto, raw, or tar")
+	output := flag.String("output", "", "output path for tar format (default: overwrite input)")
+	mtimeArg := flag.String("mtime", "", "modification time for every entry, as Unix epoch seconds or RFC3339 (default: now)")
+	flag.Parse()
+
+	if *manifest == "" || flag.NArg() != 1 {
+		fmt.Fprintln(os.Stderr, "usage: distro-overlay --manifest M [--source D] [--format auto|raw|tar] [--output O] [--mtime T] <distro>")
+		os.Exit(2)
+	}
+	mtime, err := parseMtime(*mtimeArg)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "distro-overlay:", err)
+		os.Exit(2)
+	}
+	if err := run(*manifest, *source, *format, *output, flag.Arg(0), mtime); err != nil {
+		fmt.Fprintln(os.Stderr, "distro-overlay:", err)
+		os.Exit(1)
+	}
+}
+
+// parseMtime reads the --mtime flag as Unix epoch seconds or an RFC3339
+// timestamp, defaulting to the current time when the flag is empty. A build
+// passes a deterministic value, such as a git commit time, for a reproducible
+// result.
+func parseMtime(arg string) (time.Time, error) {
+	if arg == "" {
+		return time.Now(), nil
+	}
+	if secs, err := strconv.ParseInt(arg, 10, 64); err == nil {
+		return time.Unix(secs, 0).UTC(), nil
+	}
+	if t, err := time.Parse(time.RFC3339, arg); err == nil {
+		return t, nil
+	}
+	return time.Time{}, fmt.Errorf("invalid --mtime %q: want Unix epoch seconds or RFC3339", arg)
+}
+
+func run(manifestPath, sourceDir, format, output, distro string, mtime time.Time) error {
+	m, err := overlay.LoadManifest(manifestPath)
+	if err != nil {
+		return err
+	}
+	if sourceDir == "" {
+		sourceDir = filepath.Dir(manifestPath)
+	}
+	if format == "auto" {
+		if format, err = detectFormat(distro); err != nil {
+			return err
+		}
+	}
+	switch format {
+	case "raw":
+		d, err := overlay.OpenImage(distro)
+		if err != nil {
+			return err
+		}
+		return overlay.Apply(d, m, sourceDir, mtime)
+	case "tar":
+		return applyTarFile(distro, output, m, sourceDir, mtime)
+	default:
+		return fmt.Errorf("unknown format %q", format)
+	}
+}
+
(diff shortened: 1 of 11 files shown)


jandubois:
We need a way to specify the mtime of symlinks and directories so that we end up with an immutable release with a stable digest.

One way would be to use the mtime of the manifest, but that only works if the manifest is static and checked into git.

rdd/cmd/distro-overlay/manifest.example.yaml
```diff
@@ -0,0 +1,64 @@
...
+    gid: 472
+
+  # --- A directory -----------------------------------------------------------
+  # You only need an explicit `dir` entry to give a directory non-default
+  # ownership or permissions, or to create an empty directory. Parent
+  # directories of files and symlinks are created automatically as root:root
+  # 0755, so most directories need no entry at all. Explicit `dir` entries are
+  # applied before files, so their permissions are not overwritten by the
+  # implicit parent creation that a file underneath them would trigger.
+  - path: /var/lib/rancher-desktop
```
mook-as, line 31: Being able to copy a whole directory in would make our lives a lot easier (think `/usr/local/lib/systemd/…`)

rdd/cmd/distro-overlay/manifest.example.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  # --- A symlink -------------------------------------------------------------
+  - path: /usr/local/bin/rdctl
+    type: symlink
+    # target is stored verbatim; it may be absolute or relative.
+    target: /usr/local/bin/rancher-desktop-guest-agent
+
+# Notes:
+#   * Overwrites win. If `path` already exists in the distro, the overlay entry
+#     replaces it; the original is dropped, not duplicated.
```
mook-as: The bit after the semicolon is excessive verbiage (probably too much AI)

mook-as (commented):
(I haven't looked at the code at all, just the docs-ish things)

Is there any reason the manifest _wouldn't_ be static? We could have a checked-in manifest per platform if desired, but I feel like dynamically generating it is just confusing.

Note that we still get lima to download the opensuse image at runtime, so I'm not sure how the output image is supposed to be used.

rdd/cmd/distro-overlay/manifest.example.yaml
jandubois: Good idea. I've implemented it: you just have to add a `source` field to a `dir` entry. Updates are ordered, so individual files, dirs, symlinks are added first (so you can override uid/gid/mode).

jandubois:
> Is there any reason the manifest _wouldn't_ be static? We could have a checked-in manifest per platform if desired, but I feel like dynamically generating it is just confusing.

I realized that `git` doesn't actually preserve `mtime`. All the times on disk are the checkout times and basically random. So it also doesn't matter if the manifest is checked in or not.

Instead I've added an `--mtime` option to specify the `mtime` for all entries added by the overlay (defaults to current time). We can maybe set it to the `mtime` of either the manifest, or `dependencies.yaml`, whichever is later.

Works for tarballs, but doesn't give us reproducible images because `go-diskfs` will update the inode change times to the current time. Needs an upstream change to allow us setting it explicitly. Not really important right now.

rdd/pkg/overlay/overlay_test.go
```diff
@@ -0,0 +1,919 @@
...
+}
+
+// e2fsprogs runs an e2fsprogs command and returns its exit code and output. It
+// skips the test where e2fsprogs is missing, unless RDD_REQUIRE_E2FSPROGS is
+// true, which CI sets where it provides e2fsprogs.
+//
+// cmd/distro-overlay keeps its own copy. Two of them beat a test-helper package
+// that nothing else would use, so long as the RDD_REQUIRE_E2FSPROGS contract
+// stays the same in both; a third caller is when to extract it.
+func e2fsprogs(t *testing.T, name string, args ...string) (code int, output string) {
```
Nino-K, line 773: Something I noticed on the default local run with `e2fsprogs` missing and `RDD_REQUIRE_E2FSPROGS` unset, every image test is skipped, but `make test` still prints ok, so the ext4 backend looks like it's covered when it isn't. It will be ok on the CI's since it sets the flag, I think, could we just make the local skip mroe clear/noisier (a t.Log, or a README line to install e2fsprogs)?

rdd/pkg/overlay/manifest.go
```diff
@@ -0,0 +1,160 @@
...
+// mode parses the octal Mode, returning def when it is unset.
+func (e *Entry) mode(def os.FileMode) (os.FileMode, error) {
+	if e.Mode == "" {
+		return def, nil
+	}
+	v, err := strconv.ParseUint(e.Mode, 8, 32)
+	if err != nil {
+		return 0, err
+	}
+	if v > 0o777 {
```
Nino-K, line 156: This appears to reject anything above `0o777`. Just wanted to confirm that no shipped asset needs a `setgid` dir or `setuid` helper.

rdd/pkg/overlay/tar.go
```diff
@@ -0,0 +1,216 @@
...
+	_, err = t.tw.Write(buf)
+	return err
+}
+
+func (t *tarDistro) Symlink(link, target string, mtime time.Time) error {
+	if err := t.checkType(link, tar.TypeSymlink); err != nil {
+		return err
+	}
+	t.written[tarClean(link)] = tar.TypeSymlink
+	// The header leaves the link root-owned, while the image backend gives it
```
Nino-K, line 175: The two backends give a symlink different owners, tar writes it `root:root`, the image leaves it owned by its parent dir. Only shows up for a link inside a non root owned, so I'm not asking you to fix it. Just confirming it's on purpose that "symlink owner is don't-care" call and not something that'll surprise us later.

rdd/cmd/distro-overlay/main.go
```diff
@@ -0,0 +1,259 @@
...
+		if err != nil {
+			return emptyOutput(output, err)
+		}
+		target = output
+	}
+	d, err := overlay.OpenImage(target)
+	if err == nil {
+		err = overlay.Apply(d, m, sourceDir, mtime)
+	}
+	if err != nil && output != "" {
```
Nino-K, line 141: Without `--output`, this edits the real image in place and only fails after `go-diskfs` has already written to it. So if a run fails partway, the distro is left corrupted with no way to undo it?

rdd/pkg/overlay/overlay_test.go
jandubois: I added it to the README. A message will not show if the test is otherwise successful.

rdd/pkg/overlay/tar.go
jandubois: Yes, it is a limitation of go-diskfs.

rdd/cmd/distro-overlay/main.go
jandubois: Correct. I've added a `--in-place` option that must be specified if you don't specify `--output`. So it is obvious that you intend to modify the same file.

If you are not sure your command is going to work, then you should not use `--in-place`. In case of error it will truncate the file.

None of this will happen in our CI or when invoked from our `Makefile`.

Nino-K (approved):

PR #454, opened 2026-06-16
Repair Windows console after wsl.exe breaks newline auto-return

jandubois:
`wsl.exe` (Lima's WSL2 driver) boots the VM in the console it shares with rdd and leaves it in `DISABLE_NEWLINE_AUTO_RETURN` mode, so our later output "staircases" down the terminal.

Detaching the hostagent isn't an option: graceful shutdown sends `CTRL_BREAK`, which only reaches a process sharing our console. So we clear the bit ourselves, before each log write and before running a child command.

The tempting fix - detach the hostagent from the console `DETACHED_PROCESS`/`CREATE_NO_WINDOW`) - would break graceful shutdown. The hostagent is stopped with `GenerateConsoleCtrlEvent(CTRL_BREAK)` (`process.Interrupt`), and that signal only reaches a process that shares the caller's console.

### DIFF
--- a/rdd/pkg/cli/console/console_windows.go
+++ b/rdd/pkg/cli/console/console_windows.go
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Package console repairs the controlling console after a child process leaves
+// it in a broken output mode.
+package console
+
+import (
+	"io"
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+// disableNewlineAutoReturn (DISABLE_NEWLINE_AUTO_RETURN) suppresses the implicit
+// carriage return on a line feed. wsl.exe — Lima's WSL2 driver — sets it, with
+// virtual-terminal processing, on the console it shares with us and never
+// restores it, so every later '\n' moves down without returning to column 0:
+// the "staircase".
+const disableNewlineAutoReturn = 0x0008
+
+// Repair clears DISABLE_NEWLINE_AUTO_RETURN on the controlling console, undoing
+// the staircase a child leaves behind while preserving virtual-terminal
+// processing (ANSI colors). Best-effort: a no-op when the console is redirected.
+func Repair() {
+	for _, f := range []*os.File{os.Stdout, os.Stderr} {
+		h := windows.Handle(f.Fd())
+		var mode uint32
+		if windows.GetConsoleMode(h, &mode) != nil {
+			continue // not a console handle (redirected)
+		}
+		if mode&disableNewlineAutoReturn != 0 {
+			_ = windows.SetConsoleMode(h, mode&^disableNewlineAutoReturn)
+		}
+		return // stdout and stderr share one console buffer; one repair suffices
+	}
+}
+
+// RepairingWriter wraps w so each Write first repairs the console mode, keeping
+// logged lines from staircasing after a child corrupts the console.
+func RepairingWriter(w io.Writer) io.Writer { return repairingWriter{w} }
+
+type repairingWriter struct{ w io.Writer }
+
+func (rw repairingWriter) Write(p []byte) (int, error) {
+	Repair()
+	return rw.w.Write(p)
+}
--- a/rdd/cmd/rdd/main.go
+++ b/rdd/cmd/rdd/main.go
@@ -23,6 +23,7 @@ import (
 	_ "k8s.io/component-base/metrics/prometheus/clientgo"
 	_ "k8s.io/component-base/metrics/prometheus/version"
 
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/cli/console"
 	cliexit "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/cli/exit"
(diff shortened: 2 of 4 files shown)


jandubois:
Without this PR I get "stairways" output on Windows:

```console
PS C:\Users\jan> rdd run kubectl get node
time="2026-06-15T16:07:30-07:00" level=info msg="starting \"rancher-desktop-2\" control plane"
time="2026-06-15T16:07:31-07:00" level=info msg="waiting for API server"
time="2026-06-15T16:07:32-07:00" level=info msg="waiting for CRDs to be established"
time="2026-06-15T16:07:40-07:00" level=info msg="App created"
time="2026-06-15T16:07:40-07:00" level=info msg="Waiting for App to settle"
time="2026-06-15T16:07:40-07:00" level=info msg="KubernetesReady=False: VM is not running (NotRunning)"
time="2026-06-15T16:07:40-07:00" level=info msg="ContainerEngineReady=False: Container engine stopped (Stopped)"
time="2026-06-15T16:07:40-07:00" level=info msg="Created=Unknown: Reconciliation in progress (Pending)"
time="2026-06-15T16:07:40-07:00" level=info msg="Settled=False: Waiting for LimaVM to report its state (WaitingForLimaVM)"
time="2026-06-15T16:07:40-07:00" level=info msg="Settled=False: Applying the configuration change to the VM (ApplyingTemplate)"
time="2026-06-15T16:08:20-07:00" level=info msg="Created=True: Lima instance created successfully (Created)"
time="2026-06-15T16:08:20-07:00" level=info msg="Settled=False: Waiting for LimaVM to report its state (WaitingForLimaVM)"
time="2026-06-15T16:08:20-07:00" level=info msg="Settled=False: LimaVM has not yet reached Started (Reconciling)"
time="2026-06-15T16:08:20-07:00" level=info msg="Running=Unknown: Verifying instance state after controller restart (Reconciling)"
time="2026-06-15T16:08:20-07:00" level=info msg="Settled=False: LimaVM has not yet reached Started (Starting)"
time="2026-06-15T16:08:20-07:00" level=info msg="Running=False: Lima instance is starting (Starting)"
time="2026-06-15T16:09:11-07:00" level=info msg="Settled=False: Container engine stopped (Stopped)"
                                                                                                   time="2026-06-15T16:09:11-07:00" level=info msg="Running=True: Lima instance is running (Started)"
                                                                             time="2026-06-15T16:09:11-07:00" level=info msg="KubernetesReady=False: Waiting for k3s API server (Probing)"
                                                                  time="2026-06-15T16:09:11-07:00" level=info msg="ContainerEngineReady=True: Container engine synced (Connected)"
                                                          time="2026-06-15T16:09:11-07:00" level=info msg="Settled=False: Waiting for k3s API server (Probing)"
                                       time="2026-06-15T16:09:16-07:00" level=info msg="KubernetesReady=True: Kubernetes cluster is ready (Ready)"
                          time="2026-06-15T16:09:16-07:00" level=info msg="Settled=True: App has reached the desired state (Settled)"
             NAME              STATUS   ROLES    AGE   VERSION
                                                              desktop-r0d3f8v   Ready    <none>   1s    v1.34.6+k3s1

```

rdd/pkg/cli/console/console_windows.go
```diff
@@ -0,0 +1,49 @@
...
+// processing (ANSI colors). Best-effort: a no-op when the console is redirected.
+func Repair() {
+	for _, f := range []*os.File{os.Stdout, os.Stderr} {
+		h := windows.Handle(f.Fd())
+		var mode uint32
+		if windows.GetConsoleMode(h, &mode) != nil {
+			continue // not a console handle (redirected)
+		}
+		if mode&disableNewlineAutoReturn != 0 {
+			_ = windows.SetConsoleMode(h, mode&^disableNewlineAutoReturn)
```
Nino-K, line 34: I know this is supposed to be best effort and ignoring the error is intentional, but I just wanted to clarify this: If `SetConsoleMode` fails silently, the bit stays set, and every further `Repair()` call tries again and fails indefinitely with no obvious indication. Would it be worth logging a one-time warning so the staircase is at least diagnosable?

rdd/pkg/cli/console/console_windows.go
```diff
@@ -0,0 +1,49 @@
...
+const disableNewlineAutoReturn = 0x0008
+
+// Repair clears DISABLE_NEWLINE_AUTO_RETURN on the controlling console, undoing
+// the staircase a child leaves behind while preserving virtual-terminal
+// processing (ANSI colors). Best-effort: a no-op when the console is redirected.
+func Repair() {
+	for _, f := range []*os.File{os.Stdout, os.Stderr} {
+		h := windows.Handle(f.Fd())
+		var mode uint32
+		if windows.GetConsoleMode(h, &mode) != nil {
```
Nino-K, line 30: This is a bit awkward. After the first successful repair, the bit is clear, so every subsequent call does `GetConsoleMode` → bit not set → return. That seems to be one syscall per log line.

Would it make sense to add some state so we can avoid this and short-circuit it? 

```go
var repaired atomic.Bool

func Repair() {
    if repaired.Load() {
        return
    }
    // ...
    repaired.Store(true)
}
```

Nino-K: Discussed this offline with Jan, and it looks like you have to recheck every line because the other processes can repeatedly call wsl.exe, which will turn the bit back on.

Nino-K (approved):

PR #456, opened 2026-06-16
UI: Windows networking: Fix unresolved promise

mook-as:
This is causing the window to not show up on startup.  It's also breaking startup in other ways (because the rest of the startup code never runs), but that is the most obvious result.


jandubois (approved):
Thanks, LGTM

PR #457, opened 2026-06-16
Disable spell checker

jandubois:


mook-as (approved):
LGTM; I assume it would be good later to split this out to its own workflow too.

jandubois:
> LGTM; I assume it would be good later to split this out to its own workflow too.

Yes, we can do this when we want to enable it again

PR #458, opened 2026-06-16
Switch package version 1.99.0 → 2.0.0-alpha.1

jandubois:


mook-as (dismissed):
I think I prefer `alpha1`, just so if we end up with an `alpha2` nobody gets confused (and `alpha` with no suffix looks kind of weird to me).  But I don't prefer it enough to block this PR.

mook-as (approved):

PR #462, opened 2026-06-16
Fix Smoke Tests

mook-as:
This is a bunch of stuff aimed at fixing the smoke tests:

- Rename the Linux package to `rancher-desktop-2`
- Make the smoke tests use RDD
- Install qemu (because we don't do that in AppImage, zip)

There's also some debugging related stuff like collecting the Lima logs in smoke tests.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -15,6 +15,18 @@ on:
           Download artifacts from release with this tag, rather than picking the
           first draft release.
         type: string
+      obs-project:
+        description: >
+          OpenSUSE Build Service project to download AppImage from, in the form
+          of "home:username:projectname". The default is the devel project used
+          in final pre-release smoke tests.
+        type: string
+        default: "isv:Rancher:dev"
+      skip-signing-checks:
+        description: >
+          Skip signature checks on downloaded artifacts (fails the tests).
+        type: boolean
+        default: false
 
 jobs:
   download-artifacts:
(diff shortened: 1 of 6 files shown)


.github/workflows/smoke-test.yaml
```diff
@@ -316,10 +405,32 @@ jobs:
...
       with:
         name: logs-appimage-${{ matrix.id }}.zip
         path: ${{ github.workspace }}/logs
         if-no-files-found: warn
+
+  flag-skipped-signing-checks:
+    name: Flag that signing checks were skipped
+    needs: [smoke-test,repository-smoke-test,appimage-smoke-test]
+    runs-on: ubuntu-latest
+    if: inputs.skip-signing-checks == 'true'
```
jandubois: This can never be true because you are comparing a boolean to a string here with `==`, which will force numeric conversions.

```suggestion
    if: inputs.skip-signing-checks
```

jandubois (changes requested):
Mostly looks fine, but I don't think the `skip-signing-checks` flag works like this.

.github/workflows/smoke-test.yaml
mook-as: Good catch; tested in a throwaway repo† and you seem to be correct.  Commit amended, and rebased to pull in your Windows changes.

† The two runs at https://github.com/mook-as/junk/actions/workflows/conditional.yaml ; see the workflow on each run.

jandubois (approved):
Thanks, LGTM

PR #470, opened 2026-06-19
Merge from upstream

mook-as:
This merges the current state of Rancher Desktop.

This mainly pulls in (unused) container stats, plus a bunch of version updates.


mook-as:
Marking this as ready now because it looks like I forgot about it for three weeks…

jandubois (approved):
Thanks, LGTM

PR #474, opened 2026-06-19
CI: Smoke tests: Install qemu for Rocky

mook-as:
Since that is not available in the main repositories, we need to pull in a third-party repository for this.  In practice, this _probably_ means we cannot support Rancher Desktop 2 on Rocky and other RHEL derivatives.

With this, smoke tests should pass.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -347,14 +347,16 @@ jobs:
             suse|opensuse)
               zypper --non-interactive install qemu-headless
               exit 0;;
-            rocky|rhel|centos|fedora)
-              dnf install --assumeyes qemu-kvm
-              # On RHEL/CentOS, the qemu exectuable is not actually exposed;
-              # make a symlink to qemu-kvm, even though it's unsupported.
-              if ! command -v qemu-system-x86_64; then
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

jandubois:
Did we ever try something like this:

```
ln -s /usr/libexec/qemu-kvm /usr/bin/qemu-system-x86_64 
```

This is just from web search; I don't know which version they are shipping, and how compatible it is going to be.

mook-as:
Yes, I tried that first; eventually I got it to work enough to the point that it fell over because it didn't have once of the device drivers:

> `{"level":"debug","msg":"qemu[stderr]: qemu-kvm: -virtfs local,mount_tag=lima-23932c75e539853f,path=/home/ci-user,security_model=none: 'virtio-9p-pci' is not a valid device model name","time":"2026-06-16T22:58:26Z"}`

jandubois:
> > `{"level":"debug","msg":"qemu[stderr]: qemu-kvm: -virtfs local,mount_tag=lima-23932c75e539853f,path=/home/ci-user,security_model=none: 'virtio-9p-pci' is not a valid device model name","time":"2026-06-16T22:58:26Z"}`

We could check for this and disable the `9p` mountType. Maybe `reverse-sshfs` would still work?

Issue #463, opened 2026-06-16
GUI doesn't appear until the control plane is running


jandubois:
When there is an error starting the control plane (e.g. it exists, but was created by an incompatible version), then the GUI is never displayed, and no error is shown.

Even when everything is fine there is a multi-second delay before the GUI appears. We should display a disabled UI immediately, with a hints that we are waiting for an API connection to the backend.

If this is difficult with the main window, then maybe we need to consider a splashscreen, so users don't wonder if the application started at all, or already failed.

PR #475, opened 2026-06-20
RDA: Report error on incompatible control plane

mook-as:
This updates the UI so that if there is an incompatible control plane, we show a brief error instead of just waiting forever to start.

image

To test:
- Run `rdd service start` etc. on a different branch.
- Run `RDD_DEVELOPER_MODE=false yarn dev` on this branch.

See also: #463 — the second commit in the series makes the initial config fetch asynchronous, so we can show the window earlier.  (Note that we still need `rdd service paths` to finish, thogh.)

### DIFF
--- a/pkg/rancher-desktop/main/networking/index.ts
+++ b/pkg/rancher-desktop/main/networking/index.ts
@@ -41,7 +41,7 @@ export default async function setupNetworking() {
   http.globalAgent = proxyAgent;
   https.globalAgent = proxyAgent;
 
-  mainEvents.handle('rdd/kube-config-ready', (kubeConfigStr) => {
+  mainEvents.emit('rdd/certificate-callback', (kubeConfigStr) => {
     const kubeConfig = new KubeConfig();
     kubeConfig.loadFromString(kubeConfigStr);
     configureRDDAuthentication(kubeConfig);
@@ -61,10 +61,10 @@ export default async function setupNetworking() {
       verifyCertificate.bind(null, kubeCerts, getSystemCertificates));
 
     mainEvents.emit('network-ready');
-
-    return Promise.resolve();
   });
-  // Trigger RDD config loading, ignoring the result.
+  // Trigger RDD config loading, ignoring the result.  This lets us move the
+  // certificate handling setup outside of a potential hot path.  If this has
+  // already occurred, it would just ignore the cached result.
   mainEvents.invoke('rdd/kube-config').catch(console.error);
 }
 
--- a/pkg/rancher-desktop/main/networking/index.ts
+++ b/pkg/rancher-desktop/main/networking/index.ts
@@ -41,25 +41,31 @@ export default async function setupNetworking() {
   http.globalAgent = proxyAgent;
   https.globalAgent = proxyAgent;
 
-  const kubeConfig = new KubeConfig();
-  kubeConfig.loadFromString(await mainEvents.invoke('rdd/kube-config'));
-  configureRDDAuthentication(kubeConfig);
-  const kubeCerts =
-    Buffer.from(kubeConfig.getCurrentCluster()?.caData ?? '', 'base64')
-      .toString('utf-8')
-      .split(/(?=-----BEGIN CERTIFICATE-----)/g)
-      .map(cleanupCert)
-      .filter(x => x);
+  mainEvents.handle('rdd/kube-config-ready', (kubeConfigStr) => {
+    const kubeConfig = new KubeConfig();
+    kubeConfig.loadFromString(kubeConfigStr);
+    configureRDDAuthentication(kubeConfig);
+    const kubeCerts =
+      Buffer.from(kubeConfig.getCurrentCluster()?.caData ?? '', 'base64')
+        .toString('utf-8')
+        .split(/(?=-----BEGIN CERTIFICATE-----)/g)
+        .map(cleanupCert)
+        .filter(x => x);
 
-  // Set up certificate handling for specific hosts; we ignore the certificate
-  // completely on these, but limit them to specific ports.
-  Electron.app.on('certificate-error', handleCertificateError.bind(null, kubeCerts));
-  // Set up certificate handling for system certificates, as well as handling
-  // custom certificate authorities for RDD.
-  Electron.session.defaultSession.setCertificateVerifyProc(
-    verifyCertificate.bind(null, kubeCerts, getSystemCertificates));
+    // Set up certificate handling for specific hosts; we ignore the certificate
+    // completely on these, but limit them to specific ports.
+    Electron.app.on('certificate-error', handleCertificateError.bind(null, kubeCerts));
+    // Set up certificate handling for system certificates, as well as handling
+    // custom certificate authorities for RDD.
+    Electron.session.defaultSession.setCertificateVerifyProc(
+      verifyCertificate.bind(null, kubeCerts, getSystemCertificates));
 
-  mainEvents.emit('network-ready');
+    mainEvents.emit('network-ready');
+
+    return Promise.resolve();
+  });
+  // Trigger RDD config loading, ignoring the result.
+  mainEvents.invoke('rdd/kube-config').catch(console.error);
 }
 
 /**
--- a/pkg/rancher-desktop/main/rdd-ctl.ts
+++ b/pkg/rancher-desktop/main/rdd-ctl.ts
@@ -4,16 +4,19 @@ import stream from 'node:stream';
 import Electron from 'electron';
 
 import { getIpcMainProxy } from '@pkg/main/ipcMain';
-import mainEvents, { NoMainEventsHandlerError } from '@pkg/main/mainEvents';
+import mainEvents from '@pkg/main/mainEvents';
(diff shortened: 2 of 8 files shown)


rdd/pkg/cli/exit/exit.go
```diff
@@ -15,10 +15,13 @@ import (
 // reserved for cobra usage errors, so named codes start at 3.
 const (
 	// CodeRejected indicates the API server's admission controller rejected the request.
-	CodeRejected = 3
+	CodeRejected = iota + 3
```
jandubois: I'm not a fan of using `iota` for constants that are part of an external API. You can't renumber them without breaking the API.

The new exit code should be documented in `rdd/docs/design/cmd_app.md`.

pkg/rancher-desktop/main/networking/index.ts
```diff
@@ -41,25 +41,31 @@ export default async function setupNetworking() {
...
-  // Set up certificate handling for specific hosts; we ignore the certificate
-  // completely on these, but limit them to specific ports.
-  Electron.app.on('certificate-error', handleCertificateError.bind(null, kubeCerts));
-  // Set up certificate handling for system certificates, as well as handling
-  // custom certificate authorities for RDD.
-  Electron.session.defaultSession.setCertificateVerifyProc(
-    verifyCertificate.bind(null, kubeCerts, getSystemCertificates));
-
-  mainEvents.emit('network-ready');
+  mainEvents.handle('rdd/kube-config-ready', (kubeConfigStr) => {
```
jandubois: AI tells me that `fetchConfig` could be "resolved" before this handler is registered. And since the result of the first call is cached, this would mean certs would never get loaded.  It claims the fix is registering the handler before

```
for await (const cert of getSystemCertificates()) { ... }
```

Not sure this is actually true, but sounds plausible.

pkg/rancher-desktop/main/rdd-ctl.ts
```diff
@@ -1,49 +1,123 @@
...
+          .catch(err => console.error('Error fetching RDD version:', err));
 
         return stdout;
       }
     } catch (err) {
+      if (err instanceof SpawnError && err.code === 5) {
+        // Fatal: the server is using an incompatible backend.
+        let message = 'Unsupported RDD service version';
+        try {
+          message = JSON.parse(stderr).msg ?? message;
```
jandubois: You are relying on the loglevel to be `warn` or higher, so there is only a single log line. Maybe that is fine, but the dialog is not very informative now:

CleanShot 2026-06-24 at 18 06 58@2x

jandubois (changes requested):
Looks mostly fine. Personally I'm most interested in keeping the list of exit codes documented.

I can't concentrate enough to figure out if the race condition AI is blathering on is actually real or not.

pkg/rancher-desktop/main/rdd-ctl.ts
mook-as: Ah, you're getting the fallback message because the output had too many JSON messages? Hmm, will need to fix that somehow…

jandubois: We should loop over the lines and search for the one with the version error with a regex. Fallback only if none of the lines have a matching msg.

pkg/rancher-desktop/main/networking/index.ts
mook-as: The bug sounds plausible.  The suggested fix sounds extremely fragile, though.  So I added a thing to make sure catch `fetchConfig` would remember the fact that the handler wasn't correctly called, and just call it the next time it was called.

rdd/pkg/cli/exit/exit.go
mook-as: Reverted the use of `iota`, and updated the comment just above to say why we don't use it (so that the next time I come around here I don't try using `iota` again), dropping a link to the docs.  Also updated the docs, of course.

jandubois (dismissed):
I was just about to say "LGTM", but then I double checked with Claude, and it finds a bunch to complain about. I don't think any of them are blocking issues, so I will approve the PR, but the first 2 may be worth considering.

# PR #475 — RDA: Report error on incompatible control plane

## Findings (most-severe first)

### 1. The `rdd/kube-config-ready` event + `missingReadyCallback` global are an unnecessary, fragile push mechanism `pkg/rancher-desktop/main/rdd-ctl.ts:130` / `networking/index.ts:44,68`

`mainEvents.invoke('rdd/kube-config')` already *returns* the kubeconfig string (`mainEvents.ts:283`), and networking already issues that exact call (`networking/index.ts:68`). Networking could consume the return value directly (`invoke('rdd/kube-config').then(configureCerts).catch(...)`), deleting the second event, the module-global `missingReadyCallback`, and the deferred re-fire block (`rdd-ctl.ts:133-142`). Because `fetchConfig` memoizes the promise, every caller gets the same config regardless of who fetched first — the pull model has no ordering race for the flag to work around.

**Failure scenario:** `missingReadyCallback` is a one-shot cleared unconditionally (`rdd-ctl.ts:141`) even when its re-fire `invoke` itself throws `NoMainEventsHandlerError` (swallowed by `.catch` on line 139). If any non-networking caller (a renderer IPC via `getIpcMainProxy`) triggers `fetchConfig` before `setupNetworking` registers the handler, the signal is consumed and lost, and cert/auth setup never runs. Currently masked only because `setupNetworking` registers the handler before the first fetch and the window opens afterward — any reordering silently breaks it.

### 2. A throw from the `rdd/kube-config-ready` handler is swallowed with no retry, and the config is cached anyway `pkg/rancher-desktop/main/rdd-ctl.ts:73`

The networking handler does synchronous throwing work (`kubeConfig.loadFromString`, and `configureRDDAuthentication` throws `'No currently active cluster'` when the cluster/server is missing).

**Failure scenario:** if `stdout` passes `includes('apiVersion')` but the handler rejects with anything other than `NoMainEventsHandlerError`, the `else` branch (line 78-81) just logs `'Error processing new RDD configuration'` and `fetchConfigWithoutCache` returns `stdout` normally — `setCertificateVerifyProc` / Authorization injection are never installed, `network-ready` never fires, and the resolved promise is cached so no retry happens. The pre-PR code ran this inline in the awaited `setupNetworking`, so a throw surfaced loudly instead of degrading silently. (Trigger is unlikely since the daemon emits a valid kubeconfig, but the error path is a regression in visibility.)

### 3. In JSON mode, a propagated child-exit error with `nil` Err now logs an empty-message error line

`rdd/cmd/rdd/main.go:181`

The previous code guarded on `exitErr.Err != nil` precisely to avoid a bare `level=error` line for child-exit-code propagation. The new JSON branch logs unconditionally.

**Failure scenario:** `rdd run <cmd> --log-format=json` where the child exits non-zero yields `&cliexit.Error{Code:N}` with `nil` Err → emits `{"level":"error","msg":"","exit-code":N}`. The `exitErr.Err != nil` comment on line 184 is now stale in JSON mode. Low severity (this PR's own `service config` path always carries a message), but it reintroduces the noise the guard prevents.

### 4. The fatal quit is gated solely on the numeric exit code 5, not on a positive incompatibility signal

`pkg/rancher-desktop/main/rdd-ctl.ts:93`

The JSONL parse (lines 98-106) only customizes the displayed message; the decision to `showErrorBox` + `app.quit()` rests entirely on `err.code === 5`. Meanwhile code 5 (`CodeIncompatibleServer`) is now a documented public contract (`cmd_app.md`), and `rdd run` (`run.go:177`) propagates arbitrary child exit codes including 5.

**Failure scenario:** contained today because the TS only runs `service config` (whose only source of exit 5 is the version check), but the app's life-or-death decision depends on that scoping rather than on confirming an `exit-code:5` record actually appeared in stderr. Consider gating the quit on the parsed signal, not just the process status.

### 5. `rdd --version=raw` is spawned fire-and-forget on every successful fetch purely to log a version string

`pkg/rancher-desktop/main/rdd-ctl.ts:84`

Detached from the returned promise, so its `'RDD version:'` line lands at a nondeterministic point in startup logs (reducing rather than aiding debuggability), and the inner `({ stdout })` shadows the outer `stdout`. Minor extra subprocess; worth confirming the daemon doesn't already log its version.

### 6. Typo in comment

`pkg/rancher-desktop/main/rdd-ctl.ts:28`

`stderr friom the` → `from`. Cosmetic.

## What holds up

The Go exit-code-5 plumbing itself is correct end-to-end (`checkSupportedVersion` → `Wait` → `Classify` → `main.go` → `os.Exit(5)`), the JSON field names (`exit-code`, `msg`) match the TS parser exactly, the `errors.As(err, &cliError)` pointer handling is right, and there are no dangling references to the removed `errorServerVersionUnsupported` type.

mook-as:
Adjusted the mechanism used for the cert callback so it's always called before `fetchConfigWithoutCache` returns. This makes the thing a bit cleaner, I think.

jandubois (approved):
Thanks, LGTM

I have an AI review, but I don't think anything is actionable. Please merge unless you want to change anything.

https://jandubois.github.io/rancher-desktop-2/20260702-164658-pr-475.html

mook-as:
Yeah, don't care right now

PR #478, opened 2026-06-22
E2E: Fix startup test

mook-as:
This fixes the (Playwright) E2E test for the basic startup test:
- It now explicitly runs the daemon with all controllers disabled, then separately runs the mock controller.
- The mock controller now includes all the CRDs.
- The mock controller also automatically reconciles the app object by marking it as settled.  This allows the front end to proceed.
- We can then test that the front end removes the progress bar when it is settled, and that the app object (which we manually retrieve via `rdd ctl get app`) has `running` set.
- We also test that making the app not Settled restores the progress bar.
- `backend.e2e.spec.ts` has been removed, as that tests the JS backend, which no longer exists; any RDD-specific tests are in BATS.


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-daemon/20260624-165529-pr-478.html

mook-as:
> I1. Mock Settled omits ObservedGeneration

Will fix.

> I2. startSlowerDesktop dropped a parameter that five suites still pass

Those suites are currently disabled and will need to be rewritten (or dropped) anyway.

> I3. beforeExit cleanup spawns async work

Cleaned up a bit to not use `beforeExit`.

> S1. Mock reconciler updates status on every reconcile

Fixed; the suggestion interacts badly with I1, though, so I did something that takes that into account.

> S2. waitForAppSettled leaks its timer

Switched to `AbortController` and dropped the lint comment, but the base premise is dumb: it's a test, a single 30s timeout isn't worth thinking about.

> S3. 1 s progress-bar timeout risks CI flakes

Because of the mock controller, setting `running=false` means the app will never settle, so there's no worry about racing.

> S4. Defensive CRD Get cannot fire

Good catch; we previously used the real controllers for the CRDs.  Dropped the checks in all the parallel reconcilers (image, container, etc.).

> S5. Doc comment typo and non-idiomatic godoc

Huh, I wonder why I thought it should start with `type …`.

> S6. appReconciler.Recorder is assigned but never read

Dropped; dead code copied from other reconcilers in the package.

jandubois (dismissed):
Thanks, LGTM

Some minor issues in updated AI review, in case you want to address any: https://jandubois.github.io/rancher-desktop-daemon/20260626-120041-pr-478.html

jandubois (changes requested):
Almost done, but please fix I1 and (maybe) I2.

https://jandubois.github.io/rancher-desktop-2/20260702-214637-pr-478.html

jandubois (approved):
Thanks, LGTM

Issue #319, opened 2026-04-16
Provide hooks for service stop and deletion


jandubois:
We need a hook so that `rdd svc stop` can signal other resources to shut down gracefully:

* `LimaVM` instances need to shut down
* `App` instance may need to clean up docker context

The cleanup hook also needs to be executed for `rdd svc delete`. Even if we don't stop VMs, we still need to clean up resources outside the directories that will be deleted by the service deletion.

PR #479, opened 2026-06-23
Recover from orphaned WSL2 distro registrations

jandubois:
A WSL2 distro can stay registered with WSL after its `ext4.vhdx` is gone: `rdd svc delete` removed each instance directory but left the `wsl --list` registration behind. Lima's WSL2 driver only re-imports a distro that is absent from `wsl --list`, so the next start booted that diskless registration into `ERROR_FILE_NOT_FOUND` on every attempt and the reconciler looped forever.

Two commits — recovery, then prevention:

- **lima: re-import an orphaned WSL2 distro on start** — `startInstance` detects a distro that is registered but missing its disk and unregisters it, so the hostagent re-imports a fresh `ext4.vhdx`. This heals orphans left by older builds or a failed unregister.
- **service: unregister WSL2 distros on delete** — `rdd svc delete` now unregisters each instance's distro before removing its directory, so it stops minting orphans in the first place.

Both paths are Windows-only and no-ops elsewhere.

Routing `svc delete` cleanup through the lima controller — rather than the duplicate unregister that currently lives in the service layer — is tracked by #319.

### DIFF
--- a/rdd/docs/design/cmd_service.md
+++ b/rdd/docs/design/cmd_service.md
@@ -116,8 +116,8 @@ directory.
 
 On Windows it also unregisters each Lima instance's WSL2 distro before removing
 the Lima home, because deleting the directory drops the distro's `ext4.vhdx` but
-leaves the `wsl --list` registration behind, which the next `rdd service create`
-would boot against a missing disk.
+leaves the `wsl --list` registration behind, which would cause the next
+`rdd service create` to boot against a missing disk.
 
 Delete always waits for the control plane to exit before removing files, because
 removing the instance directory under a live process corrupts it on Windows and
--- a/rdd/docs/design/cmd_service.md
+++ b/rdd/docs/design/cmd_service.md
@@ -114,6 +114,11 @@ Stops the control plane and removes the instance directory, the short directory
 (which contains the Lima home), and — unless `RDD_KEEP_LOGS` is set — the log
 directory.
 
+On Windows it also unregisters each Lima instance's WSL2 distro before removing
+the Lima home, because deleting the directory drops the distro's `ext4.vhdx` but
+leaves the `wsl --list` registration behind, which the next `rdd service create`
+would boot against a missing disk.
+
 Delete always waits for the control plane to exit before removing files, because
 removing the instance directory under a live process corrupts it on Windows and
 breaks PID-file mutual exclusion on Unix. Use `--timeout` to bound that wait
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -33,6 +32,7 @@ import (
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/util/logfile"
 	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/util/process"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/util/wsl"
 )
 
 // hostSwitchRetryInterval is the minimum time between restarts of a host-switch
@@ -812,38 +812,18 @@ func stopInstanceForcibly(ctx context.Context, logger logr.Logger, inst *limatyp
 	}
 }
 
-// terminateWSL2Distro sends `wsl.exe --terminate` for the Lima distro with
-// the given instance name. No-op on non-Windows. Best-effort with a
-// 10-second timeout: wsl.exe can hang if the WSL subsystem is degraded.
+// terminateWSL2Distro terminates the WSL2 distro backing instName, logging any
+// failure. No-op on non-Windows.
 func terminateWSL2Distro(ctx context.Context, logger logr.Logger, instName string) {
-	if runtime.GOOS != "windows" {
-		return
-	}
-	distroName := "lima-" + instName
-	wslCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
-	defer cancel()
-	if err := exec.CommandContext(wslCtx, "wsl.exe", "--terminate", distroName).Run(); err != nil {
-		logger.V(1).Info("Failed to terminate WSL2 distro", "distro", distroName, "error", err)
+	if err := wsl.Terminate(ctx, wsl.DistroName(instName)); err != nil {
+		logger.V(1).Info("Failed to terminate WSL2 distro", "instance", instName, "error", err)
 	}
 }
 
-// unregisterWSL2Distro sends `wsl.exe --unregister` for the Lima distro with
-// the given instance name, returning nil on non-Windows. This removes the WSL2
-// registration and deletes the distro's ext4.vhdx, allowing Lima to import a
-// fresh distro on the next Prepare. It uses a 30-second timeout (unregister is
-// slower than terminate and can hang if WSL2 is degraded) and leaves the
-// failure handling to the caller.
+// unregisterWSL2Distro unregisters the WSL2 distro backing instName, returning
+// the failure for the caller to handle. No-op on non-Windows.
 func unregisterWSL2Distro(ctx context.Context, instName string) error {
-	if runtime.GOOS != "windows" {
-		return nil
-	}
-	distroName := "lima-" + instName
-	wslCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-	defer cancel()
-	if err := exec.CommandContext(wslCtx, "wsl.exe", "--unregister", distroName).Run(); err != nil {
-		return fmt.Errorf("unregister WSL2 distro %q: %w", distroName, err)
-	}
-	return nil
+	return wsl.Unregister(ctx, wsl.DistroName(instName))
 }
 
 // removeStaleInstance removes a stale Lima instance directory left behind by
@@ -857,7 +837,7 @@ func removeStaleInstance(ctx context.Context, logger logr.Logger, instName, inst
 	// forceStopForDeletion guards against. Both calls are no-ops off Windows.
 	terminateWSL2Distro(ctx, logger, instName)
 	if err := unregisterWSL2Distro(ctx, instName); err != nil {
-		logger.V(1).Info("Failed to unregister WSL2 distro", "instance", instName, "error", err)
+		logger.Error(err, "Failed to unregister WSL2 distro; a stale registration may remain", "instance", instName)
 	}
 	return os.RemoveAll(instanceDir)
 }
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -38,6 +39,10 @@ import (
 // that exited unexpectedly, and the delay before handleWatchedState re-checks
 // afterward.
 const hostSwitchRetryInterval = 15 * time.Second
+
+// wsl2RootDisk is the root-disk filename Lima's WSL2 driver creates when it
+// imports a distro with `wsl.exe --import`.
+const wsl2RootDisk = "ext4.vhdx"
 
 func (r *LimaVMReconciler) handleDeletion(ctx context.Context, limaVM *v1alpha1.LimaVM) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
@@ -344,6 +349,19 @@ func (r *LimaVMReconciler) handleRestartNeeded(ctx context.Context, limaVM *v1al
 	}
 }
 
+// wsl2RegistrationIsOrphaned reports whether inst is a WSL2 distro still
+// registered with WSL but missing its root disk — a state Lima cannot recover
+// on its own, because its WSL2 driver only re-imports a distro that is absent
+// from `wsl --list`. It arises when an instance directory is removed without
+// `wsl --unregister`. Always false off Windows, where VMType is never WSL2.
+func wsl2RegistrationIsOrphaned(inst *limatype.Instance) bool {
+	if inst.VMType != limatype.WSL2 || inst.Status == limatype.StatusUninitialized {
+		return false
+	}
+	_, err := os.Stat(filepath.Join(inst.Dir, wsl2RootDisk))
+	return errors.Is(err, os.ErrNotExist)
+}
+
 // startInstance launches the hostagent and starts a watcher goroutine to track
 // its lifecycle. The watcher triggers reconciles as the hostagent progresses
 // through Starting → Running → Stopped, so no polling is needed.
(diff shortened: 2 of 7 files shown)


rdd/docs/design/cmd_service.md
```diff
@@ -114,6 +114,11 @@ Stops the control plane and removes the instance directory, the short directory
 (which contains the Lima home), and — unless `RDD_KEEP_LOGS` is set — the log
 directory.
 
+On Windows it also unregisters each Lima instance's WSL2 distro before removing
+the Lima home, because deleting the directory drops the distro's `ext4.vhdx` but
+leaves the `wsl --list` registration behind, which the next `rdd service create`
+would boot against a missing disk.
```
mook-as:
```suggestion
leaves the `wsl --list` registration behind, which would cause the next
`rdd service create` to boot against a missing disk.
```

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -344,6 +349,19 @@ func (r *LimaVMReconciler) handleRestartNeeded(ctx context.Context, limaVM *v1al
 	}
 }
 
+// wsl2RegistrationIsOrphaned reports whether inst is a WSL2 distro still
+// registered with WSL but missing its root disk — a state Lima cannot recover
+// on its own, because its WSL2 driver only re-imports a distro that is absent
+// from `wsl --list`. It arises when an instance directory is removed without
+// `wsl --unregister`. Always false off Windows, where VMType is never WSL2.
+func wsl2RegistrationIsOrphaned(inst *limatype.Instance) bool {
```
mook-as, line 357: I wish we could interrogate WSL about this, but I don't see a command line option for that.  Not even the [API](https://learn.microsoft.com/en-us/windows/win32/api/wslapi/nf-wslapi-wslgetdistributionconfiguration) has that information…

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -32,13 +32,18 @@ import (
...
 )
 
 // hostSwitchRetryInterval is the minimum time between restarts of a host-switch
 // that exited unexpectedly, and the delay before handleWatchedState re-checks
 // afterward.
 const hostSwitchRetryInterval = 15 * time.Second
 
+// wsl2RootDisk is the root-disk filename Lima's WSL2 driver creates when it
+// imports a distro with `wsl.exe --import`.
+const wsl2RootDisk = "ext4.vhdx"
```
mook-as, line 45: I can see this biting us later when WSL starts supporting, say, btrfs disks.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -363,6 +381,27 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
...
+	// Heal an orphaned WSL2 registration before the hostagent boots it (see
+	// wsl2RegistrationIsOrphaned). Unregistering makes Lima's Start (below, in
+	// the hostagent) treat the distro as uninitialized and re-import basedisk,
+	// instead of booting a missing disk into ERROR_FILE_NOT_FOUND forever.
+	if wsl2RegistrationIsOrphaned(inst) {
+		logger.Info("WSL2 distro is registered but its root disk is missing; unregistering to force re-import",
+			"instance", limaVM.Name, "disk", wsl2RootDisk)
+		// Terminate before unregistering: wsl.exe --unregister can deadlock on a
+		// distro that still holds kernel state. Both are no-ops off Windows.
+		terminateWSL2Distro(ctx, logger, limaVM.Name)
```
mook-as: How can the distro be running if the disk is missing‽ Remember that opening a file (with read access) creates a lock that prevents the file from being deleted.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -363,6 +381,27 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
...
 
+	// Heal an orphaned WSL2 registration before the hostagent boots it (see
+	// wsl2RegistrationIsOrphaned). Unregistering makes Lima's Start (below, in
+	// the hostagent) treat the distro as uninitialized and re-import basedisk,
+	// instead of booting a missing disk into ERROR_FILE_NOT_FOUND forever.
+	if wsl2RegistrationIsOrphaned(inst) {
+		logger.Info("WSL2 distro is registered but its root disk is missing; unregistering to force re-import",
+			"instance", limaVM.Name, "disk", wsl2RootDisk)
+		// Terminate before unregistering: wsl.exe --unregister can deadlock on a
+		// distro that still holds kernel state. Both are no-ops off Windows.
```
mook-as: Non-Windows should never have `wsl2RegistrationIsOrphaned` returning true…

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -773,36 +812,18 @@ func stopInstanceForcibly(ctx context.Context, logger logr.Logger, inst *limatyp
 	}
 }
 
-// terminateWSL2Distro sends `wsl.exe --terminate` for the Lima distro with
-// the given instance name. No-op on non-Windows. Best-effort with a
-// 10-second timeout: wsl.exe can hang if the WSL subsystem is degraded.
+// terminateWSL2Distro terminates the WSL2 distro backing instName, logging any
+// failure. No-op on non-Windows.
 func terminateWSL2Distro(ctx context.Context, logger logr.Logger, instName string) {
```
mook-as, line 814: Why do these wrappers need to exist? Why do the callers use `wsl.Terminate` / `wsl.Unregister` directly? They're still no-ops on not-Windows.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
mook-as, line 357: Should this live in the `wsl` package too? I guess it's here because it deals with `limatype.Instance`?

rdd/pkg/service/cmd/service.go
```diff
@@ -561,10 +562,70 @@ func Delete(ctx context.Context, timeout time.Duration) error {
...
+// removed. It terminates each distro first, because `wsl --unregister` can
+// deadlock on a distro that still holds kernel state. Both calls are best-effort
+// and time-bounded, since wsl.exe can hang when the WSL subsystem is degraded.
+// No-op on non-Windows.
+//
+// TODO: This reimplements cleanup the lima controller already owns
+// (unregisterWSL2Distro), because `svc delete` stops the daemon before
+// reaching here. Route it through the controller once the stop/delete hook
+// (rancher-desktop-2#319) exists.
+func unregisterInstanceDistros(ctx context.Context) {
```
mook-as: Shouldn't this live in package `wsl` too?

rdd/pkg/service/cmd/service.go
```diff
@@ -561,10 +562,70 @@ func Delete(ctx context.Context, timeout time.Duration) error {
...
+		}
+	}
+}
+
+// instanceDistroNames returns the WSL2 distro name ("lima-<instance>") for each
+// Lima instance directory under limaHome. It skips Lima's reserved bookkeeping
+// directories (those prefixed with "_" or "."), which are never instances, the
+// same way Lima's own store.Instances does. It returns nil when limaHome does
+// not exist (no instances were ever created).
+func instanceDistroNames(limaHome string) []string {
```
mook-as: Why is it doing this? Isn't the instance name fixed?

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -363,6 +381,27 @@ func (r *LimaVMReconciler) startInstance(ctx context.Context, limaVM *v1alpha1.L
 		return ctrl.Result{}, err
 	}
 
+	// Heal an orphaned WSL2 registration before the hostagent boots it (see
+	// wsl2RegistrationIsOrphaned). Unregistering makes Lima's Start (below, in
+	// the hostagent) treat the distro as uninitialized and re-import basedisk,
+	// instead of booting a missing disk into ERROR_FILE_NOT_FOUND forever.
+	if wsl2RegistrationIsOrphaned(inst) {
+		logger.Info("WSL2 distro is registered but its root disk is missing; unregistering to force re-import",
+			"instance", limaVM.Name, "disk", wsl2RootDisk)
```
mook-as, line 390: Note that `limaVM.Name` is always `rd`, because:
https://github.com/rancher-sandbox/rancher-desktop-2/blob/87495c727648ae2d10e75e14a8a26d09a66ee2e0/rdd/pkg/controllers/app/app/controllers/app_controller.go#L399-L401
https://github.com/rancher-sandbox/rancher-desktop-2/blob/87495c727648ae2d10e75e14a8a26d09a66ee2e0/rdd/pkg/controllers/app/app/controllers/app_controller.go#L51-L53
This means having multiple instances does not work at all on Windows (because everyone uses the `lima-rd` WSL distro).

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"testing"
+
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"gotest.tools/v3/assert"
+)
+
+func TestWSL2RegistrationIsOrphaned(t *testing.T) {
+	// newDir returns a fresh instance directory, optionally containing the
+	// WSL2 root disk so the predicate sees a registration with its disk intact.
+	newDir := func(t *testing.T, withDisk bool) string {
```
mook-as, line 19: I would trust these more if they actually ran `wsl --install --from-file` with an empty tar file and then deleted the disk as needed… but I guess that's not _quite_ what you're testing, even though I feel like that makes more sense.

rdd/pkg/util/wsl/wsl.go
```diff
@@ -0,0 +1,55 @@
...
+	return run(ctx, unregisterTimeout, "--unregister", distroName)
+}
+
+func run(ctx context.Context, timeout time.Duration, verb, distroName string) error {
+	if runtime.GOOS != "windows" {
+		return nil
+	}
+	wslCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	if err := exec.CommandContext(wslCtx, "wsl.exe", verb, distroName).Run(); err != nil {
```
mook-as: You may want to set `WSL_UTF8` whenever running `wsl.exe`.

mook-as (changes requested):
I am mostly confused by things like why we think there's multiple distro names (it's always `lima-rd`), etc.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Yeah, not sure what we should do about it now; we deal with it when it happens.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Yes, it is, but I guess all the `wsl` code is specific to Lima anyways, so maybe it makes sense to keep it all together.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: We cannot have multiple "rd" instances because the App is also a singleton. But we can have multiple VMs with different names, even though we don't advertise this yet.

But you raise a good point: on Windows the distro names must be globally unique, not just unique per RDD instance. So we need to pick a different prefix for each RDD instance to avoid collisions.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: The "unregister" wrapper is indeed redundant, but the "terminate" wrapper is called multiple times and includes the name lookup and consistent logging settings, so I'm going to keep it. Looks kind of asymmetric to have a wrapper for one and not the other, but so be it.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle_test.go
jandubois: Yeah, that should probably be done in a BATS test. But I think the current code simulates the actual behaviour correctly.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
mook-as: Filed as #482 so we can defer thinking about it

rdd/pkg/util/wsl/wsl.go
```diff
@@ -0,0 +1,119 @@
...
+	"os/exec"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// wsl.exe can hang when the WSL subsystem is degraded, so each call is
+// time-bounded. --unregister is slower than --terminate.
```
mook-as, line 23: Well, yeah, `--unregister` is `--terminate` + remove disk and stuff?

rdd/pkg/util/wsl/wsl.go
```diff
@@ -0,0 +1,119 @@
...
+
+// DistroName returns the WSL2 distro name Lima registers for instName. Lima's
+// WSL2 driver names each distro "lima-<instance>".
+func DistroName(instName string) string {
+	return "lima-" + instName
+}
+
+// Terminate runs `wsl.exe --terminate` to shut the distro down, releasing the
+// kernel state that can make a following --unregister deadlock. No-op off
+// Windows.
```
mook-as, line 37: _No-op off …_ seems like odd phrasing, but I can't think of any good alternatives.  Some so-so ones:

- Only applies to Windows.
- No-op on non-Windows.

rdd/pkg/util/wsl/wsl.go
```diff
@@ -0,0 +1,119 @@
...
+// controller once the stop/delete hook (rancher-desktop-2#319) exists.
+func UnregisterInstanceDistros(ctx context.Context, limaHome string) {
+	if runtime.GOOS != "windows" {
+		return
+	}
+	for _, distroName := range instanceDistroNames(limaHome) {
+		// Terminating is a best-effort precursor — the distro is usually already
+		// stopped — so log its failure at debug.
+		if err := Terminate(ctx, distroName); err != nil {
+			logrus.WithError(err).WithField("distro", distroName).Debug("wsl --terminate failed during delete")
```
mook-as, line 85: I keep forgetting where we use logrus and where we use klog. Not important.

rdd/pkg/util/wsl/wsl.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		// A failed unregister is the failure that matters: it leaves behind the
+		// stale registration this cleanup exists to prevent, so log it at warn.
+		// (A distro that was never imported also fails here, harmlessly.)
+		if err := Unregister(ctx, distroName); err != nil {
```
mook-as, line 91: Doesn't `wsl --unregister` already internally run `--terminate`?

mook-as (approved):
Nothing that blocks merging. It feels weird that we're duplicating bits of Lima, but I think that's just because Lima isn't very good as a library (and doesn't want to be one).

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle_test.go
jandubois: #522 implements a BATS test with a real orphaned WSL2 distro

PR #484, opened 2026-06-23
RDD: Bump our forked lima

### LINKED ISSUE #481: Windows: RDD gets wedged due to invalid host agent PID
### Actual Behavior

Attempting to start RDD can end up wedged on Windows.  The host agent PID file appears to contain an invalid PID.

### Steps to Reproduce

1. Be running Windows.
2. `rdd set running=true kubernetes.enabled=true`
   I don't know why Kubernetes is required yet, but I don't seem to be able to reproduce without it.
3. `rdd service stop`
4. `rdd set running=true`

### Result

Host agent keeps restarting; `ha.stderr.*.log` just has a single line:

> time="2026-06-23T12:28:03-07:00" level=fatal msg="failed to determine if another hostagent is running: OpenProcess: The parameter is incorrect."

### Expected Behavior

RDD should bring the VM up successfully.

### Additional Information

Running on `main` + merged #479

https://github.com/lima-vm/lima/blob/2a9cb9a4636b822057bff6abc6a138a0ec640550/pkg/store/instance.go#L214-L221

On Windows, `os.FindProcess` returns an error if the pid is invalid (because it would otherwise contain a valid handle).  On *nix, `os.FindProcess` returns a process (holding the pid) with nil error, and the user needs to signal the process to know if the result is valid.

### Rancher Desktop Version

70e78738c6164bb7057e6b4aac21ab5ee1d2cdf9 (main at 87495c727648ae2d10e75e14a8a26d09a66ee2e0 + #479 at 813f15857ce46d67847a2862f171824119a40460)

### Rancher Desktop K8s Version

1.34.6

### Which container engine are you using?

moby (docker cli)

### What operating system are you using?

Windows

### Operating System / Build Version

Windows 11 Pro 26H2 Build 26200.8655

### What CPU architecture are you using?

x64

### Linux only: what package format did you use to install Rancher Desktop?

None

### Windows User Only

_No response_

mook-as:
This pulls in only https://github.com/rancher-sandbox/lima/pull/159 and its merge commit; this fixes an issue on Windows where a stale host agent PID file causes the application to never reconcile.

### DIFF
--- a/rdd/go.mod
+++ b/rdd/go.mod
@@ -468,7 +468,7 @@ require (
 replace (
 	github.com/containerd/ltag => github.com/mook-as/ltag v0.0.0-20260224191129-27fed92c7adf
 	// Lima from the windows-plain-ci fork branch, pending lima-vm/lima#4998
-	github.com/lima-vm/lima/v2 => github.com/rancher-sandbox/lima/v2 v2.1.3-0.20260619054522-25bdf86634d8
+	github.com/lima-vm/lima/v2 => github.com/rancher-sandbox/lima/v2 v2.1.3-0.20260623214115-cb47878f2620
 	k8s.io/api => k8s.io/api v0.35.4
 	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.35.4
 	k8s.io/apimachinery => k8s.io/apimachinery v0.35.4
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #485, opened 2026-06-23
Bump embedded distro to v0.2.8

mook-as:
v0.2.8 makes k3s and the guest agent use the configured Kubernetes port when k3s runs in a network namespace (WSL), and raises the Virtualization:containers repo priority above the primary Leap repo so its packages win.

This is a replay of #477 (using the same head commit).


jandubois (approved):
Thanks, LGTM

PR #486, opened 2026-06-24
Replace Windows console-event interrupt with a role-keyed named event

jandubois:
On Windows, `rdd svc stop`/`delete` could not gracefully stop a control plane the GUI app had started in another console. `Interrupt` used `GenerateConsoleCtrlEvent(CTRL_BREAK)`, which reaches only processes sharing the caller's console, so it fell back to a hard kill that orphaned hostagents. The identity check compounded this: it read the target's command line from its PEB and compared image paths, so a daemon started from a different `rdd.exe` — app bundle vs standalone, or a binary moved aside by a reinstall — read as "not ours". `PID()` then deleted the pidfile, `delete` skipped the shutdown, and the live daemon was orphaned holding state.db.

`Interrupt` now signals a per-process named event, `Local\rdd-interrupt-<key>-<pid>`, where `<key>` is the process's role: `serve` for the daemon, `hostagent-<instance>` for each hostagent. Every process that can be a target registers its event at startup with `RegisterInterruptHandler`; `Interrupt` opens and sets it. Delivery no longer needs a shared console, and identity (`IsOurProcess`) becomes a single `OpenEvent` on that name — path-independent, so any `rdd` binary is recognised, and recycle-PID safe, since an unrelated PID created no such event. Keying by role and instance also keeps a recycled PID from one hostagent being mistaken for another. The whole PEB/command-line/image-path machinery goes away.

`Interrupt`, `RegisterInterruptHandler`, and `IsOurProcess` take a key on every platform; Unix ignores it (`Interrupt` is `SIGINT`, `IsOurProcess` a no-op). `CREATE_NEW_PROCESS_GROUP` stays only to detach background processes from the console's Ctrl-C; `Kill`/`KillTree` remain the force-kill fallback.

Two guardrails close the residual "alive but never registered" case (failed registration):

- `PID()` removes the pidfile only when the recorded PID is dead (new IsAlive); a live but unconfirmed PID keeps its file, which nothing acts on and the next Start overwrites.

- `Delete` removes `db/` first and aborts if `state.db` is still locked, before it can strip the pidfile and leave a half-deleted instance dir.

### DIFF
--- a/rdd/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
@@ -197,11 +197,9 @@ func (r *LimaVMReconciler) waitForProcessExit(ctx context.Context, name string) 
 	}
 }
 
-// signalHostagent sends a graceful shutdown signal to the hostagent process.
-// Uses process.Interrupt, which sends SIGINT on Unix and signals the hostagent's
-// named interrupt event on Windows (keyed by instance and PID).
-// Returns false if no watcher exists, the process has already been reaped (its
-// PID may be reused on Windows), or the signal could not be delivered.
+// signalHostagent sends a graceful shutdown signal (process.Interrupt) to the
+// hostagent process. Returns false if no watcher exists, the process has already
+// been reaped, or the signal could not be delivered.
 func (r *LimaVMReconciler) signalHostagent(name string) bool {
 	r.instanceStatesMu.RLock()
 	state, ok := r.instanceStates[name]
@@ -209,10 +207,9 @@ func (r *LimaVMReconciler) signalHostagent(name string) bool {
 	if !ok || state.cmd == nil || state.cmd.Process == nil {
 		return false
 	}
-	// Skip a process that has already been reaped: once procExited is closed,
-	// cmd.Wait() has released the OS handle and the PID may be reused. The named
-	// interrupt event makes a reused PID harmless — a recycled process has no such
-	// event, so Interrupt simply fails — but there is nothing to signal here.
+	// Skip a process that has already been reaped. Once procExited is closed,
+	// cmd.Wait() has released the OS handle and the PID may be reused, so there is
+	// nothing safe to signal here.
 	select {
 	case <-state.procExited:
 		return false
--- a/rdd/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
@@ -198,8 +198,8 @@ func (r *LimaVMReconciler) waitForProcessExit(ctx context.Context, name string) 
 }
 
 // signalHostagent sends a graceful shutdown signal to the hostagent process.
-// Uses process.Interrupt which sends SIGINT on Unix and CTRL_BREAK on Windows
-// (targeted at the hostagent's process group, not the parent).
+// Uses process.Interrupt, which sends SIGINT on Unix and signals the hostagent's
+// named interrupt event on Windows (keyed by instance and PID).
 // Returns false if no watcher exists, the process has already been reaped (its
 // PID may be reused on Windows), or the signal could not be delivered.
 func (r *LimaVMReconciler) signalHostagent(name string) bool {
@@ -209,17 +209,16 @@ func (r *LimaVMReconciler) signalHostagent(name string) bool {
 	if !ok || state.cmd == nil || state.cmd.Process == nil {
 		return false
 	}
-	// Once procExited is closed, cmd.Wait() has released the OS process handle
-	// and the PID may be reused; signalling it could deliver CTRL_BREAK to an
-	// unrelated console process on Windows. While it is open the handle is
-	// normally still held; cmd.Wait() releases it just before closing the
-	// channel, leaving a brief window.
+	// Skip a process that has already been reaped: once procExited is closed,
+	// cmd.Wait() has released the OS handle and the PID may be reused. The named
+	// interrupt event makes a reused PID harmless — a recycled process has no such
+	// event, so Interrupt simply fails — but there is nothing to signal here.
 	select {
 	case <-state.procExited:
 		return false
 	default:
 	}
-	return process.Interrupt(state.cmd.Process.Pid) == nil
+	return process.Interrupt(process.HostagentInterruptKey(name), state.cmd.Process.Pid) == nil
 }
 
 // enqueueReconcile sends a GenericEvent to trigger a reconcile for the named VM.
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
@@ -633,12 +633,10 @@ func (r *LimaVMReconciler) shutdownAllHostagents() {
 		return
 	}
 
-	// Send graceful shutdown signal to all hostagents in parallel. Skip any whose
-	// process has already been reaped: once cmd.Wait() closes procExited it has
-	// released the OS process handle, freeing the PID for reuse. The named
-	// interrupt event keyed by instance and PID makes a recycled PID harmless (it
-	// has no such event, so Interrupt fails), but there is nothing to signal once
-	// the process is reaped.
+	// Send a graceful shutdown signal to all hostagents in parallel. Skip any whose
+	// process has already been reaped; once cmd.Wait() closes procExited it has
+	// released the OS handle and the PID may be reused, so there is nothing to
+	// signal.
 	for name, state := range states {
 		if state.cmd == nil || state.cmd.Process == nil {
 			continue
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
@@ -635,12 +635,11 @@ func (r *LimaVMReconciler) shutdownAllHostagents() {
 
 	// Send graceful shutdown signal to all hostagents in parallel. Skip any whose
 	// process has already been reaped: once cmd.Wait() closes procExited it has
-	// released the OS process handle, freeing the PID for reuse. On Windows a
-	// CTRL_BREAK to a recycled PID escapes to the console and kills the service
-	// and terminal. While procExited is open the handle is normally still held,
-	// so the PID is still ours; cmd.Wait() releases it just before closing the
-	// channel, leaving a brief window.
-	for _, state := range states {
+	// released the OS process handle, freeing the PID for reuse. The named
+	// interrupt event keyed by instance and PID makes a recycled PID harmless (it
+	// has no such event, so Interrupt fails), but there is nothing to signal once
+	// the process is reaped.
+	for name, state := range states {
 		if state.cmd == nil || state.cmd.Process == nil {
 			continue
 		}
@@ -649,7 +648,7 @@ func (r *LimaVMReconciler) shutdownAllHostagents() {
 			continue
 		default:
 		}
-		_ = process.Interrupt(state.cmd.Process.Pid)
+		_ = process.Interrupt(process.HostagentInterruptKey(name), state.cmd.Process.Pid)
 	}
 
 	// Wait for each hostagent process to exit. The watcher goroutine may finish
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -634,7 +634,7 @@ func (r *LimaVMReconciler) killOrphanedHostagent(ctx context.Context, inst *lima
 	// confirms it is still our hostagent; stopInstanceForcibly re-screens before
 	// the forced stop below, because the PID can be recycled during the graceful
 	// wait.
-	if inst.HostAgentPID > 0 && process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
+	if process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
 		if err := process.Interrupt(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID); err != nil {
 			logger.V(1).Info("Could not signal orphaned hostagent", "pid", inst.HostAgentPID, "error", err)
 		} else {
@@ -675,17 +675,13 @@ func waitForInstanceStopped(ctx context.Context, name string) bool {
 
 // clearRecycledHostAgentPID zeroes inst.HostAgentPID unless it still names our
 // live hostagent, so a force-stop's taskkill cannot reach an unrelated process
-// that recycled the PID. The stored PID comes from on-disk state a previous
-// service wrote, which Windows may have reassigned after the hostagent exited;
-// on other platforms IsOurProcess is a no-op, so the PID is kept. DriverPID is
-// not screened here — IsOurProcess matches our registered interrupt event, which
-// the qemu/wsl driver never creates — so a recycled DriverPID is still
-// taskkilled by the caller.
+// that recycled the PID. DriverPID is not screened — the qemu/wsl driver
+// registers no interrupt event — so a recycled DriverPID is still taskkilled.
 //
 // TODO: track the hostagent and driver in a Windows Job Object so termination
-// no longer trusts a stored DriverPID that the OS may have recycled.
+// no longer trusts a stored DriverPID the OS may have recycled.
 func clearRecycledHostAgentPID(inst *limatype.Instance) {
-	if inst.HostAgentPID > 0 && !process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
+	if !process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
 		inst.HostAgentPID = 0
 	}
 }
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -634,8 +634,8 @@ func (r *LimaVMReconciler) killOrphanedHostagent(ctx context.Context, inst *lima
 	// confirms it is still our hostagent; stopInstanceForcibly re-screens before
 	// the forced stop below, because the PID can be recycled during the graceful
 	// wait.
-	if inst.HostAgentPID > 0 && process.IsOurProcess(inst.HostAgentPID, "hostagent", hostAgentPIDFile(inst)) {
-		if err := process.Interrupt(inst.HostAgentPID); err != nil {
+	if inst.HostAgentPID > 0 && process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
+		if err := process.Interrupt(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID); err != nil {
 			logger.V(1).Info("Could not signal orphaned hostagent", "pid", inst.HostAgentPID, "error", err)
 		} else {
 			stopCtx, cancel := context.WithTimeout(ctx, gracefulShutdownTimeout)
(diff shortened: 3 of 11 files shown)


jandubois:
I just realized that the event name should also include `RDD_INSTANCE`, but that is not important for Alpha 1.

rdd/pkg/service/cmd/service.go
```diff
@@ -558,6 +539,17 @@ func Delete(ctx context.Context, timeout time.Duration) error {
...
 		}
 	}
+
+	// Drop the datastore first, before anything destructive. It is the file a
+	// still-running control plane holds open; if RemoveAll fails because it is
+	// locked, a daemon is alive and we could not stop it, so abort before
+	// deleting the rest of the directory — in particular rdd.pid. A half-finished
+	// delete that strips rdd.pid would orphan that daemon (it becomes unfindable)
+	// and leave state.db behind: the exact failure this guards against.
+	if err := os.RemoveAll(filepath.Join(instance.Dir(), "db")); err != nil {
```
Nino-K: `os.RemoveAll` continues past errors, so if `state.db` is locked, `state.db-wal` and `state.db-shm` get deleted before the error is returned. Is that safe on Windows while the daemon is still using them?

Nino-K: I'm wondering if this should probe the lock first if the `state.db` is held open, the daemon is still running.

```go 
if err := os.Remove(filepath.Join(instance.Dir(), "db", "state.db")); err != nil && !os.IsNotExist(err){
    return fmt.Errorf("failed to remove datastore for %q; the control plane may still be running: %w", ...)
}
_ = os.RemoveAll(filepath.Join(instance.Dir(), "db"))

```

rdd/pkg/util/process/process_windows.go
```diff
@@ -10,246 +10,143 @@ import (
...
-	}
-	cmdline, err := processCommandLine(handle)
+// RegisterInterruptHandler creates this process's named interrupt event for key
+// and runs onInterrupt when another process signals it via Interrupt. Every RDD
+// process that can be the target of Interrupt — the control plane daemon
+// (ServeInterruptKey) and each hostagent (HostagentInterruptKey) — must call
+// this once at startup with its own key; the returned function stops the watcher
+// and releases the event. On Unix it is a no-op: Interrupt there sends SIGINT,
+// which the process already handles via signal.Notify / SetupSignalContext.
+func RegisterInterruptHandler(key string, onInterrupt func()) (func(), error) {
```
Nino-K, line 77: If this fails, the daemon starts without an interrupt event. `PID()` then calls `IsOurProcess`, gets `false`, and returns `PIDNotFound`, so `rdd svc stop` reports "not running". Also, `rdd svc start` sees the same thing and tries to spawn a second daemon?

rdd/pkg/util/process/process_unix.go
```diff
@@ -22,32 +22,46 @@ func SetGroup(cmd *exec.Cmd) {
...
+// signal.Notify / genericapiserver.SetupSignalContext, so no extra wiring is
+// needed. It exists for signature parity with the Windows build, where Interrupt
+// signals a named event the target must create at startup.
+func RegisterInterruptHandler(_ string, _ func()) (func(), error) {
+	return func() {}, nil
+}
+
+// IsOurProcess reports whether pid is a live process we may signal under key. On
+// Unix it is always true: a deliberate no-op, not a guard, so the parameters go
+// unused. They exist for signature parity with the Windows build, where the
```
mook-as:
```suggestion
// IsOurProcess reports whether pid is a live process we may signal under key. On
// Unix it is always true. They exist for signature parity with the Windows build, where the
```
We don't really need that second part, it just makes things harder to read.

rdd/pkg/util/process/process_unix.go
```diff
@@ -22,32 +22,46 @@ func SetGroup(cmd *exec.Cmd) {
...
+// check confirms a PID belongs to one of our processes (by named-event
+// registration) before a caller acts on it. Unix needs no such confirmation:
+// Interrupt/Kill use signals, and PID reuse is far less aggressive than on
+// Windows.
```
mook-as:
```suggestion
// registration) before a caller acts on it. Unix needs no such confirmation; PID reuse is far less aggressive than on
// Windows.
```
The _foo: bar baz quux_ form when the text in front of a colon is a proper clause is hard to read; please avoid. The _Residual risk: …_ sentence below is better because that's a topic comment.

rdd/pkg/service/cmd/service.go
jandubois: Yes, good catch!

rdd/pkg/util/process/process_windows.go
jandubois: I made failure to register an interrupt handler an error now.

rdd/pkg/util/process/process_windows.go
```diff
@@ -10,246 +10,145 @@ import (
...
-// Interrupt sends a graceful shutdown signal to the process with the given PID.
-// On Windows, this uses GenerateConsoleCtrlEvent with CTRL_BREAK_EVENT targeted
-// at the process group (requires CREATE_NEW_PROCESS_GROUP via SetGroup).
-func Interrupt(pid int) error {
-	return windows.GenerateConsoleCtrlEvent(windows.CTRL_BREAK_EVENT, uint32(pid))
+// interruptEventName returns the name of the event Interrupt signals to ask the
+// process identified by (key, pid) to shut down gracefully. The Local\ namespace
+// scopes the event to the current session, which is where the control plane and
+// the CLI run (same user, interactive session); a daemon running as a session-0
+// Windows service would need the Global\ namespace. instance.Suffix() namespaces
```
mook-as: We don't support running as a service (yet), so we can drop the bit about `Global\`. A four-line sentence is definitely on the long side.

rdd/pkg/util/process/process_windows.go
```diff
@@ -10,246 +10,145 @@ import (
...
-	self, err := os.Executable()
-	if err != nil {
-		return "", false
-	}
-	handle, err := windows.OpenProcess(
-		windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, uint32(pid))
+// It is also inherently safe against PID reuse: only an RDD process that called
+// RegisterInterruptHandler with this key creates the event, so OpenEvent fails
+// for a recycled or unrelated PID and Interrupt returns an error rather than
+// disturbing it. Callers fall back to Kill/KillTree on error.
```
mook-as: So if the pid was reused, we use `Kill`/`KillTree` on the unrelated PID? That sounds weird.

rdd/pkg/util/process/process_windows.go
```diff
@@ -10,246 +10,145 @@ import (
...
-	return windows.GenerateConsoleCtrlEvent(windows.CTRL_BREAK_EVENT, uint32(pid))
+// interruptEventName returns the name of the event Interrupt signals to ask the
+// process identified by (key, pid) to shut down gracefully. The Local\ namespace
+// scopes the event to the current session, which is where the control plane and
+// the CLI run (same user, interactive session); a daemon running as a session-0
+// Windows service would need the Global\ namespace. instance.Suffix() namespaces
+// the event by rdd instance and key by role (see ServeInterruptKey /
+// HostagentInterruptKey), so a PID recycled to a different instance or role
+// creates no matching event and IsOurProcess can confirm both.
+func interruptEventName(key string, pid int) string {
```
mook-as, line 40: Bundling `windows.UTF16PtrFromString` here seems to make more sense (because the event name is passed to the event APIs)? Totally optional though.

rdd/pkg/util/process/process_windows.go
```diff
@@ -10,246 +10,145 @@ import (
...
-// comparison rather than a guaranteed mismatch — for the service PID a false
-// "not ours" would orphan a live control plane.
-func longPath(p string) string {
-	if resolved, err := resolveLongPath(p); err == nil {
-		return filepath.Clean(resolved)
+	// Manual-reset so the signal latches: the watcher cannot miss a caller's
+	// SetEvent by racing it, and a later look still sees the event set.
+	evt, err := windows.CreateEvent(nil, 1, 0, name)
+	if err != nil && !errors.Is(err, windows.ERROR_ALREADY_EXISTS) {
+		return nil, fmt.Errorf("create interrupt event: %w", err)
```
mook-as: I don't see how `ERROR_ALREADY_EXISTS` can be valid (the event would have to be created by the same process‽)

rdd/pkg/service/cmd/service.go
```diff
@@ -140,57 +138,38 @@ func Exists() bool {
...
+	// Self-heal a genuinely stale file: if no process exists under the recorded
+	// PID, remove it. If a process is alive but not our control plane — a
+	// recycled PID, or (on Windows) a daemon that never registered its interrupt
+	// event — leave the file: it is never acted on while unconfirmed and the next
+	// Start overwrites it, whereas deleting it could strip a live daemon's record
+	// and orphan it (the failure that left state.db locked and undeletable).
+	if !process.IsAlive(pid) {
+		_ = os.Remove(instance.PIDFile())
+		return PIDNotFound
+	}
```
mook-as: Why is this needed at all? Can't we just call `IsOurProcess` directly?  A live process that _isn't_ ours isn't of any use to us.  (This means making `IsOurProcess` actually do enough work on Windows too.)

Also, the `!IsOurProcess` block below should also delete the PID file.

rdd/pkg/util/process/process_windows_test.go
```diff
@@ -6,163 +6,52 @@ package process
...
-}
-
-// TestIsOurProcess confirms identity matching: the running process matches its
-// own executable and command line, and rejects absent substrings and dead PIDs.
-func TestIsOurProcess(t *testing.T) {
-	pid := os.Getpid()
-	base := filepath.Base(os.Args[0])
+	assert.Assert(t, IsOurProcess(key, os.Getpid()),
+		"a process that registered key %q should be recognised as ours", key)
+	assert.Assert(t, !IsOurProcess(HostagentInterruptKey("rd"), os.Getpid()),
```
mook-as: Why not use a definitely invalid key (`!IsOurProcess("invalid-key", os.Getpid())`)?

rdd/pkg/util/process/process_windows_test.go
```diff
@@ -6,163 +6,52 @@ package process
...
-	t.Helper()
-	handle, err := windows.OpenProcess(
-		windows.PROCESS_QUERY_LIMITED_INFORMATION|windows.PROCESS_VM_READ, false, uint32(os.Getpid()))
+// TestRegisterInterruptHandlerAndInterrupt confirms the named-event round trip:
+// after a process registers a handler for a key, Interrupt(key, pid) fires it
+// and IsOurProcess(key, pid) recognises it. This is the mechanism that delivers
+// a graceful shutdown across consoles.
+func TestRegisterInterruptHandlerAndInterrupt(t *testing.T) {
+	const key = ServeInterruptKey
+	fired := make(chan struct{})
```
mook-as: Seems like using a `atomic.Bool` here instead of a channel would be clearer.

rdd/pkg/util/process/process_windows_test.go
```diff
@@ -6,163 +6,52 @@ package process
...
-	if strings.EqualFold(filepath.Clean(short), filepath.Clean(longFile)) {
-		t.Skip("8.3 short-name generation is disabled on this volume")
-	}
-	assert.Assert(t, samePath(short, longFile),
-		"short form %q and long form %q must resolve to the same executable", short, longFile)
+// TestInterruptUnregisteredFails confirms Interrupt and IsOurProcess reject a PID
+// with no registered event — a dead, recycled, or unrelated process — so callers
+// fall back to a force kill instead of disturbing it.
+func TestInterruptUnregisteredFails(t *testing.T) {
+	assert.Assert(t, Interrupt(ServeInterruptKey, 0x7FFFFFF0) != nil,
```
mook-as: FWIW, the correct "invalid" PID is 0. Or possibly `os.Getpid() + 1`, because PIDs are apparently 4-bit aligned on Windows.

rdd/pkg/util/process/keys_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// TestInterruptKeys pins the interrupt-key format. Both ends of an interrupt —
+// the process that registers (the hostagent / daemon) and the one that signals
+// or checks it (the lima controller / CLI) — must derive the same key, so the
+// format is part of the contract.
+func TestInterruptKeys(t *testing.T) {
```
mook-as, line 17: I'm not sure this test is actually useful, but it probably doesn't hurt?

rdd/pkg/service/cmd/service.go
```diff
@@ -140,57 +138,38 @@ func Exists() bool {
 const PIDNotFound = 0
 
 // PID returns the process ID of the running service, or PIDNotFound if it is
-// not running. On Windows it walks the candidate process's PEB to read its
-// command line (see process.IsOurProcessWithArg), so callers that poll it in a
-// tight loop pay a cross-process read each call.
+// not running. On Windows it confirms the PID via the daemon's interrupt event
+// (process.IsOurProcess), a cheap OpenEvent that callers may poll in a loop.
```
mook-as: Nothing in `OpenEvent`'s documentation indicates that it might be cheap…

rdd/pkg/service/cmd/service.go
```diff
@@ -466,28 +445,30 @@ func Wait(ctx context.Context) error {
 // timeout 0 to wait indefinitely (bounded only by ctx).
```
mook-as, line 432: Shouldn't `ctx` have the timeout built in, instead of passing the timeout in manually?

rdd/pkg/service/cmd/service.go
```diff
@@ -466,28 +445,30 @@ func Wait(ctx context.Context) error {
...
 
-	// Try graceful shutdown first. On Unix, Kill already sends SIGTERM which
-	// triggers the Go signal handler. On Windows, Kill uses TerminateProcess
-	// which bypasses all handlers, so we send Interrupt (CTRL_BREAK_EVENT)
-	// first to let the service run its shutdown path (shutdownAllHostagents).
+	// Try graceful shutdown first. Interrupt asks the daemon to run its orderly
+	// shutdown (drain the controller manager, which stops the hostagents): on
+	// Unix via SIGINT, on Windows by signalling the daemon's named interrupt
+	// event (process.RegisterInterruptHandler), which reaches the daemon even
+	// when the CLI runs in a different console, e.g. a daemon the GUI app started.
```
mook-as: `Interrupt`'s implementation details doesn't go here; everything after the colon is excessive detail.

rdd/pkg/service/cmd/service.go
```diff
@@ -466,28 +445,30 @@ func Wait(ctx context.Context) error {
...
-	// Hostagents survive in their own process groups and self-heal on the
-	// next service start via killOrphanedHostagent.
-	if err := process.Interrupt(pid); err != nil {
+	// Kill (SIGTERM on Unix, TerminateProcess on Windows) is the fallback when
+	// Interrupt fails. On Unix it rarely does, since SIGINT succeeds for any live
+	// PID. On Windows a failure means the target has no interrupt event — it is
+	// gone, recycled, or not ours — so terminating its single PID cannot reach an
+	// unrelated process. A hard kill skips the controller-manager drain, so
+	// hostagents survive in their own process groups and self-heal on the next
+	// service start via killOrphanedHostagent.
```
mook-as: Also too much detail here; that's relevant for `Kill`'s documentation.

rdd/pkg/service/cmd/service.go
```diff
@@ -558,6 +539,20 @@ func Delete(ctx context.Context, timeout time.Duration) error {
 			}
 		}
 	}
+
+	// Probe the datastore lock before anything destructive. state.db is the file a
+	// still-running control plane holds open; remove it alone first, so a lock
+	// error aborts the delete before os.RemoveAll could strip the -wal/-shm
+	// sidecars or rdd.pid out from under a live daemon and orphan it — the exact
+	// failure this guards against. (os.RemoveAll records the first error but keeps
```
mook-as:
```suggestion
	// sidecars or rdd.pid out from under a live daemon and orphan it.
	// (os.RemoveAll records the first error but keeps
```
Excessive verbiage.

Of course `os.RemoveAll` cannot be a probe, the order it deletes files is undefined; drop the parenthesized expression please.

This whole chunk should be done with an explicit, intentional guard: the process should `flock()` the PID file, so we can just delete _that_ as the test.

rdd/pkg/service/cmd/service.go
```diff
@@ -708,6 +703,20 @@ func NewServeCommand(ctx context.Context) *cobra.Command {
 				}
 			}
 
+			// Register the interrupt event before writing the PID file so a peer can
+			// confirm our identity (IsOurProcess) as soon as the file exists. On
```
mook-as:
```suggestion
			// confirm our identity (`process.IsOurProcess`) as soon as the file exists. On
```

Though I think if we do it as `[process.IsOurProcess]` some IDEs would actually link it, so maybe that is better.

rdd/pkg/controllers/lima/limavm/controllers/hostagent_watcher.go
```diff
@@ -209,17 +209,16 @@ func (r *LimaVMReconciler) signalHostagent(name string) bool {
...
 	}
-	// Once procExited is closed, cmd.Wait() has released the OS process handle
-	// and the PID may be reused; signalling it could deliver CTRL_BREAK to an
-	// unrelated console process on Windows. While it is open the handle is
-	// normally still held; cmd.Wait() releases it just before closing the
-	// channel, leaving a brief window.
+	// Skip a process that has already been reaped: once procExited is closed,
+	// cmd.Wait() has released the OS handle and the PID may be reused. The named
+	// interrupt event makes a reused PID harmless — a recycled process has no such
+	// event, so Interrupt simply fails — but there is nothing to signal here.
```
mook-as: The _named interrupt_ bit only applies to Windows, right? We should probably just skip the platform-specific details in cross-platform code.

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -635,12 +635,11 @@ func (r *LimaVMReconciler) shutdownAllHostagents() {
...
-	// released the OS process handle, freeing the PID for reuse. On Windows a
-	// CTRL_BREAK to a recycled PID escapes to the console and kills the service
-	// and terminal. While procExited is open the handle is normally still held,
-	// so the PID is still ours; cmd.Wait() releases it just before closing the
-	// channel, leaving a brief window.
-	for _, state := range states {
+	// released the OS process handle, freeing the PID for reuse. The named
+	// interrupt event keyed by instance and PID makes a recycled PID harmless (it
+	// has no such event, so Interrupt fails), but there is nothing to signal once
+	// the process is reaped.
```
mook-as: Same as above; we should probably avoid planting comments about implementation details everywhere.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -634,8 +634,8 @@ func (r *LimaVMReconciler) killOrphanedHostagent(ctx context.Context, inst *lima
 	// confirms it is still our hostagent; stopInstanceForcibly re-screens before
 	// the forced stop below, because the PID can be recycled during the graceful
 	// wait.
-	if inst.HostAgentPID > 0 && process.IsOurProcess(inst.HostAgentPID, "hostagent", hostAgentPIDFile(inst)) {
-		if err := process.Interrupt(inst.HostAgentPID); err != nil {
+	if inst.HostAgentPID > 0 && process.IsOurProcess(process.HostagentInterruptKey(inst.Name), inst.HostAgentPID) {
```
mook-as: Do we need the `inst.HostAgentPID > 0` check? PID 0 can never be our process.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -673,27 +673,19 @@ func waitForInstanceStopped(ctx context.Context, name string) bool {
...
 // clearRecycledHostAgentPID zeroes inst.HostAgentPID unless it still names our
 // live hostagent, so a force-stop's taskkill cannot reach an unrelated process
 // that recycled the PID. The stored PID comes from on-disk state a previous
 // service wrote, which Windows may have reassigned after the hostagent exited;
 // on other platforms IsOurProcess is a no-op, so the PID is kept. DriverPID is
-// not screened here — IsOurProcess matches the rdd image, not the qemu/wsl
-// driver — so a recycled DriverPID is still taskkilled by the caller.
+// not screened here — IsOurProcess matches our registered interrupt event, which
+// the qemu/wsl driver never creates — so a recycled DriverPID is still
+// taskkilled by the caller.
```
mook-as: Again, too much remote implementation detail in comments.

mook-as (commented):
Post-alpha, I want to rewrite this on Windows to use named file mappings backed by paging files, so it becomes a PID file that automatically goes away when the process exits.

I still hate all your extra sentences with colons in them. I'm not sure if it's because I'm noticing them more, though.

Marking as a comment instead of request changes because I see that Nino has already submitted a review.

Nino-K (approved):
LGTM

PR #491, opened 2026-06-25
Fix smoke tests on Windows

mook-as:
- Fix not erroring out when the checksum is invalid
- Fix the signing config, so we can verify signatures correctly.

With this the smoke tests for v2.0.0-alpha.1 should pass (without modifying the uploaded bits).


jandubois (approved):
Thanks, LGTM

PR #492, opened 2026-06-25
CI: smoke tests: Make download more granular

mook-as:
The repository smoke tests do not require any of the downloaded artifacts to run.  Additionally, the AppImage download is often slower, and is more likely to have issues.  Split that out from the download of the various installers / packages as well, so if the download fails the other tests could still start, and retrying the download is easier.

Sample test run: https://github.com/mook-as/rd2/actions/runs/28197415284
(Note that there's a trivial string change since that commit.)

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -29,7 +29,7 @@ on:
         default: false
 
 jobs:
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #493, opened 2026-06-25
Striped icon and nav for pre-release builds

jandubois:
Gives alpha/beta builds a distinctive striped look so they are not mistaken for stable.

## Changes
- **App icon** — diagonal stripes over the brand blue, on both masters: `logo-square-512.png` (Linux + the generated Windows `.ico`) and `mac-icon.png` (macOS). The striped icons keep the default names, so every build- and run-time reference picks them up with no wiring. The stable masters move to `*-release.png` and are excluded from the packaged app.
- **In-app nav** — the selected item in the left-hand nav (`Nav.vue`) and the Preferences dialog (`ModalNavItem.vue`) shows the same stripes, driven by a shared `--nav-active-stripes` variable. The selected label turns white to stay legible on the blue, including the light theme where it was dark on grey.

## Notes
- Unconditional by design: no prerelease detection, since the version rules treat alpha/beta as releases. The look is baked into this branch.
- At GA, undo it: copy the `*-release.png` icons back over the defaults and revert the nav commit.


jandubois:
Verified on macOS only

mook-as (changes requested):
Making this permanent doesn't sound great; we should probably get something to figure out if we're doing a non-release build, and swap in the assets.  Something like:

```diff
diff --git scripts/lib/__tests__/build-utils.spec.ts scripts/lib/__tests__/build-utils.spec.ts
index a745d48bb..42705b920 100644
--- scripts/lib/__tests__/build-utils.spec.ts
+++ scripts/lib/__tests__/build-utils.spec.ts
@@ -3,7 +3,7 @@ import childProcess from 'node:child_process';
 import { jest } from '@jest/globals';
 import semver from 'semver';
 
-import buildUtils from '../build-utils';
+import buildUtils, { _clearCachedVersion } from '../build-utils';
 
 describe('build-utils', () => {
   describe('docsUrl', () => {
@@ -65,4 +65,24 @@ describe('build-utils', () => {
       expect(version).toBe('0.0.0-fallback');
     });
   });
+
+  describe('isReleaseVersion', () => {
+    afterEach(() => {
+      delete process.env.RD_MOCK_VERSION;
+    });
+
+    test.each`
+      expected   | version
+      ${ false } | ${ '1.9.0-tech-preview' }
+      ${ false } | ${ '2.0.0-alpha.1' }
+      ${ false } | ${ '2.0.0-alpha.1-9-g0106d8077' }
+      ${ true }  | ${ '2.0.0' }
+      ${ false } | ${ '2.0.0-9-g0106d8077' }
+      ${ false } | ${ 'invalidVersion' }
+    `('version $version should return $expected', async({ version, expected }) => {
+      process.env.RD_MOCK_VERSION = version;
+      _clearCachedVersion();
+      await expect(buildUtils.isReleaseVersion).resolves.toBe(expected);
+    });
+  });
 });
diff --git scripts/lib/build-utils.ts scripts/lib/build-utils.ts
index 2a1530ab9..ce21b70fd 100644
--- scripts/lib/build-utils.ts
+++ scripts/lib/build-utils.ts
@@ -23,6 +23,14 @@ type SpawnResult = Promise<void> & {
 
 let cachedVersion: Promise<string> | undefined;
 
+/**
+ * For testing use only: clear the cached version so that the next call to
+ * `version` will recompute it.
+ */
+export function _clearCachedVersion() {
+  cachedVersion = undefined;
+}
+
 export default {
   /**
    * Determine if we are building for a development build.
@@ -108,6 +116,17 @@ export default {
     return cachedVersion;
   },
 
+  /**
+   * Get whether this is a release version.
+   * @note This may be true due to a mock version; this is required to generate
+   * the correct screenshots.
+   */
+  get isReleaseVersion(): Promise<boolean> {
+    return (async() => {
+      return semver.prerelease(await this.version) === null;
+    })();
+  },
+
   get docsUrl(): Promise<string> {
     return (async() => {
       const baseUrl = 'https://docs.rancherdesktop.io';
```

jandubois:
> we should probably get something to figure out if we're doing a non-release build, and swap in the assets.

I don't think this is necessary; we will just revert the PR once we release 2.0.0 final and be done with it. If you want this change, please push directly to the PR branch!

pkg/rancher-desktop/assets/styles/themes/_light.scss
```diff
@@ -542,3 +542,12 @@ $selected: rgba(#3D98D3, .5);
   --button-icon                : #{$dark};
   --button-icon-bg             : #{$secondary};
 }
+
+// Pre-release builds (development, alpha, beta) stripe the selected nav item to
+// match the striped app icon; entry/index.ts adds this class.  The custom
+// properties cascade into the scoped nav components, so one definition covers
+// every theme.
+BODY.prerelease {
```
mook-as: I think it would be clearer to put this in a separate file (e.g. `prerelease.scss`), that is included via `app.scss`.

pkg/rancher-desktop/entry/index.ts
```diff
@@ -21,6 +23,12 @@ const component = matched?.components?.default;
 const layoutName: string = (component as any)?.layout ?? 'default';
 const { default: layout } = await import(`../layouts/${ layoutName }.vue`);
 
+// Pre-release builds (development, alpha, beta) carry the striped app-icon look
+// into the app windows; Nav.vue and ModalNavItem.vue key off this class.
+if (!isReleaseVersion(process.env.RD_VERSION ?? '')) {
+  document.body.classList.add('prerelease');
+}
+
```
mook-as, line 29: We should be defining things at build time, and not depend on doing work at runtime that can be done earlier; see things like https://github.com/rancher-sandbox/rancher-desktop-2/blob/21d720bb97cf9380d53247fcdbcc719f6b9d536d/scripts/lib/build-utils.ts#L266-L269

So in the front end code, it should be just `if (process.env.RD_PRERELEASE) { … }` where that's a constant set via `DefinePlugin`.

scripts/lib/build-utils.ts
```diff
@@ -3,6 +3,7 @@
  */
 
 import childProcess from 'node:child_process';
+import fs from 'node:fs';
```
mook-as, line 6: 👍 I want to change all stdlib imports to `node:` slowly, to be clear that it's not a thing that can be overridden by a confusing `yarn add …`.

scripts/package.ts
```diff
@@ -127,6 +127,8 @@ class Builder {
   async package(): Promise<CliOptions> {
     log.info('Packaging...');
 
+    await buildUtils.prepareIcons();
```
mook-as: Conceptually, this should live in `build.ts` inside of `buildRenderer`, I think?

scripts/dev.ts
```diff
@@ -225,6 +225,7 @@ class DevRunner extends events.EventEmitter {
   async run() {
     process.env.NODE_ENV = 'development';
     try {
+      await buildUtils.prepareIcons();
```
mook-as: Should this call be _inside_ `startRendererProcess`?

scripts/postinstall.ts
```diff
@@ -179,6 +179,8 @@ async function runScripts(): Promise<void> {
   }
 
   await downloadDependencies(dependencies);
+
+  await buildUtils.prepareIcons();
```
mook-as: Should this be done as a "dependency"? Similar to how we do, say, `sudo-prompt`.

Speaking of; this needs to happen _before_ the `sudo-prompt` build, because that needs the icon to be embedded into the mini-app. Not that we use that at at the moment.  But that's more a reason to be a dependency (so `sudo-prompt` can mark it as a dependency).

pkg/rancher-desktop/assets/styles/themes/_light.scss
```diff
@@ -542,3 +542,12 @@ $selected: rgba(#3D98D3, .5);
...
+  --nav-active-bg              : repeating-linear-gradient(-45deg, #0081be 0, #0081be 12px, #72b9db 12px, #72b9db 20px);
```
mook-as: Should these be calculated from the various colours (e.g. `--nav-active`)? Because I see that we don't have an override for dark mode, but that has a different `--nav-active`.

mook-as (changes requested):
I haven't actually ran it; I assume it's to your satisfaction.

scripts/postinstall.ts
jandubois: Doesn't really make sense given that `sudo-prompt` isn't a dependency anymore. Sounds like a lot of scaffolding for no benefit.

pkg/rancher-desktop/assets/styles/themes/_light.scss
jandubois: I don't know. The light mode nav bar is just a light gray. Using light gray with white stripes doesn't look good to me; too low contrast. Also blends into the background colour.

I think the current stripes work fine in both dark and light mode. They are showing that it is a beta, so it is supposed to look unfinished (but not bad):

CleanShot 2026-07-02 at 21 04 38@2x

CleanShot 2026-07-02 at 21 05 05@2x

Sorry, I just noticed that I have the non-active windows dimmed, so the colours don't quite match the active window, but you should see the idea.

jandubois: Redid the light screenshot without dimming. Still shows that the light navbar is not a good candidate for striping.

CleanShot 2026-07-02 at 21 11 26@2x

pkg/rancher-desktop/vue.config.mjs
```diff
@@ -1,10 +1,18 @@
...
+import semver from 'semver';
 import webpack from 'webpack';
 
 const rootDir = path.resolve(import.meta.dirname, '..', '..');
 
+const rdVersion = process.env.RD_VERSION || '0.0.1-fallback';
+
+// A pre-release build — development, alpha, beta, or an unparsable version —
+// gets the striped nav.  This duplicates isReleaseVersion() from
+// @pkg/utils/version, which a plain-JavaScript config cannot import.
```
mook-as: Maybe it would be better to write `@pkg/utils/version` as JS (not TS), with a comment as to why, plus typing comments to make it work well in TypeScript.

scripts/postinstall.ts
mook-as: The benefit is that it wouldn't be _special_: everything would run through the (increasingly misnamed — we already have things that are not downloaded) dependency system.

mook-as (changes requested):
Image

Image

Maybe my eyes are worse than yours, but I find the word _General_ very hard to read; we should probably reduce the contrast in the background so the words can show up better.

My comment about calculating colours is just thinking that we'd need a different value in dark mode; I don't actually care if `--nav-active` itself is involved.  This is the actual reason I think the PR needs work; the comments belong are not as important.

pkg/rancher-desktop/vue.config.mjs
jandubois: Done

jandubois:
Removed `buildUtils.prepareIcons();` from `postinstall`; it is no longer needed now since the icon is always be prepared on each code path.

jandubois:
This is how it looks now on macOS:

CleanShot 2026-07-17 at 10 56 59@2x
CleanShot 2026-07-17 at 10 56 49@2x

mook-as (approved):

PR #494, opened 2026-06-26
Download guest distro as an rddepman dependency

jandubois:
First step toward embedding the guest distro in the rdd binary: add the openSUSE distro as an `rddepman` dependency so `yarn install` fetches it.

The download stages the compressed image at the gitignored go:embed path (`rdd/pkg/embedded/distro.img.xz`); this PoC keeps download = embed. A later step adds the overlay and the code that consumes the blob.


jandubois:
This is the wrong approach; we need to treat the distro and guest binaries differently from bundled host utilities.

Issue #496, opened 2026-06-26
Create separate `rdd/dependencies.yaml` for guest dependencies


jandubois:
We want to track updates to the distro images, as well as guest binaries like `nerdctl` and `openresty` via `rddepman`.

But we need download these dependencies as part of `make build-rdd` and not rely on `yarn install` to fetch them during post-install. Because during post-install it already wants to copy the fully-built `rdd` binary into the resources directory, together with the downloaded host utilities.

We should have a separate `rdd/dependencies.yaml` file for guest dependencies. It will need additional metadata, like download URLs, file formats, and the guest arch.

That way a simplistic downloader in Go can fetch the files and confirm the checksums. Any further post-processing to extract files and move them into the distro overlay will happen by the `Makefile`.

jandubois:
Closed by #511

Issue #449, opened 2026-06-13
Bundle `opensuse` distro with `rdd`


jandubois:

mook-as:
I think the order is backwards: we need to bundle the distro (as-is) with rdd, _then_ create the overlay tool. Otherwise the overlay tool output can't be used.

jandubois:
> we need to bundle the distro (as-is) with rdd,

We can still release the customized distro as part of the release artifacts instead of bundling it in the binary. Release bits would be

* rdd binaries
* RD2 app installers
* distro tarball and image

The `lima.yaml` would download from `rancher-desktop-2` releases instead of `rancher-desktop-suse` release. One snag is that we need to have reproducible builds of the custom distro for this to work because we have to embed the digest of the distro in the lima template.

If we bundle the distro inside `rdd` itself, then we don't need to care about digests. Having reproducible bits would still be nice.

mook-as:
I'm trying to figure out the flow to make a change to the (base) image.
- The `…-opensuse` repository changes.
- That makes a release in the `…-opensuse` repo.
- We update some sort of manifest in the rd2 repo.
- We push the commit to GitHub, so it can build a new disk image.
- We then make another commit to the rd2 repo, referencing the new published disk image.

That… seems kind of awkward.

jandubois:
> I'm trying to figure out the flow to make a change to the (base) image.
> 
> * The `…-opensuse` repository changes.
> * That makes a release in the `…-opensuse` repo.

We make a release by creating a tag, just like we did with `alpine-lima` and `wsl-distro`.

`rddepman` will detect the new release and update the dependencies in `rancher-desktop-2`.

> * We update some sort of manifest in the rd2 repo.

We only need to update the manifest when we change what we bundle as the overlay.
  
> * We push the commit to GitHub, so it can build a new disk image.
  
The disk image is build locally as part of the `postinstall` step.
  
> * We then make another commit to the rd2 repo, referencing the new published disk image.

I think I see your point now. We need a way to reference the local image without creating a new release.

So you were right, we need to bundle the distro first.

PR #511, opened 2026-06-26
rddepman: add the guest manifest

jandubois:
The schema half of this PR went to rancher-desktop first, as mook suggested, and came back through the July merge; this branch is rebased onto that, so only the guest side is left here.

Implements #496, part of #449.


mook-as:
Consider doing this in the RD1 repo (minus the guest stuff), so we can do merges in the future.  I'll do a review either way.

scripts/lib/dependencies.ts
```diff
@@ -77,58 +87,118 @@ export function parseSha256Checksum(value: unknown): Sha256Checksum {
...
+  arch?:    GoArch;
+  /** The fully-resolved download URL. */
+  url:      string;
+  /** The sha256 of the downloaded bytes, `sha256:`-prefixed. */
+  checksum: Sha256Checksum;
+  /**
+   * Distribution format, when one package ships several for the same
+   * platform/arch (the guest distro's `raw` ext4 image vs `tar` rootfs).
+   */
+  format?:  string;
```
mook-as: I think this is better called `variant`, because there may be non-format differences for other assets. For example, if some time down the road we supported Leap vs SLES.

scripts/lib/dependencies.ts
```diff
@@ -77,58 +87,118 @@ export function parseSha256Checksum(value: unknown): Sha256Checksum {
...
+ * One downloadable artifact for a dependency.  Every field needed to fetch and
+ * verify the bytes lives here, so a downloader needs no per-package knowledge.
+ * It selects by {@link platform}/{@link arch} (and {@link format} where one
+ * package ships several), fetches {@link url}, and checks the bytes against
+ * {@link checksum}.
  */
-export interface DependencyEntry<K extends keyof DependencyVersions = keyof DependencyVersions> {
-  version:   DependencyVersions[K];
-  checksums: Record<string, Sha256Checksum>;
+export interface DependencyAsset {
```
mook-as, line 80: I guess this structure is designing mirroring Lima's image download stuff? Because using the platform/arch as the key to a map instead of iterating through everything seems easier, but it's not as expandable (format/variant). So I guess this is fine.

scripts/lib/dependencies.ts
```diff
@@ -77,58 +87,118 @@ export function parseSha256Checksum(value: unknown): Sha256Checksum {
...
+}
 
 interface RawEntry {
-  version:    unknown;
-  checksums?: Record<string, unknown>;
+  version: unknown;
+  assets?: unknown;
+}
+
+const ASSET_PLATFORMS: readonly AssetPlatform[] = ['linux', 'darwin', 'windows', 'wsl'];
```
mook-as: It would be good to ensure that we covered all possible platforms here; it looks like that can be done via:

```ts
const ASSET_PLATFORMS = ['linux', 'darwin', 'windows', 'wsl'] satisfies readonly AssetPlatform[];
const _assertPlatformCoverage: AssetPlatform extends (typeof ASSET_PLATFORMS)[number] ? true : never = true;
```

The `const ASSET_PLATFORMS` line ensures that all the elements are a valid `AssetPlatform` (via `satisfies`), but keeps its type as the literal strings.
The `const _assert…` line then checks that all possible values of `AssertPlatform` are a value in `ASSERT_PLATFORMS`, and on failure, it declares the type of the constant to be `never`; we then assign `true` to that, so if it fails we boil down to `const foo: never = true;` which is invalid.

The same can be done for `ASSET_ARCHES`.

---

Actually, since you have `if (!ASSET_PLATFORMS.includes(raw.platform as AssetPlatform))` below, just the first line (converting the definition to `satisifies`) is enough; if `ASSET_PLATFORMS` is missing values in `AssetPlatform`, that becomes `[A].includes([A, B])` and it fails type checking.

scripts/lib/dependencies.ts
```diff
@@ -77,58 +87,118 @@ export function parseSha256Checksum(value: unknown): Sha256Checksum {
...
+
+/** Parses and validates a single asset from a manifest entry. */
+function parseAsset(name: string, path: string, raw: RawAsset): DependencyAsset {
+  if (!ASSET_PLATFORMS.includes(raw.platform as AssetPlatform)) {
+    throw new Error(`Asset for ${ name } in ${ path } has invalid platform ${ JSON.stringify(raw.platform) }`);
+  }
+  if (raw.arch !== undefined && !ASSET_ARCHES.includes(raw.arch as GoArch)) {
+    throw new Error(`Asset for ${ name } in ${ path } has invalid arch ${ JSON.stringify(raw.arch) }`);
+  }
+  if (typeof raw.url !== 'string') {
```
mook-as:
```suggestion
  if (typeof raw.url !== 'string' || !URL.canParse(raw.url)) {
```
That seems to work even for Windows `file:` URLs on a mac.

scripts/lib/dependencies.ts
```diff
@@ -138,26 +208,69 @@ const MANIFEST_HEADER = [
...
+ * drops any other comments on the next rddepman bump.
+ */
+export async function writeDependencyManifest(path: string, manifest: DependencyManifest): Promise<void> {
+  const serializable = Object.fromEntries(Object.entries(manifest).map(([name, entry]) => [name, {
+    version: entry.version,
+    assets:  entry.assets.map(serializeAsset),
+  }]));
+
+  await fs.promises.writeFile(path, MANIFEST_HEADER + YAML.stringify(serializable), { encoding: 'utf-8' });
+  manifestCache.delete(path);
```
mook-as, line 240: Consider ordering the delete above the write, so a partial write would not result in a stale cache (yes, I know it would throw and everything)

scripts/lib/dependencies.ts
```diff
@@ -138,26 +208,69 @@ const MANIFEST_HEADER = [
...
+
+/**
+ * Writes the manifest to disk and invalidates the cached read for that path so
+ * subsequent reads observe the new contents.  Always emits a leading header
+ * comment so contributors editing the YAML directly see the warning where they
+ * are editing; everything else round-trips through `YAML.stringify`, which
+ * drops any other comments on the next rddepman bump.
+ */
+export async function writeDependencyManifest(path: string, manifest: DependencyManifest): Promise<void> {
+  const serializable = Object.fromEntries(Object.entries(manifest).map(([name, entry]) => [name, {
```
mook-as: Isn't `DependencyManifest` already meant to be the on-disk format? Why is `serializable` needed? To drop the undefined value? Both `JSON.stringify` and `YAML.stringify` seem to do that already.

Tested with this code:
```js
import YAML from "https://esm.sh/yaml";

const input = {
  a: undefined,
  b: 3,
  nested: {
    value: true,
    missing: undefined,
  }
}
document.body.append(YAML.stringify(input));
document.body.append(JSON.stringify(input));
```
Output:
```
b: 3
nested:
  value: true
{"b":3,"nested":{"value":true}}
```

Of course, if we did extra things like sorting, that might make it worthwhile.

jandubois:
Switched to draft mode. Commit 1 should be ported to RD1 and then merged back here.

Commit 2 depends on commit 1 and will have to wait.

scripts/lib/dependencies.ts
mook-as:
```suggestion
    versions[name] = manifest[name].version;
```

Since you didn't actually change this function, this is optional; I just happened to be reading the file directly.

scripts/lib/dependencies.ts
```diff
@@ -175,33 +288,58 @@ export async function readDependencyVersions(path: string): Promise<DependencyVe
...
+ * unless exactly one asset matches.
+ */
+export function selectAsset(context: DownloadContext, name: string, selector: AssetSelector): DependencyAsset {
+  const matches = selectAssets(context, name, selector);
 
+  if (matches.length !== 1) {
     throw new Error(
-      `No checksum recorded for ${ name } artifact "${ artifactName }" in ${ DEP_VERSIONS_PATH }. ` +
-      `Available: ${ available }`,
+      `Expected exactly one ${ name } asset for ${ describeSelector(selector) }, found ${ matches.length }.`,
```
mook-as: It would be nice to list the matching assets (by URL or something), so we can see what the issue is.

scripts/lib/dependencies.ts
```diff
@@ -389,19 +539,19 @@ export abstract class VersionedDependency implements Dependency {
   }
 
   /**
-   * Returns the sha256 of every artifact for the given version, verifying
+   * Resolves every {@link DependencyAsset} for the given version, verifying
    * each against any upstream checksum file the source publishes.  rddepman
-   * calls this at bump time and stores the result in `dependencies.yaml`.
-   * Classes that download nothing (e.g. `check-spelling`) return an empty map.
+   * calls this at bump time and records the result in the manifest.  Classes
+   * that download nothing (e.g. `check-spelling`) return an empty list.
```
mook-as: … I just realized that `check-spelling` is obsolete, because we use it as a git submodule instead:
https://github.com/rancher-sandbox/rancher-desktop-2/blob/1154ac9c36ccf2b5b6b91f982945ff0484f246b7/.gitmodules#L17-L19

Filed #552 because that is rather out of scope for this PR.  Feel free to use `golangci-lint` as the example instead; that will probably stay?

scripts/lib/dependencies.ts
```diff
@@ -453,39 +603,35 @@ export abstract class VersionedDependency implements Dependency {
...
-  depManifestCache ||= readDependencyManifest(DEP_VERSIONS_PATH);
-
-  return depManifestCache;
-}
-
 /**
- * A GlobalDependency is a dependency where the version is managed in the file
- * {@link DEP_VERSIONS_PATH}.
+ * A GlobalDependency is a dependency whose version and assets are managed in a
+ * `dependencies.yaml` manifest.  {@link manifestPath} selects which one;
```
mook-as:
```suggestion
 * `dependencies.yaml` manifest.  {@link GlobalDependency.manifestPath} selects which one;
```
Otherwise it couldn't find what you're talking about. Unfortunately this means fixing the line wrapping.

mook-as (commented):
Partial review; I'll probably look at the RD1 version after instead.

scripts/dependencies/distro.ts
```diff
@@ -0,0 +1,49 @@
...
+ * any other GitHub dependency; the Go downloader fetches it at build time
+ * (by arch and image variant), so there is no host install.
+ */
+export class Distro extends GlobalDependency(GitHubDependency) {
+  readonly name = 'distro';
+  readonly githubOwner = 'rancher-sandbox';
+  readonly githubRepo = 'rancher-desktop-opensuse';
+  readonly manifestPath = GUEST_DEP_VERSIONS_PATH;
+
+  download(_context: DownloadContext): Promise<void> {
```
mook-as, line 28: I wonder if we'll get enough deps in rdd that it would be worthwhile having an abstract base class that just implements `download`. Probably not yet.

mook-as (approved):

PR #512, opened 2026-06-26
download: retry transient network failures, not just DNS

jandubois:
`yarn install` postinstall intermittently fails in CI when the GitHub release CDN drops a connection mid-download (`read ECONNRESET`).

`fetchWithRetry` only retried `EAI_AGAIN`, and read it off the thrown TypeError rather than its `cause`, where `fetch` reports the underlying socket error. It now inspects both the error and its `cause`, retries a small set of transient codes (ECONNRESET, ETIMEDOUT, …), and gives up after a bounded number of attempts instead of looping forever.

### DIFF
--- a/scripts/lib/__tests__/download.spec.ts
+++ b/scripts/lib/__tests__/download.spec.ts
@@ -18,7 +18,6 @@ describe('fetchWithRetry', () => {
 
   beforeEach(() => {
     jest.spyOn(console, 'log').mockImplementation(() => {});
-    jest.spyOn(console, 'dir').mockImplementation(() => {});
   });
 
   afterEach(() => {
@@ -30,6 +29,21 @@ describe('fetchWithRetry', () => {
     const response = { ok: true } as Response;
     const fetchMock = jest.fn<typeof fetch>()
       .mockRejectedValueOnce(fetchFailed('ECONNRESET'))
+      .mockResolvedValueOnce(response);
+
+    global.fetch = fetchMock;
+
+    await expect(fetchWithRetry('https://example.test/x', { baseDelayMs: 0 })).resolves.toBe(response);
+    expect(fetchMock).toHaveBeenCalledTimes(2);
+  });
+
+  it('retries when the retryable code is nested deeper in the cause chain', async() => {
+    const response = { ok: true } as Response;
+    const nested = Object.assign(new TypeError('fetch failed'), {
+      cause: Object.assign(new Error('proxy error'), { cause: fetchFailed('ECONNRESET').cause }),
+    });
+    const fetchMock = jest.fn<typeof fetch>()
+      .mockRejectedValueOnce(nested)
       .mockResolvedValueOnce(response);
 
     global.fetch = fetchMock;
--- a/scripts/lib/__tests__/download.spec.ts
+++ b/scripts/lib/__tests__/download.spec.ts
@@ -14,27 +14,24 @@ function fetchFailed(code: string): TypeError {
 }
 
 describe('fetchWithRetry', () => {
-  const realFetch = global.fetch;
+  let fetchSpy: jest.SpiedFunction<typeof global.fetch>;
 
   beforeEach(() => {
     jest.spyOn(console, 'log').mockImplementation(() => {});
+    fetchSpy = jest.spyOn(global, 'fetch').mockResolvedValue({ ok: true } as Response);
   });
 
   afterEach(() => {
-    global.fetch = realFetch;
     jest.restoreAllMocks();
   });
 
   it('retries a transient ECONNRESET and returns the eventual response', async() => {
     const response = { ok: true } as Response;
-    const fetchMock = jest.fn<typeof fetch>()
-      .mockRejectedValueOnce(fetchFailed('ECONNRESET'))
-      .mockResolvedValueOnce(response);
 
-    global.fetch = fetchMock;
+    fetchSpy.mockRejectedValueOnce(fetchFailed('ECONNRESET')).mockResolvedValueOnce(response);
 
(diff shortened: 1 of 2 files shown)


scripts/lib/download.ts
```diff
@@ -29,17 +29,56 @@ export type ArchiveDownloadOptions = DownloadOptions & {
...
     try {
       return await fetch(url, { redirect: 'follow' });
     } catch (ex: any) {
-      if (ex?.errno === 'EAI_AGAIN') {
-        console.log(`Recoverable error downloading ${ url }, retrying...`);
-        continue;
+      const code = retryableNetworkError(ex);
+
+      if (!code || attempt >= retries) {
+        console.dir(ex);
```
mook-as: Huh, I wonder if this was old debugging code we forgot about.

scripts/lib/download.ts
```diff
@@ -29,17 +29,56 @@ export type ArchiveDownloadOptions = DownloadOptions & {
...
+
+/**
+ * Returns the retryable network error code carried by `ex`, if any.  `fetch`
+ * wraps the underlying socket error as `TypeError: fetch failed` with the real
+ * error in `cause`, so we inspect `cause` as well as the error itself.
+ */
+function retryableNetworkError(ex: any): string | undefined {
+  return [ex?.code, ex?.errno, ex?.cause?.code, ex?.cause?.errno]
+    .find((code): code is string => typeof code === 'string' && RETRYABLE_CODES.has(code));
+}
```
mook-as, line 58: This should be recursive on `cause`:

```suggestion
function retryableNetworkError(ex: any): string | undefined {
  for (const prop of ['code', 'errno']) {
    if (typeof ex?.[prop] === 'string' && RETRYABLE_CODES.has(ex[prop])) {
      return ex[prop];
    }
  }
  if (ex?.cause) {
    return retryableNetworkError(ex.cause);
  }
}
```
I'm not sure if we even need the type check.

scripts/lib/download.ts
```diff
@@ -29,17 +29,56 @@ export type ArchiveDownloadOptions = DownloadOptions & {
...
+        console.dir(ex);
+        throw ex;
+      }
+      const delayMs = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs);
+
+      console.log(`Recoverable error (${ code }) downloading ${ url }, retrying in ${ Math.round(delayMs / 1_000) }s (${ attempt + 1 }/${ retries })...`);
+      if (delayMs > 0) {
+        await new Promise<void>((resolve) => {
+          setTimeout(resolve, delayMs);
+        });
```
mook-as: Also consider:

```suggestion
        await util.promisify(setTimeout)(delayMs);
```
(Importing `node:util`)

mook-as (changes requested):

jandubois:
@mook-as Thanks! I've addressed all 3 of your concerns.

scripts/lib/download.ts
```diff
@@ -29,17 +30,59 @@ export type ArchiveDownloadOptions = DownloadOptions & {
...
+
+      if (!code || attempt >= retries) {
+        throw ex;
+      }
+      const delayMs = Math.min(baseDelayMs * 2 ** attempt, maxDelayMs);
+
+      console.log(`Recoverable error (${ code }) downloading ${ url }, retrying in ${ Math.round(delayMs / 1_000) }s (${ attempt + 1 }/${ retries })...`);
+      if (delayMs > 0) {
+        await util.promisify(setTimeout)(delayMs);
       }
```
mook-as: `util.promisfy(setTimeout)(0)` seems fine too.

scripts/lib/__tests__/download.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    cause: Object.assign(new Error(`read ${ code }`), { code }),
+  });
+}
+
+describe('fetchWithRetry', () => {
+  const realFetch = global.fetch;
+
+  beforeEach(() => {
+    jest.spyOn(console, 'log').mockImplementation(() => {});
+  });
```
mook-as: Any reason we don't just mock `fetch` like normal?

```ts
    jest.spyOn(global, 'fetch').mockResolvedValue({ ok: true } as Response);
```

The tests would then look like:
```ts
  it('gives up after exhausting the retry budget', async() => {
    jest.mocked(global.fetch).mockRejectedValue(fetchFailed('ECONNRESET'));

    await expect(fetchWithRetry('https://example.test/x', { retries: 2, baseDelayMs: 0 })).rejects.toThrow('fetch failed');
    expect(jest.mocked(global.fetch)).toHaveBeenCalledTimes(3); // 1 initial attempt + 2 retries
  });
```

Alternatively, we can do:
```ts
  let fetchSpy: jest.SpiedFunction<typeof global.fetch>;
  beforeEach(() => {
    jest.spyOn(console, 'log').mockImplementation(() => {});
    fetchSpy = jest.spyOn(global, 'fetch').mockResolvedValue({ ok: true } as Response);
  });
```

Where the tests would then be:
```ts
  it('gives up after exhausting the retry budget', async() => {
    fetchSpy.mockRejectedValue(fetchFailed('ECONNRESET'));

    await expect(fetchWithRetry('https://example.test/x', { retries: 2, baseDelayMs: 0 })).rejects.toThrow('fetch failed');
    expect(fetchSpy).toHaveBeenCalledTimes(3); // 1 initial attempt + 2 retries
  });
```

mook-as (dismissed):
I think it would be better to fix `global.fetch` in the test.

scripts/lib/__tests__/download.spec.ts
jandubois: Thanks, I had to rebase anyways, so I addressed both issues as well.

mook-as (approved):

Issue #498, opened 2026-06-26
Add `virtualMachine.cpus` and `virtualMachine.memory` properties on the App controller


jandubois:
Changing these should update the template ConfigMap.

**Validation**:

`cpus` cannot be less than 1 and not be greater than number of CPUs on the host. The defaulter sets it to `2` when not specified, or specified as `0`.

`memory` cannot be less than 2GB and not more than total memory on the host. The defaulter will use 25% of available memory, but no more than 6GB.

The defaulter rules match the RD1 settings, not Lima settings.

During controller startup the host cpus and memory should be determined and stored in a `HostInfo` Configmap. Validation will use these saved values, and the GUI can fetch them from there as well.

---

I think long term we might add a `HostInfo` resource that can also monitor other metrics, like available memory or disk space, that can generate diagnostics, or be queried by the GUI. But for now let's keep it simple with a ConfigMap.

jandubois:
I think we need the actual number of CPUs (and memory size) currently assigned in the UI, even when the spec fields are unset (default).

I think the best way to do this would be for the LimaVM controller to write these settings into the `LimaVM.Status` and then copied from there to the `App.Status`. Status is not writable by the admission controller, so this has to happen in the reconciler.

This is an extension of the scope of this issue and should be done in a separate PR.

PR #523, opened 2026-06-30
Add virtualMachine cpus/memory fields to App spec

Nino-K:
Add VirtualMachineSpec with CPUs (int) and Memory (`*resource.Quantity`) fields to AppSpec. Setting either field rewrites the Lima template at reconcile time; `RDD_VM_CPUS` env var is kept as a fallback for CI.

Validation is enforced by the admission webhook: cpus must be ≥ 1 and ≤ host CPU count; memory must be ≥ 2Gi and ≤ host total memory.

Host limits are detected at controller startup and stored in the rdd-system/rdd-host-info ConfigMap so the GUI can read the valid range without inspecting the host directly.

Related to: https://github.com/rancher-sandbox/rancher-desktop-2/issues/498


jandubois:
I see you created the ConfigMap in `rdd-system`. I expected it in `App.Spec.Namespace` (so "rancher-desktop").

But I think you are right, that we should make this an rdd resource, not an App resource.
I wonder if it wouldn't be easiest to just create a `HostInfo` CRD with a minimal reconciler that just sets the 2 properties in `HostInfo.Status`. Doesn't seem too much more work than creating a ConfigMap, and we don't have to change code again when we add more stuff to it. It should be non-namespaced singleton like `App`. What do you think?

Other feedback: You cannot `rdd set virtualMachine.cpus=0` to set it back to default because you have `minimum: 1`. Same thing you cannot set memory back to default.

jandubois (changes requested):
Some quick findings first:

* You should not be able to set `cpus` to a negative value.
* The code for setting the defaults number of CPUs should be only in the admission controller, not in the `rdd set` command. We don't have other integer properties right now, but this way of treating `0` is likely not generic for such properties.

jandubois:
I don't like how we are using regular expressions to modify the template. We should construct the template as a `limayaml` structure, and then marshall it into YAML before storing it in the ConfigMap. But this is too big a change for this PR, so leaving it for later.

For now I think we should remove `cpus` and `memory` from the template and instead add them at the end the same way we add the `param` mappings in `applySpecToTemplate`:

```go
	return strings.Join([]string{
		baseTemplate,
		fmt.Sprintf("cpus: %d", cpus),
		fmt.Sprintf("memory: %q", strconv.FormatInt(mem.Value(), 10)),
		"",
		"param:",
		fmt.Sprintf("  CONTAINER_ENGINE: %s", spec.ContainerEngine.Name),
...
		"",
	}, "\n"), nil
```

jandubois (changes requested):
Looks almost done now. I think we should still address I1 and I2 from https://jandubois.github.io/rancher-desktop-2/20260710-151244-pr-523.html, even though Rancher Desktop is not actually going to work on a machine whose specs are that low.

Actually, look through the suggestions too. Some are just about BATS style, but they are something I would comment on in my manual review as well.

jandubois (changes requested):
Unfortunately your fix introduced a critical new issue: https://jandubois.github.io/rancher-desktop-2/20260713-112138-pr-523.html

jandubois (approved):
Thanks, LGTM

PR #524, opened 2026-07-01
Answer two outstanding review comments from PR #479

jandubois:
This is the final follow-up to comments on #479.

* Comments on `wsl2RegistrationIsOrphaned`.
* Reword "No-op off Windows" to "No-op on non-Windows".

### DIFF
--- a/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
+++ b/rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
@@ -354,6 +354,12 @@ func (r *LimaVMReconciler) handleRestartNeeded(ctx context.Context, limaVM *v1al
 // on its own, because its WSL2 driver only re-imports a distro that is absent
 // from `wsl --list`. It arises when an instance directory is removed without
 // `wsl --unregister`. Always false off Windows, where VMType is never WSL2.
+//
+// inst.Status can read StatusRunning here even though no live process holds
+// the disk open. The only caller that reaches this with StatusRunning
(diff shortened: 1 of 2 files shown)


rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
```diff
@@ -354,6 +354,12 @@ func (r *LimaVMReconciler) handleRestartNeeded(ctx context.Context, limaVM *v1al
 // on its own, because its WSL2 driver only re-imports a distro that is absent
 // from `wsl --list`. It arises when an instance directory is removed without
 // `wsl --unregister`. Always false off Windows, where VMType is never WSL2.
+//
+// inst.Status can read StatusRunning here even though no live process holds
+// the disk open. The only caller that reaches this with StatusRunning
+// (handleWatchedState's phaseStopped-but-shouldRun case) already ran
+// stopInstanceForcibly, which terminates the WSL2 distro first. Preserve that
+// ordering in any new startInstance caller, or this stat becomes a race.
```
mook-as, line 362: Having a random comment doesn't seem like a good way of keeping the invariant.

mook-as (approved):
Looks fine, but I worry that we'll accidentally break the invariant in the future.

rdd/pkg/controllers/lima/limavm/controllers/limavm_lifecycle.go
jandubois: Any suggestions how to preserve an invariant in the face of random future code changes? Documenting the expectation was the best I could think off.

PR #529, opened 2026-07-03
RDA: Put window logs in separate files

mook-as:
This makes things easier to track down.

### DIFF
--- a/pkg/rancher-desktop/window/index.ts
+++ b/pkg/rancher-desktop/window/index.ts
@@ -70,10 +70,10 @@ export function createWindow(name: string, url: string, options: Electron.Browse
 
   window = new BrowserWindow(options);
   window.webContents.on('console-message', (event) => {
-    const { level, lineNumber, message } = event;
+    const { level, sourceId, lineNumber, message } = event;
     const method = level === 'warning' ? 'warn' : level;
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/window/index.ts
```diff
@@ -92,7 +92,7 @@ export function createWindow(name: string, url: string, options: Electron.Browse
     return { action: 'deny' };
   });
   window.webContents.on('did-fail-load', (event, errorCode, errorDescription, url) => {
-    console.log(`Failed to load ${ url }: ${ errorCode } (${ errorDescription })`, event);
+    console.log(`Failed to load ${ name }(${ url }): ${ errorCode } (${ errorDescription })`, event);
   });
   console.debug('createWindow() name:', name, ' url:', url);
```
jandubois, line 97: Why don't these messages go to the window-level log file as well, to keep them all together. Also guarantees that the log file is created when the window is created.

jandubois (dismissed):
Thanks, LGTM

The unit test failure seems to be a flake that should go away on a re-run. I'll see if I can figure out why it is flaking.

jandubois (approved):
Thanks, LGTM

PR #530, opened 2026-07-03
RDA: Vuex: Wait before triggering `onChange` on connecting to RDD

mook-as:
During watching of resources against the RDD control plane, we made a call to the `onChange` handler immediately before listing the initial set of resources.  This meant we ended up resolving the `#loaded` promise, which caused downstream code to start assuming the load is complete before we even started listing the initial resources.  Add a timeout on the callback from the `connect` handler, so that we give some time for the listing to return; however, we still need to make the call from that spot if there were no resources.

### DIFF
--- a/pkg/rancher-desktop/store/rddConnection.ts
+++ b/pkg/rancher-desktop/store/rddConnection.ts
@@ -568,20 +568,21 @@ class Watcher<
   K extends string,
   T extends RDDClient.KubernetesObject,
 > {
-  #type:           K;
-  #path:           () => string;
-  #listFn:         () => Promise<RDDClient.KubernetesListObject<T>>;
-  #doneFn:         (error?: any) => void;
-  #watch:          RDDClient.Watch;
-  #commit:         Commit<any>;
-  #namespace?:     string;
-  #labelSelector?: () => string | undefined;
-  #fieldSelector?: () => string | undefined;
-  #items:          readonly T[] = [];
+  #type:               K;
+  #path:               () => string;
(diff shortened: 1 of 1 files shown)


pkg/rancher-desktop/store/rddConnection.ts
```diff
@@ -634,17 +635,24 @@ class Watcher<
...
+      // If there are no existing resources, we should still eventually call the
+      // change handler at some point.  However, we need an extra long timeout
+      // here because the initial list could take a while.
+      clearTimeout(this.#initialListTimeout);
+      this.#initialListTimeout = setTimeout(() => {
+        // Only trigger a change if this is the current watcher.
+        if (Object.is(watcher, this.#watcher)) {
+          this.onChange();
+        }
+      }, 1_000);
```
jandubois, line 647: I don't understand how this is going to work. Isn't this timeout adding back the race this PR is trying to fix (#loaded gets resolved before the data arrives). 1s doesn't sound "extra long" to me.

Can't ListWatch generate a dedicated event when the initial list has been delivered?

mook-as: Given that RDD is running on the same machine, 1s is way more than plenty: at that point the user would have noticed the stuck view, so at least showing an empty list and then it being updated later is better.

I was trying to avoid changing `cache.ts` too much, since that's upstream code, and the more changes we make the harder it would be for us to potentially update (e.g. when they have it all using `fetch` in the browser). But if you feel like that's a better way to go we can do that instead.

jandubois (approved):
Thanks, LGTM

PR #531, opened 2026-07-03
RDA: Vuex: Use app as backing for preferences

mook-as:
This rewrites the vuex preferences store to use `App.spec` instead of the `Settings` object that used IPC.  The last commit is a bunch of tests.

### DIFF
--- a/pkg/rancher-desktop/entry/store.ts
+++ b/pkg/rancher-desktop/entry/store.ts
@@ -37,11 +37,6 @@ export type Modules = typeof modules;
 export type RootState = {
   [K in keyof Modules]: ReturnType<Modules[K]['state']>;
 };
-export type RootGetters = {
-  [K in keyof Modules]: Modules[K] extends { getters: any } ? {
-    [key in keyof Modules[K]['getters']]: ReturnType<Modules[K]['getters'][key]>;
-  } : never;
-};
 
 export default createStore<any>({
   modules: Object.fromEntries(Object.entries(modules).map(([k, v]) => [k, { namespaced: true, ...v }])),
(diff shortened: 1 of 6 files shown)


jandubois (dismissed):
Thanks, LGTM

I did run an AI review. I think most findings are just because the whole code is still mid-migration to the new mechanism. Only I3 seems worth a look. If that is covered in later work too, then go ahead and merge!

https://jandubois.github.io/rancher-desktop-2/20260706-112728-pr-531.html

mook-as:
I was originally going to make follow-on commits, but doing the fixes with AI review rounds led to even more changes, to the point where it didn't seem useful to push this as-is anymore; sorry.

> I1, I2

Irrelevant, none of the callers exist yet (it's in my [WIP branch](https://github.com/mook-as/rd2/tree/rda/settings)).

> I3. commit JSON-Patch add fails when the parent object is absent (enabling Kubernetes)

It exists, through the magic of the defaulter on the RDD side.

> S1. MutationsType loosened globally to admit one non-SET_ mutation

I intend to have more non-`SET_` mutations in the future; turns out doing _just_ the sets are too restrictive.  That's mostly for use by the actions, though; the actual vue components will still need to go through `dispatch`.

> S2. RecursiveReadonly not updated to object like its siblings 

Valid; will fix.

> S3. state.preferences typed Required<AppSpec> but initialized to {}

Fair; this eventually ended up going away.

> S4. commit can drop a modify that interleaves during its await 

The suggested fix just moves the problem: if `state.changes` is modified (with the same keys), it'll get clobbered too.  This lead to… significant changes.

> S5. modify's merge-patch dry-run cannot represent a field removal

Will fix; basically ended up not having the preferences as a spec, just a set of changes, and a getter to return the merged result.  So `modify` and `commit` end up both using JSON patch.

> S6. Watcher leaves unrelated background updates stale after a revert

No longer relevant, see S5.

(I haven't pushed my changes yet.)

jandubois (approved):
Thanks, LGTM

PR #532, opened 2026-07-03
RDA: Convert transient preferences

mook-as:
This converts the existing transient settings to transient preferences; it's still stored in the main process (and reflected to the windows), the main difference is that it's now reflected in a vuex store.  All interaction is done via that store, and any changes are sent to the main process and reflected to all windows.

This needs to be reviewed after #531 is merged.


jandubois (approved):
Thanks, LGTM

There are still left-over references to `@pkg/config/transientSettings`:

```console
❯ grep -r pkg/config/transientSettings *
components/Preferences/BodyApplication.vue:import type { TransientSettings } from '@pkg/config/transientSettings';
components/Preferences/BodyContainerEngine.vue:import { TransientSettings } from '@pkg/config/transientSettings';
components/IncompatiblePreferencesAlert.vue:import type { NavItemName } from '@pkg/config/transientSettings';
```

But neither `yarn lint` nor `yarn build` notice (see
https://github.com/rancher-sandbox/rancher-desktop/issues/10461), so I guess it doesn't matter.

PR #533, opened 2026-07-04
Wait for async kubeconfig write before removeKubeContext returns

jandubois:
The `manageKubeContext` goroutine signaled completion before it wrote current-context to `~/.kube/config`, so `removeKubeContext` could return while that write was still in flight. In unit tests the write then raced `t.TempDir` cleanup and produced the "directory not empty" flake in `Test_Reconcile_KubernetesReady_GatedUntilNodeReady`. In production it could restore a current-context that `removeKubeContext` had just cleared.

The early signal was meant to keep a slow kubeconfig write from blocking the wait, but `removeKubeContext` writes the same file right after the wait, so a slow write delays it there anyway.

The regression test parks the probe inside its kubeconfig write and asserts that `removeKubeContext` does not return until the write lands. It fails against the old code.

The second commit drops `contextProbeGen`. Every path that supersedes the probe cancels its `probeCtx` first, so `probeCtx.Err() == nil` already means it is the current probe.

Example of a flaked job: https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/28683701733/job/85072203379

### DIFF
--- a/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
+++ b/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
@@ -464,14 +464,9 @@ func (r *KubernetesReconciler) manageKubeContext(ctx context.Context) error {
 
 	r.contextProbeWg.Add(1)
 	go func() {
-		// Use sync.Once so contextProbeWg.Done() fires exactly once, either
-		// explicitly after the HTTP probe or via defer on early return.
-		// Crucially, Done() must be called BEFORE setCurrentKubeContext so that
-		// removeKubeContext.Wait() is not held hostage by a slow or blocking
-		// write to ~/.kube/config.
-		var wgDone sync.Once
-		signalDone := func() { wgDone.Do(r.contextProbeWg.Done) }
-		defer signalDone()
+		// Done fires after the kubeconfig write below, so removeKubeContext.Wait()
+		// covers that write instead of racing it.
+		defer r.contextProbeWg.Done()
 		defer func() {
 			r.contextMu.Lock()
 			if r.contextProbeGen == myGen {
@@ -488,10 +483,6 @@ func (r *KubernetesReconciler) manageKubeContext(ctx context.Context) error {
 		}
 
 		healthy := probeCurrentKubeContext(probeCtx, current)
-
-		// Signal the WaitGroup before writing to ~/.kube/config so that
-		// removeKubeContext.Wait() is not blocked by a slow file write.
-		signalDone()
 
 		// Guard against writing after removeKubeContext cancelled probeCtx.
 		if !healthy && probeCtx.Err() == nil {
--- a/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
+++ b/rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
@@ -102,12 +102,10 @@ type KubernetesReconciler struct {
 	// inject a path under a temp directory.
 	InstanceKubeConfigPath string
 
-	// contextMu protects contextProbeCancel and contextProbeGen.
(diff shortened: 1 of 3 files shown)


rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
```diff
@@ -464,14 +464,9 @@ func (r *KubernetesReconciler) manageKubeContext(ctx context.Context) error {
 
 	r.contextProbeWg.Add(1)
 	go func() {
-		// Use sync.Once so contextProbeWg.Done() fires exactly once, either
-		// explicitly after the HTTP probe or via defer on early return.
-		// Crucially, Done() must be called BEFORE setCurrentKubeContext so that
-		// removeKubeContext.Wait() is not held hostage by a slow or blocking
-		// write to ~/.kube/config.
```
mook-as, line 471: So this is basically removing the _Crucially, …_ part? It then feels like this should the PR description should explain why it is _not_ crucial. That is, what changed to make the statement invalid; the fact that you're changing the behaviour is clear enough.

Perhaps this PR should be reviewed by Nino instead…

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
mook-as: Out of scope for this PR, probably, but:

`contextProbeGen` is not really necessary: we can always clear `contextProbeCancel` if `probeCtx` hasn't been cancelled (that is, `probeCtx.Err()` is `nil`, like we checked below).

mook-as (dismissed):
I think the PR description (and the commit message) could be reworded for clarity:

The `manageKubeContext` goroutine signaled completion before `~/.kube/config` was updated.  This made it possible for `removeKubeContext` to complete while the write is still in flight.  In unit tests, the write could then occur while the test is being torn down, cause errors with `t.TempDir` cleanup.  In production, this could cause the context to be incorrectly set to context that have been removed.

Hopefully this sounds a bit less like Claude.

---
The PR looks fine, but I'd like to make sure the _crucial_ bit is… not very.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
jandubois: I don't understand the reasoning behind not waiting at all; how would you not wait and still expect not to have a race condition?

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
jandubois: Seems small enough and simplifies the code, so I'll fold it in.

jandubois:
Asking @Nino-K for review, to hopefully clarify the intention of the old "don't wait for deletion before signalling 'done'" logic.

jandubois:
Made a draft. Should be rebased once #547 has been merged.

rdd/pkg/controllers/app/kubernetes/controllers/kube_reconciler.go
Nino-K: When I originally created the controller, my intention from the original design was: `removeKubeContext` calls `contextProbeWg.Wait()` to block teardown until the goroutine finishes. I guess I was worried about that if the probe's final `setCurrentKubeContext` write to `~/.kube/config` were slow or blocking, it would "hold hostage" that `Wait()`. So the goroutine called `Done()` before the write, intentionally excluding the write from what `Wait()` covers. But, I can see that idea doesn't hold because of what `removeKubeContext` does right after `Wait()`, it writes the same file anyway via `clearCurrentKubeContext` and `deleteKubeContext`, so a slow write delays teardown there regardless.

Nino-K (approved):

PR #544, opened 2026-07-08
CI: Get rddepman running via vault app credentials

jandubois:
Replace repo secret with vault token.

Previous run: https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/28963422125/job/85940448568 (created PRs that have been merged)

Latest run: https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/28970280660 (but does not need to create more PRs)

### DIFF
--- a/.github/actions/get-token/action.yaml
+++ b/.github/actions/get-token/action.yaml
@@ -12,24 +12,28 @@ inputs:
 outputs:
   token:
     description: The GitHub token retrieved
-    value: ${{ github.repository_owner == 'rancher-sandbox' && steps.copy-token.outputs.token || steps.get-secret.outputs.token }}
+    value: ${{ github.repository == 'rancher-sandbox/rancher-desktop-2' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
 runs:
   using: composite
   steps:
   - id: vault
     name: Read vault secrets
-    if: github.repository_owner == 'rancher-sandbox'
+    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
     uses: rancher-eio/read-vault-secrets@7282bf97898cd1c16c89f837e0bb442e6d384c89 # main
     with:
       secrets: |
-        github/token/${{ github.repository_owner }}--rancher-desktop-daemon--pull_requests--write token | GH_TOKEN
-  - id: copy-token
-    if: github.repository_owner == 'rancher-sandbox'
-    shell: bash
-    run: echo "token=$GH_TOKEN" >> "$GITHUB_OUTPUT"
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials privateKey | PRIVATE_KEY
+  - id: gen-token
+    name: Generate token
+    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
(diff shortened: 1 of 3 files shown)


.github/actions/get-token/action.yaml
```diff
@@ -12,24 +12,28 @@ inputs:
...
     description: The GitHub token retrieved
-    value: ${{ github.repository_owner == 'rancher-sandbox' && steps.copy-token.outputs.token || steps.get-secret.outputs.token }}
+    value: ${{ github.repository == 'rancher-sandbox/rancher-desktop-2' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
 runs:
   using: composite
   steps:
   - id: vault
     name: Read vault secrets
-    if: github.repository_owner == 'rancher-sandbox'
+    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
```
mook-as, line 21: I wonder if we should just always run this, with `continue-on-error`, and fall back to `inputs.token-secret` if this fails? (The `get-token` step would be `if: steps.value.output == success`)

.github/actions/get-token/action.yaml
```diff
@@ -12,24 +12,28 @@ inputs:
...
-    shell: bash
-    run: echo "token=$GH_TOKEN" >> "$GITHUB_OUTPUT"
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials privateKey | PRIVATE_KEY
+  - id: gen-token
+    name: Generate token
+    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
+    uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+    with:
+      app-id: ${{ env.APP_ID }}
```
mook-as, line 32: > Warning: Input 'app-id' has been deprecated with message: Use 'client-id' instead.

.github/workflows/rddepman.yaml
```diff
@@ -17,20 +18,33 @@ jobs:
     - id: calc
       run: echo "HAS_SECRET=${HAS_SECRET}" >> "${GITHUB_OUTPUT}"
       env:
-        HAS_SECRET: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW != '' }}
+        HAS_SECRET: ${{ github.repository == 'rancher-sandbox/rancher-desktop-2' || secrets.RUN_WORKFLOW_FROM_WORKFLOW != '' }}
```
mook-as, line 21: This seems somewhat hard to maintain.

I guess it's like this so we don't have to clone the repo if we can't get a token? At this point it might be easier to just always clone, and simply not run `yarn rddepman` if we failed to get a token…

mook-as (changes requested):
The only thing blocking merge is the `app-id` thing; others are optional (and probably out of scope).

.github/actions/get-token/action.yaml
mook-as: Jan notes that it's not actually a key rename. I guess https://github.blog/changelog/2024-08-23-client-ids-are-now-included-in-app-api-responses/ provides information, which indicates that `app-id` is a database ID. And Jan points to https://github.blog/changelog/2024-05-01-github-apps-can-now-use-the-client-id-to-fetch-installation-tokens/ which also says something similar.  Sigh.

mook-as (approved):

PR #547, opened 2026-07-09
Serialize kubeconfig writes and make them atomic

jandubois:
Nothing serialized the writers of `~/.kube/config`, and each rewrote the file in place, so overlapping writers could tear it into invalid YAML (the failed ubuntu bats run on #533) or read it empty mid-truncate and then silently drop the user's other clusters. The file was also rewritten on every reconcile even when unchanged, giving these rare races endless chances to fire.

The first commit adds a shared `atomicfile` helper that owns the whole update cycle (shell-profile editing will reuse it later); the second routes all kubeconfig writes through it.

Merging this before #533 forecloses the race that failed its CI; #533 stays as-is since it fixes a separate ordering race.

### DIFF
--- a/rdd/pkg/util/atomicfile/atomicfile.go
+++ b/rdd/pkg/util/atomicfile/atomicfile.go
@@ -1,0 +1,132 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Package atomicfile updates file contents atomically and race-free: updates
+// to the same file are serialized, and the new content is staged in a
+// temporary file and renamed into place, so readers never see an empty or
+// partially written file.
+package atomicfile
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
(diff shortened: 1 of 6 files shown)


rdd/pkg/util/atomicfile/atomicfile.go
```diff
@@ -0,0 +1,132 @@
...
+func resolveTarget(path string) string {
+	resolved, err := filepath.EvalSymlinks(path)
+	if err == nil {
+		return resolved
+	}
+	var pathErr *fs.PathError
+	if errors.Is(err, fs.ErrNotExist) && errors.As(err, &pathErr) {
+		return pathErr.Path
+	}
+	return path
```
Nino-K: If the `err` is a permission-related error or symlink loop `ELOOP`, this function ignores them and returns the path, which could cause `os.ReadFile(target)` or `os.CreateTemp(...)` to fail with a confusing message.

jandubois: Thanks, there is a whole class of errors potentially hiding behind this ignored error, so I think exposing the error to the caller is the right think.

Nino-K (approved):
LGTM

PR #550, opened 2026-07-09
Create `k3s-versions` config map

### LINKED ISSUE #500: Write `k3s-version` data into a ConfigMap
The data is currently only available in the App controller for validating the `kubernetes.version` field, but it is also needed in the Settings dialog to populate the dropdown list.

Save the data structure into the `k3s-versions` ConfigMap. Overwrite if it already exists.

mook-as:
This adds a controller to force create a `k3s-versions` config map in the `rdd-system` namespace, containing the channels and versions of k3s available.

This is done on `app` creation, but the app controller (defaulter + validator) just uses the raw data structure instead of round-tripping it through the control plane.  This means the config map is only for use by the front end and any other introspection tools.

Since we already have a config map validator elsewhere (for the lima yaml), the webhook machinery has changed to always use custom paths based on the webhook name; otherwise the two would conflict and only one would be called.


rdd/.golangci.yml
```diff
@@ -128,6 +128,11 @@ linters:
     - common-false-positives
     - legacy
     - std-error-handling
+    rules:
+    - # Exclude blank imports of "embed", that is used with `//go:embed` directives.
+      linters: [ revive ]
+      source: ^(?:import)?\s+_\s+"embed"$
+      text: '^blank-imports:'
```
jandubois: Why is this needed? `revive` only flags a blank `embed` import if there are no `//go:embed` directives in the file. I think you just forgot to remove the one reference that is now unused.

rdd/pkg/controllers/app/k3sversions/controller.go
```diff
@@ -0,0 +1,121 @@
...
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Package k3sversions manages the k3s-versions config map, providing
+// information about supported Kubernetes versions.
+package k3sversions
+
+import (
+	_ "embed"
+
```
jandubois: Why is this here? There is no `//go:embed` in this file.

rdd/pkg/controllers/app/k3sversions/controllers/k3s_versions_validator.go
```diff
@@ -0,0 +1,50 @@
...
+	return nil, nil
+}
+
+// ValidateUpdate implements [admission.Validator].
+func (k *K3sVersionsValidator) ValidateUpdate(context.Context, *v1.ConfigMap, *v1.ConfigMap) (admission.Warnings, error) {
+	return nil, nil
+}
+
+// ValidateDelete implements [admission.Validator].
+func (k *K3sVersionsValidator) ValidateDelete(ctx context.Context, obj *v1.ConfigMap) (admission.Warnings, error) {
```
jandubois, line 70: Should we have unit tests/BATS to confirm that this works?

rdd/pkg/controllers/app/k3sversions/controller.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return err
+	}
+	c.webhookManagers = append(c.webhookManagers, managers...)
+	return nil
+}
+
+// RegisterWithManager implements the complete controller registration for both embedded and external modes.
+// It registers the CRD types with the scheme and sets up the controller with the manager.
+// It also registers Lima types, which allows App controller to create and watch LimaVM resources.
```
jandubois: Looks like unedited copy: there is no CRD to register, and there are no LimaVM resources to watch.

rdd/pkg/controllers/app/k3sversions/controllers/k3s_versions_reconciler.go
```diff
@@ -0,0 +1,121 @@
...
+
+// DesiredLabels returns the desired labels for the k3s-versions config map.
+// This is exported so the webhook can use it to filter for the config map.
+var DesiredLabels = sync.OnceValue(func() map[string]string {
+	return map[string]string{
+		"app.kubernetes.io/name":       "rdd-k3s-versions",
+		"app.kubernetes.io/component":  "k3s-versions",
+		"app.kubernetes.io/managed-by": "rancher-desktop-daemon",
+	}
+})
```
jandubois: This doesn't make sense: if you use `sync.OnceValue`, then there is just a single map, and you return the same map on every call. You could just make the map a variable. Or if you want to avoid the risk of a caller modifying the map, then you need to remove the `OnceValue`.

jandubois (changes requested):
I don't think the ConfigMap should be created in the `"rdd-system"` namespace. It is an App resource and should live in the `App.spec.namespace` namespace.

The docs in `api_k3sversion.md` are out-of-date; they claim this is a resource with a CRD, not a ConfigMap.

And some smaller issues...

jandubois (approved):
Thanks, LGTM

AI review: https://jandubois.github.io/rancher-desktop-2/20260717-174447-pr-550.html

Take a look at I1, I think it is worth it to mirror the code LimaVM does in protecting the ConfigMap. But ultimately we can't prevent people from trashing their own instance, so this is more a consistency of implementation argument.

Theoretically "Lexical version sort on duplicate builds" is also a concern, but I don't think there will ever be a `k3s10` release, so I don't think it can happen.

Issue #453, opened 2026-06-15
Use vault secrets instead of GitHub action secrets for OBS uploads


jandubois:
Same thing for `OBS_WEBHOOK_TOKEN`.

mook-as:
Do we have vault secrets available so I can test it out? It'll be hard to make PRs if those are not available.

PR #553, opened 2026-07-10
CI: Use vault secrets for OBS uploads

mook-as:
For #453; draft because I need to be on this repo to test the vault action.

The OBS trigger step has been moved to a new job, to help scope access to the tokens to fewer things.  The overall permissions for the workflow has been explicitly set to `contents: read` as well.

Sample run: https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/29133328785/job/86494046525

That had an extra (intentionally not signed-off) commit to force trigger when not on `main`; that has been dropped.

#554 is a PR from a fork; that should fail (because it can't get secrets).

### DIFF
--- a/.github/actions/get-token/action.yaml
+++ b/.github/actions/get-token/action.yaml
@@ -8,7 +8,7 @@ description: >-
 inputs:
   token-secret:
     description: Secret to fall back to
-    required: false
+    required: true
 outputs:
   token:
     description: The GitHub token retrieved
@@ -18,7 +18,7 @@ runs:
   steps:
   - id: vault
     name: Read vault secrets
-    continue-on-error: true
+    if: github.repository_owner == 'rancher-sandbox'
     uses: rancher-eio/read-vault-secrets@7282bf97898cd1c16c89f837e0bb442e6d384c89 # main
     with:
       secrets: |
@@ -32,9 +32,9 @@ runs:
       client-id: ${{ env.CLIENT_ID }}
       private-key: ${{ env.PRIVATE_KEY }}
   - id: get-secret
-    name: Fetch secret.
-    if: steps.gen-token.outcome != 'success'
-    run: echo "token=$SECRET" >> "$GITHUB_OUTPUT"
+    name: Fall back to GitHub-hosted secrets
+    if: steps.vault.outcome == 'skipped'
+    run: echo "token=${SECRET:?Secret not found}" >> "$GITHUB_OUTPUT"
     shell: bash
     env:
       SECRET: ${{ inputs.token-secret }}
--- a/.github/actions/get-token/action.yaml
+++ b/.github/actions/get-token/action.yaml
@@ -12,28 +12,28 @@ inputs:
 outputs:
   token:
     description: The GitHub token retrieved
-    value: ${{ github.repository == 'rancher-sandbox/rancher-desktop-2' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
+    value: ${{ steps.gen-token.outcome == 'success' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
 runs:
   using: composite
   steps:
   - id: vault
     name: Read vault secrets
-    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
+    continue-on-error: true
     uses: rancher-eio/read-vault-secrets@7282bf97898cd1c16c89f837e0bb442e6d384c89 # main
     with:
       secrets: |
-        secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials clientId | CLIENT_ID ;
         secret/data/github/repo/${{ github.repository }}/github/app-credentials privateKey | PRIVATE_KEY
   - id: gen-token
     name: Generate token
-    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
+    if: steps.vault.outcome == 'success'
     uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
     with:
-      app-id: ${{ env.APP_ID }}
+      client-id: ${{ env.CLIENT_ID }}
       private-key: ${{ env.PRIVATE_KEY }}
   - id: get-secret
     name: Fetch secret.
-    if: github.repository != 'rancher-sandbox/rancher-desktop-2'
+    if: steps.gen-token.outcome != 'success'
     run: echo "token=$SECRET" >> "$GITHUB_OUTPUT"
     shell: bash
     env:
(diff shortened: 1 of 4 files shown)


jandubois (changes requested):
`package` workflow looks fine, but not sure about `rddepman`:

I haven't tested this, but in general tokens created by `create-github-app-token` are revoked in the `post` step and cannot be used across jobs.

There is a `skip-token-revoke` option (https://github.com/actions/create-github-app-token#skip-token-revoke), but I think in the interest of repo hardening we should probably duplicate the `get-token` action in each job that needs it instead of keeping the token valid across jobs.

.github/actions/get-token/action.yaml
```diff
@@ -12,28 +12,28 @@ inputs:
...
     description: The GitHub token retrieved
-    value: ${{ github.repository == 'rancher-sandbox/rancher-desktop-2' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
+    value: ${{ steps.gen-token.outcome == 'success' && steps.gen-token.outputs.token || steps.get-secret.outputs.token }}
 runs:
   using: composite
   steps:
   - id: vault
     name: Read vault secrets
-    if: github.repository == 'rancher-sandbox/rancher-desktop-2'
+    continue-on-error: true
```
jandubois: Why do we continue on error? That way we don't get any kind of failure if the vault access stops working because something has changed in the permissions. We continue on error here, and then skip all the remaining steps.

Now this is somewhat at odds with what AI tells me, in that `continue-on-error` doesn't work on nested composite actions: https://github.com/actions/runner/issues/3510

So I don't know if this is fixed or not, but I don't know why we wouldn't want to fail. I guess it is for forks, but there the previous condition would work, so why do we replace it?

.github/workflows/package.yaml
```diff
@@ -110,32 +113,54 @@ jobs:
         name: Rancher Desktop-linux.zip
         path: dist/rancher-desktop-*-linux.zip
         if-no-files-found: error
+
+  trigger-obs:
+    name: Trigger OBS build
+    needs: package
```
jandubois, line 119: I guess it doesn't matter, but now the OBS build will wait until all platforms have been packaged (including Windows, which runs very long). Previously we would run as part of the Linux matrix entry.

.github/actions/get-token/action.yaml
```diff
@@ -12,28 +12,28 @@ inputs:
...
     uses: rancher-eio/read-vault-secrets@7282bf97898cd1c16c89f837e0bb442e6d384c89 # main
     with:
       secrets: |
-        secret/data/github/repo/${{ github.repository }}/github/app-credentials appId | APP_ID ;
+        secret/data/github/repo/${{ github.repository }}/github/app-credentials clientId | CLIENT_ID ;
```
jandubois, line 25: Did you test this? I guess we can simply see if it works or not after we merge it, and figure it out afterwards if it doesn't.

.github/actions/get-token/action.yaml
mook-as: We continue on error because on forks, `rancher-eio/read-vault-secrets` should fail (because the vault does not contain any secrets for the forks).  This also includes pull requests made from forks.

The linked issue does not seem to be relevant; it seems to say:
- If you have a step that uses a composite action (call it _alpha_);
- And if that composite action itself uses a composite action (_beta_);
- And if, in the main workflow, you mark _alpha_ as `continue-on-error: true`;
- And if that inner composite action (_beta_) fails in post-execution;
- Then, that workflow fails.

We are not expecting to catch any errors in any _post_- steps anywhere here, so we should be safe.

I wanted to replace the previous condition because it has literal references, and I wanted to excise all of those.

.github/workflows/package.yaml
mook-as: Yeah, but in return, we limit the `id-token: write` to a much smaller scope.

.github/actions/get-token/action.yaml
mook-as: Yes, see the sample run linked in the description. It eventually failed because the branch name was bad, but it did manage to get a token.  That's why I changed the `curl` to include `--fail-with-body` so it would fail properly instead.

.github/workflows/rddepman.yaml
```diff
@@ -36,15 +24,15 @@ jobs:
...
+      - id: get-token
         uses: ./.github/actions/get-token
+        continue-on-error: true
         with:
           token-secret: ${{ secrets.RUN_WORKFLOW_FROM_WORKFLOW }}
 
-      - uses: ./.github/actions/yarn-install
-
       - run: yarn rddepman
+        if: steps.get-token.outcome == 'success'
```
jandubois: We should check that the local secret actually exists. Because the `get-secret` fallback will always succeed.

```suggestion
        if: steps.get-token.outcome == 'success' && steps.get-token.outputs.token != ''
```

But this condition will also skip running `rddepman` if the `get-token` step is broken, which doesn't seem ideal.

mook-as: I made `get-token` fail if the fallback is needed and the secrets are empty.  We only use it in one other place right now, so I marked _that_ as `continue-on-error`.

jandubois: I don't understand: `get-token` fails because there is a problem with the vault. It is `continue-on-error: true`, but `outcome != 'success'`, so `yarn rddepman` will be silently skipped, no?

jandubois (commented):
We need comments explaining the `OBS_DEVEL_PROJECT` repo secret.

.github/workflows/rddepman.yaml
mook-as: Ah, I fixed the "`get-secret` fallback will always succeed" part (by making it error if the secret is empty), but not the vault-is-broken part.

I suppose I can just drop `continue-on-error`/`if … == 'success'`, and make this just fail if the set up is not correct? Forks that don't want to deal with it can just disable the whole workflow, in that case.

.github/workflows/rddepman.yaml
jandubois, line 1: Line 3 comment is wrong: "needs the RUN_WORKFLOW_FROM_WORKFLOW secret, and skips when that is unset."

.github/workflows/rddepman.yaml
```diff
@@ -36,15 +24,13 @@ jobs:
           # github.token would override that, so keep it out of git config.
           persist-credentials: false
 
-      - # rddepman needs a token whose pull requests trigger CI; the
```
jandubois, line 39: I think this comment is useful: we need special token because we want to create PRs that will trigger workflows. Otherwise we could have used the default token.

.github/actions/get-token/action.yaml
```diff
@@ -8,33 +8,33 @@ description: >-
 inputs:
   token-secret:
     description: Secret to fall back to
-    required: false
+    required: true
```
jandubois, line 11: Technically it is not required. You can run the workflow without it as long as you execute in the upstream repo.

jandubois (approved):
Thanks, LGTM

I only have a few comment-level comments

Issue #499, opened 2026-06-26
Add back the Settings dialog


jandubois:
We used to call it `Preferences`, but nobody does this anymore, so let's call it `Settings` everywhere.

For Alpha 2 we want to limit to just a few settings:

`application.addPath` (see #497)
`virtualMachine.cpus` and `virtualMachine.memory` (see #498)
`containerEngine.name`
`kubernetes.enabled` and `kubernetes.version`

mook-as:
I believe Rancher deshboard still calls it Preferences:
https://github.com/rancher/dashboard/blob/c957970f51e053b490310b5109622b2e1ae5009a/shell/assets/translations/en-us.yaml#L300

<img width="220" height="179" alt="Image" src="https://github.com/user-attachments/assets/6edc9d37-501e-44f2-82f2-4140a15b905b" />

(From Rancher 2.13.3 installed via helm)

If anything, we should excise the _settings_ phrasing.

mook-as:
I think with #585 we have everything we need for the next alpha:
- #560 is `kubernetes.enabled`
- #569 is `application.updates.enabled` (which doesn't do anything, because the updater is still stubbed out)
- #571 is `kubernetes.version`
- #585 is `virtualMachine.cpus` and `virtualMachine.memory`
- `containerEngine.name` is intentionally not listed yet.

jandubois:
> If anything, we should excise the _settings_ phrasing.

I don't think so. We should drop `Preferences` and start standardizing on `Settings` instead.

Windows went from `Control Panel` to `Settings`. macOS went from `Preferences` to `Settings`. GNOME and KDE both have `Settings`.

Most cross-platform apps, especially when derived from Chrome, use `Settings`: `Slack`, `VS Code`, `Chrome`, `Discord`, `Spotify`, `Zoom`.

Firefox went from `Options` on Windows and `Preferences` on macOS to `Settings` in Firefox 89.

jandubois:
This has shipped, so it must be done. 😄

PR #559, opened 2026-07-14
RDD: Introduce app prefs

mook-as:
This adds a new `application` section to the app spec, for use by the Electron front end.

The last commit is just regenerating; it is therefore recommended to review this one commit at a time, and just run `yarn generate:rdd-client` and checking for differences for the last one.

This is part of #499.


rdd/pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -193,6 +222,9 @@ type AppSpec struct {
 	// virtualMachine specifies the VM resource allocation (CPUs and memory).
 	// +optional
 	VirtualMachine VirtualMachineSpec `json:"virtualMachine,omitempty"`
+	// Application specifies the settings for the Rancher Desktop Electron frontend.
+	// +optional
```
jandubois, line 215: The problem with optional substructures is that if the field is omitted, then the defaults in the substructure are not being applied. So you have to repeat them on the structure itself:

```suggestion
	// +optional
	// +kubebuilder:default={updates:{enabled:true},telemetry:{enabled:false}}
```

You can see we do the same thing for `ContainerEngine` already.

rdd/pkg/apis/app/v1alpha1/app_types.go
```diff
@@ -193,6 +222,9 @@ type AppSpec struct {
 	// virtualMachine specifies the VM resource allocation (CPUs and memory).
 	// +optional
 	VirtualMachine VirtualMachineSpec `json:"virtualMachine,omitempty"`
+	// Application specifies the settings for the Rancher Desktop Electron frontend.
```
jandubois: We always use the JSON spelling in the API doc string:

```suggestion
	// application specifies the settings for the Rancher Desktop Electron frontend.
```

This is also wrong in the `Namespace` comment above, if you want to correct it as well.

scripts/generate-rdd-client.ts
```diff
@@ -94,7 +94,7 @@ async function run() {
       proxy.kill();
     }
 
-    await execFile(rddPath, ['set', 'running=true']);
+    await execFile(rddPath, ['set', 'running=true', 'containerEngine.name=moby', '--wait']);
```
jandubois: `--wait` is the default.

```suggestion
    await execFile(rddPath, ['set', 'running=true', 'containerEngine.name=moby']);
```

Since you are reconfiguring the rdd instance, maybe use a custom instance?

```
process.env.RDD_INSTANCE ||= 'rdd-client-gen';
```

Not sure. 🤷

jandubois (changes requested):
Please remove the `telemetry` setting as discussed on Slack

scripts/generate-rdd-client.ts
mook-as: I was trying to be explicit, in case the defaults ever change. But it probably wouldn't…

rdd/docs/design/api_app.md
```diff
@@ -85,6 +90,12 @@ status:
     observedGeneration: 1
 ```
 
+- **spec.application.telemetry.enabled**: For use by the Electron front end, to control whether
+  telemetry collection and reporting is enabled.  Defaults to `false`.
```
jandubois: These lines need to be deleted.

jandubois (changes requested):
It looks like the `regenerate` has deleted not just the `telemetry` field but the whole `application`: https://github.com/rancher-sandbox/rancher-desktop-2/pull/559/commits/3ecac4f6973a9ec1e31f4bee1029aa510cb8514a.

mook-as:
Good catch about the bad generated code, thanks!

jandubois (changes requested):
The script runs `docker` subcommands. We need to either specify the `--context` or run them via `rdd run`. If the user already has a working docker endpoint, then `rdd set ...` will not change the default context.

Minor: Since we now run `rdd svc delete`, there will never be a cached image; we will rebuild it on every run.

jandubois (approved):
Thanks, LGTM

PR #560, opened 2026-07-14
RDA: Restore pref button

mook-as:
This restores the preferences button, and fixes the preferences dialog enough to show up.

It's enough to show the "enable Kubernetes" checkbox.

### DIFF
--- a/background.ts
+++ b/background.ts
@@ -21,6 +21,7 @@ import getWSLVersion from '@pkg/utils/wslVersion';
 import * as window from '@pkg/window';
 import '@pkg/main/rdd-ctl';
 import { closeDashboard, openDashboard } from '@pkg/window/dashboard';
+import { openPreferences } from '@pkg/window/preferences';
 
 // https://www.electronjs.org/docs/latest/breaking-changes#changed-gtk-4-is-default-when-running-gnome
 if (process.platform === 'linux') {
@@ -236,6 +237,14 @@ ipcMainProxy.on('dashboard-open', () => {
 
 ipcMainProxy.on('dashboard-close', () => {
   closeDashboard();
+});
+
+ipcMainProxy.on('preferences-open', () => {
+  openPreferences();
+});
+
+ipcMainProxy.on('preferences-close', () => {
+  window.getWindow('preferences')?.close();
 });
 
 mainEvents.on('dialog-info', (args) => {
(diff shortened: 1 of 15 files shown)


jandubois:
AI review at https://jandubois.github.io/rancher-desktop-2/20260716-131947-pr-560.html shows a bunch of issues. I can't tell if they are expected because the code is in the middle of the refactoring, or if some of them genuinely should be addressed in this PR.

mook-as:
Going through everything, they're pretty much just because the rest of the dialog isn't re-implemented yet, plus a couple harmless things.

> C1. RdCheckbox API change breaks every caller outside BodyKubernetes.vue

Yes, they will need to be (slowly) fixed as the various pages come back.  Since they're _already_ busted, I'm not too worried about fixing them in one go (that would make the PR too big to review).

> I1. Normal loading renders as "Unable to fetch preferences"

Yeah, since we no longer load prefs as a separate step.  I do not believe this would be an issue, but we can revisit if that turns out to be wrong.  Note that the suggested fix makes no sense: once we have the watched resource, we automatically have the preferences (because it's just part of the watched resource).

> I2. ModalBody drops the preferences prop its mapped components still require

None of the _actually loaded_ components require that prop, and they will need to be modified as they come back anyway.

> S1. Main-window nav reads the dialog's uncommitted-edit getter

The code is effectively the same (because the only time it would be different is when there are uncommitted prefs in the _main window_, which doesn't make any sense); the only difference is that it is now clearer what we're waiting for, instead of a confusing `app.spec` which isn't obvious why that might block the button.

> S2. Commented-out markup splits tags across comment boundaries

That would be fixed as the page is filled in.  Unfortunately, HTML does not have nested comments (very few languages do, mostly FP-focused), so this is an ugly workaround.

> S3. Lifecycle dispatches drop their rejections

Logging to the console (that users can't even open) isn't really a helpful fix; we will need to correct the error handling globally later, right now we're just trying to get things to at least work on the happy path.

> S4. severity prop is unreachable and its guard is redundant 

This will be fixed once the backend has some way to signal that some changes may require restarting the VM, etc.

> S5. Dead Transition import; v-if on the <transition> element itself

Not really worth amending.

> S6. The only new translation key belongs to a commented-out tab

Ah, because we won't have container engine switchers until containerd is fixed (I originally wrote the PR thinking that would be enabled). Mind leaving that in for now?

> S7. Alert nav-name type no longer covers its remaining producers 

That's just because they're not implemented yet.

jandubois (approved):
Thanks, LGTM

Issue #503, opened 2026-06-26
Run E2E tests in CI


jandubois:
This will be done in multiple pull requests:

- [x] #478 
- [ ] #563

mook-as:
I don't believe there's anything left to do here — the E2E tests now run in CI; the `fixme` tests are all because the features are missing.

PR #563, opened 2026-07-15
CI: Enable running E2E Tests

mook-as:
This enables running E2E tests in CI.  We only do this on Linux, because we use the mock controller everywhere so the host differences _should_ not matter.

Part of #503. Presumably we still want to add more tests to exercise the rest of the UI.

Potentially, we can adjust how the E2E tests work in the future to run on a packaged build.

### DIFF
--- a/.github/workflows/e2e-playwright.yaml
+++ b/.github/workflows/e2e-playwright.yaml
@@ -2,7 +2,7 @@ name: UI E2E Tests
 
 on:
   workflow_dispatch:
-  push:
+  push: # We limit this further in the check-paths job.
   pull_request:
 
 permissions:
@@ -10,6 +10,13 @@ permissions:
 
 jobs:
   check-paths:
+    # On push, on the upstream repository (`rancher-sandbox/*`), limit to `main`
+    # and `release-*` branches.  For forks or pull requests, run on all branches.
+    if: >-
+      github.event_name != 'push' ||
+      github.repository_owner != 'rancher-sandbox' ||
+      github.ref_name == 'main' ||
+      startsWith(github.ref_name, 'release-')
     uses: ./.github/workflows/paths-ignore.yaml
   e2e-tests:
     needs: check-paths
--- a/.github/workflows/e2e-playwright.yaml
+++ b/.github/workflows/e2e-playwright.yaml
@@ -1,0 +1,39 @@
+name: UI E2E Tests
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  check-paths:
+    uses: ./.github/workflows/paths-ignore.yaml
+  e2e-tests:
+    needs: check-paths
+    if: needs.check-paths.outputs.should-run == 'true'
+    timeout-minutes: 30
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/paths-ignore.yaml
```diff
@@ -62,12 +55,14 @@ jobs:
       if: github.event_name == 'pull_request'
       run: |
         MERGE_BASE=$(git merge-base $BASE $HEAD)
-        diff="$(git diff --name-only $MERGE_BASE $HEAD -- $PATHS_IGNORE)"
-        if [[ -z "$diff" ]]; then
+        if git diff --name-only $MERGE_BASE $HEAD \
+          | git -c core.excludesFile=${{ runner.temp }}/PATHS_IGNORE \
+            check-ignore --non-matching --no-index --verbose --stding \
```
jandubois: Typo alert:

```suggestion
            check-ignore --non-matching --no-index --verbose --stdin \
```

.github/workflows/e2e-playwright.yaml
```diff
@@ -0,0 +1,39 @@
+name: UI E2E Tests
+
+on:
+  workflow_dispatch:
+  push:
```
jandubois: Do we really want both `push` and `pull_request`? This is a long running job that will now run twice for each push to a PR. Maybe scope `push` to `main` and `release-*`?

.github/workflows/paths-ignore.yaml
```diff
@@ -39,20 +39,13 @@ jobs:
...
-        while read -r line; do
-          if [[ -n $line ]]; then
-            PATHS_IGNORE="${PATHS_IGNORE} :!/${line}"
-          fi
-        done <<< "$INPUT"
-        echo "$PATHS_IGNORE"
-        echo "$PATHS_IGNORE" >> "$GITHUB_ENV"
-      env:
-        INPUT: ${{ inputs.paths-ignore-globs }}
+      shell: cp {0} PATHS_IGNORE
```
jandubois: I think I know what this does, but maybe a comment would be good?

.github/workflows/paths-ignore.yaml
jandubois, line 1: Pre-existing:

* `./.github/workflows/actions/paths-ignore.yaml` should be `./.github/workflows/paths-ignore.yaml`
* `jobs.check-paths.outputs` should be `needs.check-paths.outputs`

jandubois (changes requested):

mook-as:
I believe everything is addressed now. New run: https://github.com/mook-as/rd2/actions/runs/29881489274/job/88803167693

e2e/containers.e2e.spec.ts
jandubois, line 532: Maybe add a reminder to enable `serial` mode again when the block gets un-fixme'd? You use the same in several other files (e.g. `lockedFields.e2e.spec.ts`).

```ts
  test.describe.configure({ mode: 'serial' });
```

jandubois (approved):
Thanks, LGTM

Should we have an issue tracking the `fixme`s?

```
  Running 121 tests using 1 worker
    115 skipped
    6 passed (18.8s)
```

e2e/containers.e2e.spec.ts
mook-as: I think it's fine to not have it, because if it ever runs in parallel things will fall over spectacularly. The fixmes will need to turn into serial everywhere, anyway.

mook-as (commented):
> Should we have an issue tracking the fixmes?

I don't think so at this point; they show up on `yarn run test:e2e` anyway as skipped things.

PR #567, opened 2026-07-18
dependabot: hold etcd at 3.6.x and move its group to /rdd

jandubois:
Dependabot's #564 bumped only `go.etcd.io/etcd/api/v3` to 3.7.0 and broke the build. Both kine and the embedded k8s apiserver still implement etcd's 3.6 gRPC API, so the whole module set is stuck on 3.6 until they adopt 3.7.

This holds etcd below 3.7 and groups the modules in the `/rdd` block — the only one Dependabot updates here — while still allowing 3.6.x patch bumps. It also reverts #565, which added the group to `/src/go` instead; that block mirrors RD1 to keep merges clean, and RD1 does not group etcd.

Once this merges, Dependabot will close #564 (3.7 becomes ignored).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -52,8 +52,6 @@ updates:
   - # Swagger dependencies must match whatever the CLI generate.
     dependency-name: "github.com/go-openapi/swag"
   groups:
-    etcd:
-      patterns: ["go.etcd.io/etcd/*"]
     golang-x:
       patterns: ["golang.org/x/*"]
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I was hoping [`allow`](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#allow--) would let us define dependency-specific rules, but it's not clear if that would prevent unlisted dependencies from being updated.

Issue #527, opened 2026-07-03
RDA: Implement updates


mook-as:
Updates are currently not possible in the app.  It needs:

- [ ] Implement a preference for updates
  - [ ] Fix the update UI to use the preference
- [ ] Test shipping app updates

PR #569, opened 2026-07-20
RDA: Restore update preferences

mook-as:
This restores the preferences for enabling the updater:

- The prefs dialog has the app prefs again
- The main window can now handle prefs that are applied immediately


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260720-153217-pr-569.html

I know the updater does not even exist yet, but I1 seems like an issue for all settings.

mook-as:
> I1. A failed immediate write leaves a stale value on screen and reports nothing

Fixed, hopefully.

> I2. Nothing wires the updater into the main process, so the restored main-window checkbox never renders

Yeah, that will need to happen later as part of #527

> I3. The new writeNow action and the immediate branch have no tests

Will add.

> S1. The writeNow comment and docstring describe something the window architecture forbids

That's a bug in the architecture; will fix in a focused follow-up PR.  The change state should be the same in all windows.

> S2. The revert/set block is duplicated verbatim between modify and writeNow

Fixed.

> S3. sudoAllowedTooltip is now dead, and the block that used it cannot be uncommented

Will be fixed once that comes back.

> S4. The commented-out <telemetry-opt-in> uses a prop the component does not have

Will be fixed once that comes back. I originally implemented it in this branch (as a pref that exists but does nothing), but Jan has convinced me it's better to have it once it's actually useful.

> S5. Store state destructured at setup rather than read through a computed

Fixed.

> S6. The navigate dispatch is neither awaited nor caught

Added a catch, and a comment referencing the issue.

> S7. tabSelected reads tab.name through a type that does not describe it

Fixed.

> S8. @/* added without resolveJsonModule

Fixed, and removed `experimentalDecorators` because we no longer have any.

> S9. One type-plumbing change is inert, the other deserves a comment

Changes split into its own commit, and comment added for the latter. The former is just mentioned in the new commit message.

> S10. The action payload type does not tie value to key

This ended up being too much of a rabbit hole.  Will leave it for now, and do a follow-up later.

> S11. An unrelated pre-existing test fix rides along in this PR

This seems too trivial for its own PR; it's already in its own commit.

> S12. Inconsistent key alignment in the reworked spec

Clearly a linter configuration issue; will be handled as part of fixing the linter configuration.

jandubois (approved):
Thanks, LGTM

Updated AI review https://jandubois.github.io/rancher-desktop-2/20260721-190147-pr-569.html has I1/I2, but they are not blockers for merging.

mook-as:
Yeah, I filed #574 to better holistically handle errors; at this point they're mostly recorded but never shown.

PR #571, opened 2026-07-20
RDA: Preferences: Add kube versions

mook-as:
This modifies the preferences dialog to allow selecting Kubernetes versions.

### DIFF
--- a/e2e/pages/preferences/kubernetes.ts
+++ b/e2e/pages/preferences/kubernetes.ts
@@ -5,7 +5,6 @@ export class KubernetesNav {
   readonly nav:                           Locator;
   readonly kubernetesToggle:              Locator;
   readonly kubernetesVersion:             Locator;
-  readonly kubernetesPort:                Locator;
   readonly kubernetesOptions:             Locator;
   readonly kubernetesVersionLockedFields: Locator;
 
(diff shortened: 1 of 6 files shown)


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260720-202335-pr-571.html

Check especially I1: the hardcoded `.k3s1` is wrong when there have been patch releases.

mook-as:
The `k3s1` thing is an existing thing since #278 (and we currently don't have a way to auto-update it either).  In fact, if it's _not_ `k3s1` it would fail to install:
https://github.com/rancher-sandbox/rancher-desktop-2/blob/86da1394fd71209147f49a050c4d1d1327714506/rdd/pkg/controllers/app/app/lima-template.yaml#L106

The correct thing would be a different PR to:
- Have the app validator accept explicit build versions in `.spec.kubernetes.version`
- Have the defaulter automatically convert any bare versions (`v1.23.4`) to with the build part (`v1.23.4+k3s99`)
- Make sure the full version is passed down to lima, so we don't need to do the weird thing in the install script

I plan to have a follow-up PR to have better E2E tests, with the versions filled in from the real k3s-versions reconciler (running under mock controller, without the validator).  That coverts I2, S2, and S4.

S3 is inane, because the app reconiler/validator ensures the values are valid.

The rest is trivial stuff that I may look at in a future PR, but it's not worth fixing right _now_.

jandubois (approved):
Thanks, LGTM

PR #577, opened 2026-07-22
RDA: E2E Tests for Kubernetes versions preferences

mook-as:
This adds E2E tests to the pref dialog for Kubernetes versions (see #571).

- We now load the real k3s-versions reconciler into the mock controller, so we get the correct format for the k3s-versions config map.
- This means we can test the pref dialog, comparing the expected versions (introspected from the config map) to the actually displayed items.

If we ever end up doing more complicated version fetching in the reconciler, we may need to duplicate it into a mock version with static assets instead; but that's not today.

### DIFF
--- a/e2e/preferences.e2e.spec.ts
+++ b/e2e/preferences.e2e.spec.ts
@@ -10,7 +10,7 @@ import { reopenLogs } from '@pkg/utils/logging';
 
 import type { ElectronApplication, Page } from '@playwright/test';
 
-let page: Page;
+let mainNav: NavPage;
 
 /**
  * Using test.describe.serial make the test execute step by step, as described on each `test()` order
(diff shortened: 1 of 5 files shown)


e2e/preferences.e2e.spec.ts
```diff
@@ -1,47 +1,48 @@
...
-    await new NavPage(page).preferencesButton.click();
-    preferencesWindow = await electronApp.waitForEvent('window', page => /preferences/i.test(page.url()));
+    mainNav = new NavPage(await electronApp.firstWindow());
   });
 
   test.afterAll(async({ colorScheme }, testInfo) => {
     await teardown(electronApp, testInfo);
   });
 
+  test('should finish loading', () => mainNav.waitForAppSettled());
```
mook-as, line 33: This fixes the current CI failures on `main`; sorry about that.

jandubois (approved):
Thanks, LGTM

But check the feedback in https://jandubois.github.io/rancher-desktop-2/20260722-141013-pr-577.html anyways

mook-as:
> I1. toContainText array form cannot detect extra or unexpected options

Intentional: there may be headings (recommended, etc.)

> I2. The expected labels re-derive the component's presentation logic, and the copies already differ 

Good idea, maybe later.

The rest is inane; we don't need to deal with them.

PR #578, opened 2026-07-22
Move host-switch from RDD controller to hostagent

### LINKED ISSUE #432: Move host-switch from the rdd process into host-agent
It is the natural place for it, so every host resource gets cleaned up automatically when the VM shuts down.

Nino-K:
The WSL2 host-switch virtual network switch (gvisor-tap-vsock bridge, AF_VSOCK handshake, and Docker socket bridge) ran as a goroutine in the long-lived RDD controller. As a result, its host resources outlived the VM and required explicit teardown (stopHostSwitch, unexposeAllForwards, orphan cleanup).

Move it into the per-VM hostagent process, where it shares the VM's lifecycle. When the hostagent exits, the OS automatically reclaims the vsock listeners, gvisor network, and host ports. If the bridge fails unexpectedly, the hostagent retries it internally (rate-limited) while keeping the VM running.

Fixes: https://github.com/rancher-sandbox/rancher-desktop-2/issues/432


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260722-123905-pr-578.html

jandubois (approved):
Thanks, LGTM

PR #584, opened 2026-07-23
Cap the BATS job at 120m so a wedged step fails fast

jandubois:
A `bats-app` job wedged in the Linux "Install prerequisites" step and, with no job-level timeout, would have held the runner until GitHub's 6h default. It was observed running for over 2h before being cancelled.

`timeout-minutes: 120` caps a hang in any setup or build step. The bats invocation itself is already bounded by `BATS_TIMEOUT` (60m), and the slowest real job (macos bats-app) runs ~90m all-in, so 120m leaves margin without false-failing slow runs.

The `wget` of jq had no timeout, so a stalled download could retry for a long time; `--timeout=30 --tries=3` bounds it. That is the most likely cause of this stall, though the in-progress job's logs were not available to confirm.


Nino-K (approved):
LGTM

Issue #581, opened 2026-07-23
HostInfo should not be visible on Windows


mook-as:
Because it should not be possible to set VM CPU / memory on Windows, the HostInfo data should not be available.  Alternatively, some other form of signal is required for the UI to know that the [Virtual Machine](https://docs.rancherdesktop.io/ui/preferences/virtual-machine/hardware/) tab of the preferences should not be displayed.

The "easy" thing to do would be to check the platform the UI is running on, but that has disadvantages:

- That makes it hard to test: we'd need to run tests on each platform in CI.  I think right now the `yarn install` actually dominates CI time…
- In the future, if we ever want to manage remote RDD instances, we'll need to know what platform that is running on anyway.

mook-as:
Hmm, or we could just have extra stuff hanging off `HostInfo` that says whether this host supports resizing VMs, etc.?

jandubois:
In #498 I wrote:

> I think long term we might add a HostInfo resource that can also monitor other metrics, like available memory or disk space, that can generate diagnostics, or be queried by the GUI. But for now let's keep it simple with a ConfigMap.

I don't know if that makes sense, or if we need some other way to collect metrics. Because we don't want to refresh e.g. free memory all the time if nobody is watching it.

But I could see us monitoring free disk space every couple of minutes and raising a diagnostic when disk space runs low on the device that contains the VM storage. Because that file can grow dynamically.

We may also offer different optional drivers for Windows in the future. E.g. QEMU already works with Lima on Windows, and support for HCS/Hyper-V is in progress. Both should support setting CPU and memory. So the inability to set cpu/memory is a property of the os/driver combo, and not the os alone.

But for now let's not hide `HostInfo` on Windows. We can still display the host settings, even when the user cannot pick the settings for the VM.

PR #585, opened 2026-07-23
RDA: Preferences: VM (memory & CPU)

mook-as:
This restores the part of the preferences dialog that adjusts how much CPU & memory are used by the VM.

Per #581, for now we're showing this part even on Windows.  This simplifies the E2E somewhat; we just always test it exists.

As usual, I have tries to make each commit a bit isolated.

### DIFF
--- a/e2e/pages/preferences/index.ts
+++ b/e2e/pages/preferences/index.ts
@@ -1,4 +1,4 @@
-import { expect, Page } from '@playwright/test';
+import { expect, Page, Locator } from '@playwright/test';
 
 import { ApplicationNav } from './application';
 import { ContainerEngineNav } from './containerEngine';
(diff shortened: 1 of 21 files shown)


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260723-155313-pr-585.html

Check out I1, which reports a crash when memory size is set to a fractional value, like `2.5GiB`.

This leaves us with an interesting questions: what do we do with them? If the user never tries to modify them, we should leave them as-is. But if they change them, then only the slider stops are available. This is fine for now, but we may revisit this in the future (e.g. providing a logarithmic scale on machine with a large amount of RAM).

mook-as:
Oh, I didn't realize `2.5Gi` is a valid value. Will fix.

jandubois (approved):
Thanks, LGTM

Latest AI review: https://jandubois.github.io/rancher-desktop-2/20260724-172401-pr-585.html

It mentions additional edge cases in parsing Kubernetes quantities (memory sizes), but I think we can leave that to a follow-up PR.

Same for the remaining suggestions in the report.

PR #589, opened 2026-07-24
Linux: ship chrome-sandbox setuid root in the RPM and DEB

### LINKED ISSUE #588: chrome-sandbox is missing from the RPM and not setuid in the DEB
### Actual Behavior

The Linux packages mishandle Chromium's setuid sandbox helper. The RPM leaves `/opt/rancher-desktop-2/chrome-sandbox` out of the package entirely, and the DEB ships it without the setuid bit. Rancher Desktop 1 ships the same file at mode 4755.

### Steps to Reproduce

Against packages published to `isv:Rancher:stable` and `isv:Rancher:dev`:

```
rpm --query --list --package rancher-desktop-2-2.0.0alpha.1-lp156.2.1.x86_64.rpm | grep chrome-sandbox
dpkg-deb --contents rancher-desktop-2_9.main.1784841277.37b1ca371-71.1_amd64.deb | grep chrome-sandbox
rpm --query --list --verbose --package rancher-desktop-1.23.1-lp156.3.8.x86_64.rpm | grep chrome-sandbox
```

### Result

| package | `chrome-sandbox` |
|---|---|
| RD1 RPM 1.23.1 | `-rwsr-xr-x` |
| RD2 RPM 2.0.0alpha.1 | absent from the package |
| RD2 DEB from `main` @37b1ca371 | `-rwxr-xr-x` |

### Expected Behavior

Both RD2 packages ship `/opt/rancher-desktop-2/chrome-sandbox` owned by root at mode 4755, as RD1 does.

### Additional Information

`%files` in `packaging/linux/rancher-desktop.spec` ends with:

```
/opt/%{name}*
%exclude /opt/%{name}/chrome-sandbox
%attr(4755,root,root) /opt/%{name}/chrome-sandbox
```

The glob picks the file up with default permissions and the `%attr` line then wants to ship it setuid, which produced a "file listed twice" warning. 69dd9e5a4 added `%exclude` to silence that warning. But `%exclude` applies to the finished file list, so it drops the `%attr` entry as well and the RPM loses the file. debbuild drops only the `%attr` entry, which is why the DEB keeps the file at 755.

Removing the `%exclude` line restores RD1's arrangement, where the later `%attr` entry overrides the glob. I have not built and tested that yet.

I found this while testing #509 and have not checked whether the app fails to launch without the helper. Modern kernels let Chromium fall back to the user-namespace sandbox, and the repo smoke test writes `0` to `/proc/sys/kernel/apparmor_restrict_unprivileged_userns` before running, which would hide a missing helper. That is a guess at why it went unnoticed, not a tested claim.

### Rancher Desktop Version

2.0.0-alpha.1, and `main` at 37b1ca371

### Rancher Desktop K8s Version

N/A

### Which container engine are you using?

N/A

### What operating system are you using?

Other Linux

### Linux only: what package format did you use to install Rancher Desktop?

rpm and deb

---

Report by Claude Opus 5

jandubois:
The `%exclude` that silenced the "file listed twice" warning also dropped chrome-sandbox from the RPM. The `%attr` line never worked for the DEB either, because debbuild ignores it once the glob has picked the file up. Setting the mode in `%install` covers both builders and keeps the file list free of duplicates.

OBS builds the real packages and CI never does, so the guard against a repeat goes into the release smoke test, which installs from there.

| variant | Leap 15.6 | Fedora 42 | Tumbleweed | debbuild |
|---|---|---|---|---|
| current | absent | absent | absent | 755 |
| `%exclude` dropped | 4755 | 4755 | 4755 | 755 |
| this PR | 4755 | 4755 | 4755 | 4755 |

Dropping `%exclude` alone, as the issue suggested, fixes only the RPM. RD1's DEB confirms that, since `rancher-desktop_1.23.1-3.1_amd64.deb` from OBS ships chrome-sandbox at 755. This PR leaves RD1 alone.

Fixes #588

### DIFF
--- a/packaging/linux/rancher-desktop.spec
+++ b/packaging/linux/rancher-desktop.spec
@@ -184,6 +184,10 @@ cp -ra ./share "%{buildroot}%{_prefix}"
 rm -rf ./share
 cp -ra ./* "%{buildroot}/opt/%{name}"
 
+# Chromium's sandbox helper must be setuid root. Set the mode here because
+# debbuild ignores the attributes given in the file list.
+chmod 4755 "%{buildroot}/opt/%{name}/chrome-sandbox"
+
 # Link to the binary
 ln -sf "/opt/%{name}/%{name}" "%{buildroot}%{_bindir}/%{name}"
 
@@ -196,8 +200,6 @@ true
 %defattr(-,root,root,-)
 %dir /opt/%{name}
 /opt/%{name}*
-%exclude /opt/%{name}/chrome-sandbox
-%attr(4755,root,root) /opt/%{name}/chrome-sandbox
 %{_bindir}/%{name}
 %{_prefix}/share/applications/%{name}.desktop
 %{_prefix}/share/icons/hicolor/*
--- a/.github/workflows/smoke-test/install-from-repo.sh
+++ b/.github/workflows/smoke-test/install-from-repo.sh
@@ -68,12 +68,24 @@ install_linux_fedora() {
     dnf --assumeyes install "${PACKAGE_NAME}-${version}"
 }
 
+# Chromium only uses its setuid sandbox when the helper is setuid root, so a
+# packaging mistake here degrades the sandbox silently.
+verify_sandbox_helper() {
+    local sandbox="/opt/${PACKAGE_NAME}/chrome-sandbox" mode
+    mode=$(stat --format='%a %U' "$sandbox")
+    if [[ "$mode" != "4755 root" ]]; then
+        printf "%s is '%s', expected '4755 root'\n" "$sandbox" "$mode" >&2
+        exit 1
+    fi
+}
+
 main() {
     RD_VERSION=$(grep --only-matching '\([0-9]\+\.[0-9]\+\)' <<< "$RD_VERSION")
     source /etc/os-release
     for id in ${ID:-} ${ID_LIKE:-}; do
         if [[ "$(type -t "install_linux_$id")" == function ]]; then
             eval "install_linux_$id"
+            verify_sandbox_helper
             exit 0
         fi
     done


packaging/linux/rancher-desktop.spec
```diff
@@ -184,6 +184,10 @@ cp -ra ./share "%{buildroot}%{_prefix}"
 rm -rf ./share
 cp -ra ./* "%{buildroot}/opt/%{name}"
 
+# Chromium's sandbox helper must be setuid root. Set the mode here because
+# debbuild ignores the attributes given in the file list.
+chmod 4755 "%{buildroot}/opt/%{name}/chrome-sandbox"
```
mook-as, line 189: Ah, because `%defattr(-, …)` means it doesn't override the file mode, so setting it early does the correct thing.

That means we must have `#needsrootforbuild` somewhere? I don't see it though?

mook-as (commented):
Do you have sample logs from building the RPM?

This can be done by:

1. Make a derived branch, make it upload to something like GitHub Pages, e.g. https://github.com/rancher-sandbox/rancher-desktop/commit/94661c63d239dba547b2d29788c5e6258a86e314
1. Set up a branch of the OBS project
1. Configure its [`_service`](https://build.opensuse.org/projects/home:mook_work:branches:isv:Rancher:dev/packages/rancher-desktop-main/files/_service?expand=1) file to grab it from the page in step 1.

packaging/linux/rancher-desktop.spec
jandubois: No, you don't need to be `root` to set `setuid` on your own files.

```
  ┌────────────────────────┬──────┬──────────────────────┐
  │        builder         │ uid  │    packaged mode     │
  ├────────────────────────┼──────┼──────────────────────┤
  │ rpmbuild, Leap 15.6    │ 1000 │ -rwsr-xr-x root root │
  ├────────────────────────┼──────┼──────────────────────┤
  │ rpmbuild, Fedora 42    │ 1000 │ -rwsr-xr-x root root │
  ├────────────────────────┼──────┼──────────────────────┤
  │ debbuild, Ubuntu 24.04 │ 1001 │ -rwsr-xr-x root/root │
  └────────────────────────┴──────┴──────────────────────┘
```

jandubois:
Here are some log from an rpm build in a VM (with a slightly hacked up setup because the linux.zip still had the old names): [rpmbuild-after.log](https://github.com/user-attachments/files/30365519/rpmbuild-after.log)

mook-as (approved):
Ah, sorry about that.
I missed that the `chmod` is still running against a file owned by the current user (in CI, `abuild`); `%defattr(-, root, root, -)` sets the owner without touching permissions, so it ends up as `o4755` but owned by root.

PR #595, opened 2026-07-28
RDA: Follow-up to #585

### LINKED ISSUE #590: GUI doesn't parse some valid memory quantities correctly
Numbers with leading or trailing dot: `5.GiB` or `.5GiB`.

See https://jandubois.github.io/rancher-desktop-2/20260724-172401-pr-585.html

mook-as:
This fixes additional review comments from #585.  In particular:
- Makes the quantity parsing (currently used for the VM memory spec) more conformant to Kubernetes.
- Fixes the issues with kebab casing; we now use a custom version to have correct typing.

Fixes #590

### DIFF
--- a/pkg/rancher-desktop/components/Preferences/ModalNav.vue
+++ b/pkg/rancher-desktop/components/Preferences/ModalNav.vue
@@ -1,6 +1,4 @@
 <script lang="ts" setup>
-import kebabCase from 'lodash/kebabCase';
-
 import NavItem from '@pkg/components/Preferences/ModalNavItem.vue';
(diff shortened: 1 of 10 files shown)


jandubois (dismissed):
Thanks, LGTM

From https://jandubois.github.io/rancher-desktop-2/20260730-180434-pr-595.html I think I1 may deserve some attention, the rest seems more a matter of opinion.

Also #596 should be merged first.

jandubois (approved):
Thanks, LGTM

FYI: https://jandubois.github.io/rancher-desktop-2/20260731-170545-pr-595.html

PR #596, opened 2026-07-29
Merge RD1 into RD2

mook-as:
This is the next set of merges from RD1 to RD2, pulling in i18n and blog feed, both of which are currently broken :D

I'll do a follow-up to fix the feed shortly; it's not in this PR to make it easier.


jandubois (approved):
Thanks, LGTM

There are a bunch of issues in https://jandubois.github.io/rancher-desktop-2/20260730-171224-pr-596.html, but I think they can be addressed in follow-up PRs instead of squeezing them into the merge PR.

PR #599, opened 2026-07-29
CI: Package: Fix OBS package name

mook-as:
It tried to trigger a `rancher-desktop-2-release-2.0.0-alpha.2` package, which does not exist.

Note that this is rather untested; I'll try to figure out how to test in a bit.


jandubois (approved):
Thanks, LGTM

PR #601, opened 2026-07-30
CI: Linux-release: Use vault for secrets

mook-as:
I don't actually know if this is needed; it's not like I can test that in a fork… So this PR is here if we need it.  This also means the PR is untested :(

Example release zip name: `rancher-desktop-linux-v2.0.0-alpha.1.zip`
Example S3 zip name: `rancher-desktop-2-linux-2.0.zip`
Example package name: `rancher-desktop-2-2.0`


jandubois (approved):
Thanks, LGTM

PR #602, opened 2026-07-30
Restore the DEB chrome-sandbox setuid bit and follow up on #601

jandubois:
The DEB shipped chrome-sandbox at 755, and the first commit restores the setuid bit. The other two follow up on #601. One moves the vault step above the download so a throwaway tag can exercise it without publishing a release; the other records why the OBS zip name and `--fail-with-body` have to stay.

The spec fix ships without a new tag. The stable package takes its spec from the release branch, and its version from `@PARENT_TAG@`.

Publishing still won't rebuild stable. All five repositories in `isv:Rancher:stable` have the `rancher-desktop-2-2.0` package disabled, so someone has to enable the builds first.

### DIFF
--- a/packaging/linux/rancher-desktop.spec
+++ b/packaging/linux/rancher-desktop.spec
@@ -192,9 +192,11 @@ chmod 4755 "%{buildroot}/opt/%{name}/chrome-sandbox"
 ln -sf "/opt/%{name}/%{name}" "%{buildroot}%{_bindir}/%{name}"
 
 %post
-# This is needed to ensure Debian packages have proper file permissions;
-# otherwise the postinst script is not generated correctly.
-true
+# debbuild's generated postinst chowns /opt/%{name} recursively, and chown
+# clears the setuid bit that %install set. %post runs after those chowns, so
+# the mode set here sticks. rpm applies the mode from the package header, so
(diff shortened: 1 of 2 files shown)


packaging/linux/rancher-desktop.spec
```diff
@@ -192,9 +192,11 @@ chmod 4755 "%{buildroot}/opt/%{name}/chrome-sandbox"
...
 
 %post
-# This is needed to ensure Debian packages have proper file permissions;
-# otherwise the postinst script is not generated correctly.
-true
+# debbuild's generated postinst chowns /opt/%{name} recursively, and chown
+# clears the setuid bit that %install set. %post runs after those chowns, so
+# the mode set here sticks. rpm applies the mode from the package header, so
+# the chmod is a no-op there.
+chmod 4755 /opt/%{name}/chrome-sandbox
```
mook-as, line 199: Well, no, it's because we stopped doing `%attr(4755,root,root)` in the `%files` section in 02cedd3b6f6a77396505d053cb068460b512b610.  But I can't think of a better solution; other than perhaps generating a file list and using `%files -f`, which has its own problems.

mook-as (approved):
I have, quite obviously, not tested this.

PR #606, opened 2026-07-31
RDA: E2E: Restore some container tests

mook-as:
This reinstates some of the container E2E tests, mostly around the list and logs.


jandubois (approved):
Thanks, LGTM

FYI only: https://jandubois.github.io/rancher-desktop-2/20260731-114648-pr-606.html

PR #609, opened 2026-07-31
Follow up for #596

mook-as:
This addresses some of the comments from https://jandubois.github.io/rancher-desktop-2/20260730-171224-pr-596.html

Note that the stuff around i18n is mostly unaddressed; another PR will enable it and then we can figure out what's broken there.

~This is a draft because it depends on #595 (see last commit).  I do not expect it to even build right now.~

This is now finished, because #595 has been merged.


pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -69,7 +69,7 @@ mainMenu:
   quit: Quit {appName}
 tray:
   # @no-translate Rancher Desktop
-  tooltip: Rancher Desktop
+  tooltip: {appName}
```
Nino-K: Does this need quotes like `tooltip: '{appName}'` because it would parse as a YAML map `({appName: null})`?

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -69,7 +69,7 @@ mainMenu:
   quit: Quit {appName}
 tray:
   # @no-translate Rancher Desktop
```
Nino-K: This comment is stale now.

pkg/rancher-desktop/assets/translations/en-us.yaml
```diff
@@ -89,7 +89,7 @@ tray:
   # @no-translate Kubernetes
   kubernetesContexts: Kubernetes Contexts
   # @no-translate Rancher Desktop
```
Nino-K: This one too, stale comment.

Nino-K (approved):

PR #617, opened 2026-08-03
RDD: HostInfo: report memory as a resource quantity

jandubois:
`rdd ctl get hostinfo` printed 137438953472 instead of 128Gi. Kubernetes uses resource.Quantity for every memory size, from pod limits to node capacity, so status.memory now holds one; spec.virtualMachine.memory already did.

Quantities canonicalize to the largest suffix that divides exactly, so a host whose RAM isn't a whole number of GiB shows Ki. Linux and Windows usually report such a total; macOS reports the exact physical size.

The second commit is the mechanical `yarn generate:rdd-client` output. It types `memory` as `any`, the same type `spec.virtualMachine.memory` already generates.

```console
❯ rdd ctl get hostinfo
NAME     CPUS   MEMORY
system   20     137438953472
```

Should be

```console
❯ rdd ctl get hostinfo
NAME     CPUS   MEMORY
system   20     128Gi
```

---

_Sighs_

AI now tells me that I already have a pending PR that implements the same thing: #558

I guess splitting the change out is not a bad idea, but I wish I had checked earlier...


rdd/pkg/controllers/mock/crd.yaml
mook-as, line 1: Oh, this wasn't marked as a generated file (it shows up in the PR diff view)

mook-as (approved):

Issue #426, opened 2026-02-23
Restore feature: Preferences


mook-as:
Preferences will be disabled in rancher-sandbox/rancher-desktop-app#19; this the tracking task to re-implement it.

The initial implementation may just include
- Switching container engines (moby/containerd)
- Enable/Disable Kubernetes
- Select Kubernetes version

If that is the case, we need to file further issues to restore the rest.

PR #623, opened 2026-08-06
RDA: E2E: Prefs: locale switching

mook-as:
This implements E2E tests for the prefs dialog on switching languages.  It switches to Korean, and then tests that the _General_ tab turns into purely Hangul text.

Goes on top of #622.

### DIFF
--- a/e2e/preferences.e2e.spec.ts
+++ b/e2e/preferences.e2e.spec.ts
@@ -40,48 +40,61 @@ test.describe('Preferences Dialog', () => {
     await prefPage.waitForLoad();
   });
 
-  test.fixme('should show application page and render general tab', async() => {
-    const { application } = prefPage;
+  test.describe('Application', () => {
+    test('should show application page and render general tab', async() => {
+      const { application } = prefPage;
 
-    await expect(application.nav).toHaveClass('preferences-nav-item active');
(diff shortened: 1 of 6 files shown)


e2e/preferences.e2e.spec.ts
```diff
@@ -40,48 +40,61 @@ test.describe('Preferences Dialog', () => {
...
+      const { application } = prefPage;
 
-    if (!os.platform().startsWith('win')) {
-      await expect(application.tabEnvironment).toBeVisible();
-    } else {
-      await expect(application.tabEnvironment).not.toBeVisible();
-    }
+      await expect(application.nav).toHaveClass('preferences-nav-item active');
+      await expect(application.tabGeneral).toHaveText('General');
+      // TODO: environment tab visiblity.
```
Nino-K, line 49: Should there be an issue that includes all these todos?

Nino-K (approved):
Looks good, I think we just need an issue that includes the "TODO" items.

e2e/preferences.e2e.spec.ts
mook-as: That happens when that tab _exists_. I think that would be somewhere under #426 (but #638 is definitely relevant…)

PR #624, opened 2026-08-06
RDA: Prefs: Fix incrementing VM memory by arrows

### LINKED ISSUE #616: cpu/memory spinners are not working properly
The memory spinner can be clicked once to decrement memory from 6GB to 5GB. It displays an unexpected an incorrect validation message. Clicking the spinner again gets the dialog to hang, but you can still cancel it.

Image

Doing the same for CPUs also is not correct: Decrementing the `2` briefly flashes a `1` but then clears the field. Decrementing it again goes to `-1`, but no validation error is displayed.

mook-as:
When increasing the memory via the spinners in the text field, we ended up getting `NaNGi` because the change event emitted the actual event object instead of the value.  By declaring the emit, Vue does enough magic to just emit the value instead, which fixes the issue.

This includes E2E tests to ensure that stays the case.

Fixes #616.


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260911-181019-pr-624.html

Issue #625, opened 2026-08-06
Spike: GPU Passthrough for Windows


gaktive:
### Problem Description

GPU Passthrough is gaining some interest, so we should look into what it'll take to run on it.

Windows should be the 1st platform to tackle, which will require some research. Whether this becomes an epic or not, I'll leave to @Nino-K. One key area of focus is an install mechanism for two types of toolkits

### Proposed Solution

Make GPUs work.

PR #634, opened 2026-08-12
AMD ROCm GPU Passthrough

Nino-K:
This PR is driven by all the findings that are documented here: https://gist.github.com/Nino-K/f240bab7059acaaf9c5b59f6e2b33a89

This adds experimental support for using an AMD GPU from workloads running in the RDD VM on Windows/WSL2. It targets ROCm on WSL, where the GPU comes through `/dev/dxg` rather than the usual `/dev/kfd` path as it would on native Linux.

ROCm-on-WSL needs two files that don't ship anywhere convenient: `librocdxg.so` (the bridge/shim that talks to `/dev/dxg`) and its `dids.conf` device-ID table. WSL doesn't project them, they're not in the Windows driver store, and they're not baked into ROCm container images, so a workload pod has to bind-mount them from the guest. An alternative to this PR would be getting them into the VM via a manual copy from a separate ROCm-on-WSL install, which isn't something we can ask users to do.

## What this does

Turning on `gpu.enabled` now handles the whole thing:

* The app controller downloads AMD's pinned [rocdxg](https://github.com/ROCm/librocdxg) package from its GitHub release, checks the SHA-256, and extracts the two files into a host staging directory(`$HOME\AppData\Local\rancher-desktop-2\gpu`). A version marker makes it idempotent, if the files are already current, it skips the download entirely.
* Lima provisioning installs them into the guest's `/opt/rocm` on boot, and removes them again when the feature is turned off, so the toggle is symmetric.
* For air-gapped hosts (or people who already have a ROCm-on-WSL install), `gpu.source` points to a local directory instead of fetching, this can also be considered a way to override the pinned release version.

Everything is gated to Windows hosts and stays a no-op elsewhere. There's a design doc under `docs/design/gpu_passthrough.md` with the details, and the App CRD/schema picks up the new GPU fields (visible in `rdd set --help`).

## Reproducible example

The second commit adds `rdd/examples/vllm-gfx1151-repro/` which is driven from [vllm](https://apps.rancher.io/applications/vllm), which is limited to NVIDIA's CUDA only. This self-contained example modifes the original docker that is used by the helm chart and runs vLLM on an AMD Strix Halo iGPU (gfx1151 / Radeon 8060S) through the passthrough above. It has a small Dockerfile over AMD's prebuilt gfx1151 vLLM image, the pod spec with the `/dev/dxg` mount contract, a PowerShell verify script, and a README.

The one important piece is a tiny `amdsmi` shim: vLLM's ROCm platform detection calls into [amdsmi](https://github.com/ROCm/rocm-systems/tree/develop/projects/amdsmi), which reads GPU topology from `/sys/class/kfd` absent under WSL, so it's shadowed with a stub that reports the single dxg GPU. Compute still runs through torch/HSA, which works fine over `/dev/dxg`. The README explains the why, along with the socat/port-forward workaround.

## Testing

Validated end to end on a real Ryzen AI Max+ 395 box:

```
1. Start the daemon and enable Kubernetes

   rdd start
   rdd set kubernetes.enabled=true
   rdd kubectl get nodes          

2. Confirm the new GPU properties exist in the CRD

   rdd set --help                   # expect: gpu.enabled and gpu.source listed

3. Enable GPU passthrough

   rdd set gpu.enabled=true         # this blocks until reconcile comes back

4. Verify the host staged the fetched files

   dir  $HOME\AppData\Local\rancher-desktop-2\gpu
   type $HOME\AppData\Local\rancher-desktop-2\gpu\.rocdxg-version

   Expect: librocdxg.so , dids.conf, .rocdxg-version = 1.2.2.

5. Verify provisioning installed them into the guest

   wsl -d lima-rd -- ls -l /opt/rocm/lib/librocdxg.so /opt/rocm/share/rocdxg/dids.conf

   Expect: librocdxg.so -rwxr-xr-x (0755), dids.conf -rw-r--r-- (0644).

6. Run a real GPU workload (vLLM repro)

   cd rdd\examples\vllm-gfx1151-repro
   docker build -t vllm-gfx1151:local .        # ~10 GB base image, so expect this to take a while
   rdd kubectl apply -f vllm-pod.yaml
   rdd kubectl logs -f vllm-gfx1151            # wait for "Application startup complete"
   .\test.ps1                                                  

    Note: If running the test.ps1 gives some sort of execution error, run with:
    powershell -ExecutionPolicy Bypass -File .\test.ps1

   Expect: /v1/models returns the model; /v1/completions returns coherent text (e.g. "... Paris ..."); server log shows a generation throughput line with GPU KV cache usage (proves it ran on the GPU, not CPU, that is an important distinction to pay attention to).

7. Verify the disable path

   rdd set gpu.enabled=false
   wsl -d lima-rd -- ls -l /opt/rocm/lib/librocdxg.so /opt/rocm/share/rocdxg/dids.conf   # expect: No such file
   dir $HOME\AppData\Local\rancher-desktop-2\gpu                                          # expect: empty dir

8. Verify idempotent re-enable

   rdd set gpu.enabled=true         # re-stages; version marker means no re-download if already current

9. Cleanup

   rdd kubectl delete -f rdd\examples\vllm-gfx1151-repro\vllm-pod.yaml
```

## Notes

* Experimental, AMD/ROCm-on-WSL only for now.
* `librocdxg` is pinned to v1.2.2 (ROCm 7.2.x series); bump `Version` + `DebSHA256` together to move it.
* Think about multiple GPUs and how they show up in WSL? (e.g `/dev/dxg0`, `/dev/dxg1`)
* Think about CDI, the AMD's [container toolkit](https://github.com/rocm/container-toolkit) could offer the CDI, however it only works on native linux and, not WSL, and the only supported distros are Ubuntu/RHEL + Docker 25+. Furthermore, this project seems very new and not as battle-tested as Nvidia's container toolkit.
* https://github.com/rancher-sandbox/rancher-desktop-2/issues/635
* https://github.com/rancher-sandbox/rancher-desktop-2/issues/636

Related to: https://github.com/rancher-sandbox/rancher-desktop-2/issues/625


rdd/docs/design/gpu_passthrough.md
```diff
@@ -0,0 +1,134 @@
...
+   them together to move to a newer release. Staging is idempotent — a version
+   marker means an already-staged directory does no network I/O.
+2. **Source override:** if `spec.gpu.source` points at a host directory holding
+   the two files, they are copied from there instead (for air-gapped hosts or an
+   existing ROCm-on-WSL install).
+
+Provisioning (a `mode: system` step in `lima-template.yaml`, driven by the
+`GPU_ENABLED` / `GPU_LIB_SOURCE` params) then installs the staged files into the
+guest's `/opt/rocm/lib/librocdxg.so` and `/opt/rocm/share/rocdxg/dids.conf` on
+every boot. When passthrough is disabled the same step **removes** them, so the
```
mook-as, line 48: Do we really need to remove the files? It just sits in our VM (WSL distro), right?

Not really against it, though; I guess it's fine for symmetry.

rdd/examples/vllm-gfx1151-repro/amdsmi.py
mook-as, line 1: Looking at [`librocdxg` releases](https://github.com/ROCm/librocdxg/releases), they provide a `amd-smi` implementation as of version 1.2.1; the commit messages say they use python `.pth` magic to ensure it's loaded before the native version.  Maybe we can use that instead of the stub?

Ah, I see you mentioned it at the bottom of `gpu_passthrough.md`.

rdd/pkg/controllers/app/app/lima-template.yaml
```diff
@@ -156,6 +156,40 @@ provision:
 
     # rancher-desktop.target is restarted by the EXIT trap above.
 
+# Stage or remove the AMD ROCm-on-WSL bridge files based on GPU passthrough
+# state. librocdxg.so + dids.conf ship only with ROCm-on-WSL; they are not
+# projected by WSL, not in the Windows driver store, and not in ROCm container
```
mook-as, line 161: _projected by WSL_? Oh, you mean it's not automatically mounted by WSL, like `/usr/lib/wsl/lib`?

I wonder why they don't, since the nvidia drivers do… anyway, please rephrase? Maybe _… they are not automatically mounted by WSL,_?

rdd/pkg/gpu/gpu.go
```diff
@@ -0,0 +1,311 @@
...
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// Pinned release of https://github.com/ROCm/librocdxg. librocdxg must be
+// ABI-compatible with the ROCm userspace in the workload image; v1.2.2 targets
+// the ROCm 7.2.x series. Bump Version and DebSHA256 together when moving to a
```
mook-as, line 42: Per the chart at https://github.com/ROCm/librocdxg#wsl-compatibility-matrix — v1.2.2 targets ROCm 7.14.  I _hope_ it's backwards-compatible with ROCm 7.2.x, but that table is a bit vague on that point.

rdd/pkg/gpu/gpu.go
```diff
@@ -0,0 +1,311 @@
...
+	if dids == nil {
+		return nil, nil, fmt.Errorf("%s not found in package", DidsName)
+	}
+	return lib, dids, nil
+}
+
+// arMember returns the contents of the named member of a Unix ar archive. The
+// format is a global magic followed by fixed 60-byte member headers; only the
+// name and size fields are needed here.
+func arMember(archive []byte, name string) ([]byte, error) {
```
mook-as, line 264: Ouch. Yeah, I think we should at least build rocdxg by ourselves, and output a tarball instead of a deb or something.

That actually looks… sort of doable, on GitHub infra; just need to fork the repo.

mook-as (commented):
I think for actual shipping (not just a PoC), it would be better to simply ship `librocdxg` as part of Rancher Desktop instead of download on demand; that is, either as Jan's planned overlay, or just as part of the openSUSE distro. (It doesn't look like there's a `librocdxg` on OBS, so we may need to build from source, so that kind of sucks… and leans more towards the overlay).  librocdxg + amd-smi-wsl is like, 3MB tgz. ~And 1.22 no longer ships with their non-MIT blob anyway, despite what the license file says…~ I don't know how I missed the file that was still there.

In general, I think that's a good start, though I'm not sure this is meant to be merged; I'll leave part alone.

mook-as:
FWIW, I now have a [test GitHub Actions workflow](https://github.com/ROCm/librocdxg/compare/develop...mook-as:librocdxg:develop) to build librocdxg from source in CI, which emits two tar files (librocdxg, and amdsmi-wsl).  It can be trivially changed to RPM if we prefer it. Not sure that's useful…

----

I have managed to run this on the exact hardware (as in, the same physical box) you used.
I can see that, in task manager, _Dedicated GPU memory_ is fully used, and _Shared GPU memory_ is about 2/3rds used.  The GPU graph even shows _Compute 0_ has usage.  However, the output still looks like:
> (APIServer pid=1) INFO 08-19 21:05:53 [loggers.py:259] Engine 000: Avg prompt throughput: 0.3 tokens/s, Avg generation throughput: 18.5 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%
> (APIServer pid=1) INFO 08-19 21:06:03 [loggers.py:259] Engine 000: Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 0.0 tokens/s, Running: 0 reqs, Waiting: 0 reqs, GPU KV cache usage: 0.0%, Prefix cache hit rate: 0.0%

(For some reason both _It was a_ and _My name is_ seems to generate some kind of grade school comprehension test…)

I can confirm that the pod contains the files in `/opt/rocm`.  On the Lima VM, `lspci -k` does show devices using `dxgkrnl` is loaded.  There is a `/dev/dxg` in the pod container.

Possibly relevant log lines in the vllm output:
> (EngineCore pid=285) WARNING 08-19 20:52:02 [interface.py:525] Using 'pin_memory=False' as WSL is detected. This may slow down the performance.
> (EngineCore pid=285) INFO 08-19 20:52:04 [rocm.py:496] Using ROCM_ATTN backend out of potential backends: ['ROCM_ATTN', 'TRITON_ATTN'].
(EngineCore pid=285) INFO 08-19 20:52:54 [gpu_worker.py:436] Available KV cache memory: 91.39 GiB
(EngineCore pid=285) INFO 08-19 20:52:54 [kv_cache_utils.py:1319] GPU KV cache size: 7,985,856 tokens
(EngineCore pid=285) INFO 08-19 20:52:54 [kv_cache_utils.py:1324] Maximum concurrency for 4,096 tokens per request: 1949.67x
(EngineCore pid=285) WARNING 08-19 20:53:19 [chunked_prefill_paged_decode.py:398] Cannot use ROCm custom paged attention kernel, falling back to Triton implementation.
(APIServer pid=1) WARNING 08-19 20:53:26 [interface.py:525] Using 'pin_memory=False' as WSL is detected. This may slow down the performance.

jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260904-142758-pr-634.html

Also needs a rebase

Issue #497, opened 2026-06-26
Implement "path management" in the App controller


jandubois:
We need a new field `application.addPath`.

Possible values are `front`, `back`, and `manual`.

`front` and `back` designate where `~/.rd2/bin` should be added to the `PATH`: either `~/.rd2/bin:$PATH` or `$PATH/~/.rd2/bin`. `manual` means not adding it at all.

This setting applies to all OSes, including Windows.

Look at the RD1 implementation for the details on which files to edit (`.bash_profile`, `.zshrc`), 

#547 implements atomic file update functions.

To be discussed: The default setting will be `manual` when the App instance is created by the CLI, but `front` when created by the GUI. In RD1 the default was `front` on macOS/Linux, and `back` on Windows (by the installer, not configurable).

gaktive:
@mook-as to check if we have a PR or other ticket for the UI component here.

PR #638, opened 2026-08-13
Add PATH management to the App controller

Nino-K:
Add a `spec.application.addPath` field (front/back/manual) to control whether `~/.rd<suffix>/bin` is added to the user's PATH and where. front prepends it, back appends it, and manual leaves PATH unchanged (and removes anything previously added by front or back).

Add a new path-management controller that watches the App singleton and applies the setting on the host. On Unix, it updates the usual shell startup files (`.bash_profile/.bash_login/.profile, .bashrc, .zshrc, .cshrc/.tcshrc, and fish's config.fish`), using the same login-file precedence as Rancher Desktop 1. On Windows, it updates the user's Path in `HKCU\Environment` and broadcasts `WM_SETTINGCHANGE`.

The controller runs regardless of VM state since this is all host-side. It should also be idempotent so repeated reconciles don't keep changing the files.

Use a new managedlines helper for the edits, based on RD1's `manageLinesInFile`. It keeps the managed lines inside suffix-tagged markers so different instances (and RD1) don't interfere with each other, and writes the files atomically.

The field defaults to manual so CLI-created instances don't modify shell files unless requested. The GUI sends front when the user opts in.

Addresses: https://github.com/rancher-sandbox/rancher-desktop-2/issues/497

### DIFF
--- a/rdd/docs/design/api_app.md
+++ b/rdd/docs/design/api_app.md
@@ -59,6 +59,7 @@ spec:
     locale: en-zz
     updates:
       enabled: true
+    addPath: manual
   containerEngine:
     name: moby
   kubernetes:
@@ -95,6 +96,27 @@ status:
 - **spec.application.updates.enabled**: For use by the Electron front end, to control whether the
   application will attempt to update itself.  Defaults to `true`.
 
+- **spec.application.addPath**: Controls whether and where the Rancher Desktop bin directory
+  (`~/.rd<suffix>/bin`) is added to the user's `PATH`. The path-management controller edits the
+  user's shell startup files on Unix (`.bash_profile`/`.bash_login`/`.profile`, `.bashrc`,
+  `.zshrc`, `.cshrc`/`.tcshrc`, and fish's `config.fish`) and the user `Environment` in the
+  registry on Windows. Valide values:
+  - `front`: prepend, so the bin directory takes precedence (`bin:$PATH`).
+  - `back`: append, so existing `PATH` entries win (`$PATH:bin`).
+  - `manual`: do not manage `PATH`; remove any lines a previous `front`/`back` added.
+
+  Defaults to `manual`, so instances created by the CLI never edit shell startup files without user's
+  user's knowledge; the GUI is expected to send `front` explicitly when the user opts in.
+
+  On Unix, `manual` removes only the block between our start/end markers, so a hand-written
+  `PATH` line survives anywhere. On Windows the user `Environment` `Path` is a plain
+  semicolon-separated list with nowhere to fence our entry, so under `manual` we can't tell an
+  entry we added from one the user typed. We remove the bin directory only when it sits at the
+  very front or back of `Path` (the two spots `front`/`back` would have written) and leave a copy
+  placed in the middle alone. The trade-off: an entry parked at either end is treated as ours and
+  removed, so to put the bin directory at the front or back on Windows use `front`/`back` rather
+  than placing it there under `manual`.
+
 - **spec.namespace**: The namespace where the owned `LimaVM` and its ConfigMaps are created. Defaults to `default`. **Immutable after creation** — changing it would orphan resources in the original namespace.
 
 - **spec.running**: Set to `true` to start the LimaVM, `false` to stop it. The App controller propagates this value to `LimaVM.spec.running` on every reconcile.
@@ -107,7 +129,7 @@ status:
 
 - **status.kubernetesPort**: The host TCP port allocated for the k3s API server (`7441 + instance.Index()` by default). Set by the App reconciler on the first reconcile after `spec.kubernetes.enabled` becomes `true`, and cleared when `spec.kubernetes.enabled` is set back to `false` so that a fresh port is resolved on the next enable. The `KUBERNETES_PORT` Lima template param is set to this value; Lima's identity port-forward rule binds the same port on the host and forwards it to the guest.
 
-- **status.conditions**: Multiple controllers write here. The App controller mirrors `Created` and `Running` from the owned `LimaVM` and computes `Settled`, the engine controller writes `ContainerEngineReady`, and the Kubernetes controller writes `KubernetesReady`. All writers use `retry.RetryOnConflict` with a re-Get so concurrent status updates do not 409.
+- **status.conditions**: Multiple controllers write here. The App controller mirrors `Created` and `Running` from the owned `LimaVM` and computes `Settled`, the engine controller writes `ContainerEngineReady`, the Kubernetes controller writes `KubernetesReady`, and the PATH management controller writes `PathManagementReady`. All writers use `retry.RetryOnConflict` with a re-Get so concurrent status updates do not 409.
 
   | Type                   | Status    | Reason           | Description                                                       |
   |------------------------|-----------|------------------|-------------------------------------------------------------------|
@@ -130,14 +152,19 @@ status:
   | `KubernetesReady`      | `False`   | `Probing`        | Waiting for k3s API server to respond                             |
   | `KubernetesReady`      | `False`   | `WaitingForNode` | API server answers but no node has reached Ready                  |
   | `KubernetesReady`      | `False`   | `MergeFailed`    | k3s is reachable but merging the instance kubeconfig failed (see `message`) |
+  | `PathManagementReady`  | `True`    | `Applied`        | `spec.application.addPath` was applied for the current generation |
+  | `PathManagementReady`  | `False`   | `Failed`         | Editing a shell startup file or the user Environment failed with a repairable error, e.g. I/O (see `message`) |
+  | `PathManagementReady`  | `False`   | `Malformed`      | A startup file was hand-edited into an unparseable state (a lone marker, or markers in the wrong order); only the user can fix it |
   | `Settled`              | `True`    | `Settled`        | Reconcile chain has caught up with the current spec               |
   | `Settled`              | `False`   | `WaitingForLimaVM` | The App has no `Running` condition yet (nothing mirrored from LimaVM) |
   | `Settled`              | `False`   | `WaitingForEngine` | Engine controller has not yet written `ContainerEngineReady`    |
   | `Settled`              | `False`   | `EngineStale`    | Engine controller has not yet observed the current generation     |
   | `Settled`              | `False`   | `WaitingForKubernetes` | Kubernetes controller has not yet written `KubernetesReady`  |
   | `Settled`              | `False`   | `KubernetesStale` | Kubernetes controller has not yet observed the current generation |
+  | `Settled`              | `False`   | `WaitingForPathManagement` | PATH management controller has not yet written `PathManagementReady` |
+  | `Settled`              | `False`   | `PathManagementStale` | PATH management controller has not yet observed the current generation |
   | `Settled`              | `False`   | `ApplyingTemplate` | A spec change rewrote the template; the LimaVM has not yet restarted into it |
-  | `Settled`              | `False`   | *(forwarded)*    | Forwarded from the blocking `Running`, `ContainerEngineReady`, or `KubernetesReady` reason |
+  | `Settled`              | `False`   | *(forwarded)*    | Forwarded from the blocking `Running`, `ContainerEngineReady`, `KubernetesReady`, or `PathManagementReady` reason (the latter under `front`/`back`, or under `manual` for a repairable `Failed`) |
 
   `Running=True` means the Lima guest has finished booting and SSH is reachable. It says nothing about the container engine socket; consumers that depend on the engine (container/image/volume mirrors, `docker` against the forwarded socket) must also check `ContainerEngineReady`, which flips to `True` only after the engine controller has connected to the socket and completed its initial full sync.
 
@@ -147,6 +174,23 @@ status:
   - The LimaVM must be running, reported by the mirrored `Running` condition. `Created` and `Running` are both mirrored from LimaVM, including their `lastTransitionTime`, so the timestamps track the LimaVM transition.
   - `ContainerEngineReady` must be `True`. The engine reconciler stamps its `observedGeneration` with the App's `metadata.generation` as it writes the condition, keeping the condition current with the App.
   - `KubernetesReady` must be `True` when `spec.kubernetes.enabled` is set. The Kubernetes reconciler stamps its `observedGeneration` the same way.
+  - `PathManagementReady` must be `True` when the PATH management controller is enabled, but the gate
+    depends on `spec.application.addPath`. It runs host-side, independent of the VM, and stamps its
+    `observedGeneration` with the App's `metadata.generation`. Freshness always gates: until the
+    controller has observed the current generation (`PathManagementReady` absent or its
+    `observedGeneration` behind), `Settled` is `False`/`WaitingForPathManagement` or
+    `False`/`PathManagementStale`, so `rdd set --wait` can't report success before the edit runs —
+    this holds even for `manual`, which still removes a block a previous `front`/`back` left behind.
+    A *failure* holds `Settled` at `False` under `front`/`back`: there the user asked us to edit
+    files, so a failed startup-file or user-Environment edit is surfaced. Under `manual` (the CRD
+    default) it depends on whether the failure is repairable. A *repairable* failure
+    (`PathManagementReasonFailed` — I/O, a read-only home, a full disk) over an intact block still
+    gates: the reconciler requeues and clears it, so `rdd set --wait` shouldn't report success over a
+    residue that's about to go away. A *permanent* failure (`PathManagementReasonMalformed` — a
+    startup file the user hand-edited into an unparseable state) does not gate `Settled` under
+    `manual`: the controller can't repair it and gating would wedge `Settled` forever. Either way the
+    failure is recorded on `PathManagementReady`, so consumers should surface that condition rather
+    than assume `Settled` covers it.
 
   `Settled` carries its own `observedGeneration`: the App's generation observed when the reconciler computed the condition. `rdd set` therefore waits for `Settled=True` with `observedGeneration` at least the post-patch `metadata.generation`, so it settles on its own change rather than a stale snapshot from an earlier generation.
 
--- a/rdd/pkg/controllers/app/pathmanagement/controller.go
+++ b/rdd/pkg/controllers/app/pathmanagement/controller.go
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+// Package pathmanagement registers the PATH management controller. The
+// controller adds or removes the Rancher Desktop bin directory
+// (~/.rd<suffix>/bin) from the user's PATH according to spec.application.addPath,
+// by editing shell startup files on Unix and the user Environment on Windows.
+package pathmanagement
+
+import (
+	"context"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	appv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/app/pathmanagement/controllers"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/controllers/base"
+	"github.com/rancher-sandbox/rancher-desktop-daemon/pkg/instance"
+)
+
+func init() {
+	base.RegisterController(&controller{})
+}
+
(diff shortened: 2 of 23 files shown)


rdd/docs/design/api_app.md
```diff
@@ -95,6 +96,18 @@ status:
 - **spec.application.updates.enabled**: For use by the Electron front end, to control whether the
   application will attempt to update itself.  Defaults to `true`.
 
+- **spec.application.addPath**: Controls whether and where the Rancher Desktop bin directory
+  (`~/.rd<suffix>/bin`) is added to the user's `PATH`. The path-management controller edits the
+  user's shell startup files on Unix (`.bash_profile`/`.bash_login`/`.profile`, `.bashrc`,
+  `.zshrc`, `.cshrc`/`.tcshrc`, and fish's `config.fish`) and the user `Environment` in the
+  registry on Windows. Values:
```
mook-as:
```suggestion
  registry on Windows. Valid values:
```

rdd/pkg/controllers/app/pathmanagement/controllers/pathlist.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// computeWindowsPath returns the new user Path after applying strategy, and
+// whether it changed. It drops any existing binDir entry (case-insensitive, to
+// match Windows) and any empty segments, then re-adds binDir at the front or
+// back. Removing binDir first keeps it idempotent and lets us move the entry
+// between front and back. It takes the raw semicolon-separated Path so we can
+// test it off Windows.
+func computeWindowsPath(current, binDir string, strategy appv1alpha1.AddPathStrategy) (string, bool) {
+	segments := strings.Split(current, ";")
```
mook-as:
```suggestion
	segments := filepath.SplitList(current)
```

rdd/pkg/controllers/app/pathmanagement/controllers/pathlist.go
```diff
@@ -0,0 +1,41 @@
...
+	kept := make([]string, 0, len(segments)+1)
+	for _, seg := range segments {
+		trimmed := strings.TrimSpace(seg)
```
mook-as, line 62: Also need to call [`filepath.Clean`](https://pkg.go.dev/path/filepath#Clean) on it, I think, to avoid issues with `\` suffix vs not. But if we want to test it on not-Windows, we need to bring our own subset of `Clean`…

rdd/pkg/controllers/app/pathmanagement/controllers/pathlist_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	appv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+)
+
+func TestComputeWindowsPathFront(t *testing.T) {
```
mook-as: These tests look like they should be table-driven.

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_unix.go
```diff
@@ -0,0 +1,153 @@
...
+			exists, err := fileExists(p)
+			if err != nil {
+				return err
+			}
+			if exists {
+				if err := managedlines.Manage(p, start, end, []string{posixLine}, !added); err != nil {
+					return err
+				}
+				added = true
+			} else if err := managedlines.Manage(p, start, end, nil, false); err != nil {
```
mook-as: I think I'm missing something: we get here if `!exists` (i.e. the file does _not_ exist, and there was no error).  But `present` (the last argument) is also hard-coded to false. So we want to remove lines from a file that doesn't exist?

I believe the RD1 code just did nothing in this case:
https://github.com/rancher-sandbox/rancher-desktop/blob/920dc45da19e5e2631971527f0ec57cd5df94a28/pkg/rancher-desktop/integrations/pathManagerImpl.ts#L101-L107

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_unix_test.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func exists(path string) bool {
+	_, err := os.Stat(path)
+	return err == nil
+}
+
+func TestApplyPosixFrontCreatesBashProfile(t *testing.T) {
+	home := t.TempDir()
+	configHome := filepath.Join(home, ".config")
```
mook-as: Since it's passed in, we should make `configHome` _not_ be `~/.config` to make sure the overrides work.
```suggestion
	configHome := filepath.Join(home, "xdg-config-home")
```

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_windows.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// broadcastEnvironmentChange tells running applications the environment changed.
+// Errors are ignored: the Path is already written and new processes pick it up.
+func broadcastEnvironmentChange() {
+	env, err := windows.UTF16PtrFromString("Environment")
+	if err != nil {
+		return
+	}
+	user32 := windows.NewLazySystemDLL("user32.dll")
```
mook-as: We should have this be a global variable (as well as `sendMessageTimeout`), so they get re-used between invocations.

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_windows.go
```diff
@@ -0,0 +1,84 @@
...
+	sendMessageTimeout := user32.NewProc("SendMessageTimeoutW")
```
mook-as: If we use `SendMessageTimeoutA`, then we don't need to call `UTF16PtrFromString`, right?

rdd/pkg/controllers/app/pathmanagement/controller.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (c *controller) RegisterWithManager(mgr ctrl.Manager) error {
+	if err := appv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return err
+	}
+	r := &controllers.PathManagementReconciler{
+		Client: mgr.GetClient(),
+		BinDir: instance.BinDir(),
+		Suffix: instance.Suffix(),
```
mook-as, line 62: This is mixed case. That means we end up with `### MANAGED BY RANCHER DESKTOP bats-app-controller START (DO NOT EDIT)`

We might not care, though.

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_windows.go
mook-as, line 1: This is complicated enough that a test might be good. Of course that test would need to be platform-specific…

rdd/pkg/util/managedlines/managedlines.go
```diff
@@ -0,0 +1,137 @@
...
+// ends in a single newline; an empty result is "".
+func computeTargetContents(current, startMarker, endMarker string, desired []string) (target string, changed bool, err error) {
+	before, managed, after, hasBlock, err := splitByMarkers(strings.Split(current, "\n"), startMarker, endMarker)
+	if err != nil {
+		return "", false, err
+	}
+
+	var currentBlock []string
+	if hasBlock {
+		currentBlock = append([]string{startMarker}, append(slices.Clone(managed), endMarker)...)
```
mook-as:
```suggestion
		currentBlock = append(append([]string{startMarker}, managed...), endMarker)
```
Very optional, but this gets rid of a clone.

rdd/pkg/util/managedlines/managedlines.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if slices.Equal(currentBlock, desired) {
+		return "", false, nil
+	}
+
+	lines := make([]string, 0, len(before)+len(desired)+len(after))
+	lines = append(lines, before...)
+	lines = append(lines, desired...)
+	lines = append(lines, after...)
```
mook-as: We can use [`slices.Concat`](https://pkg.go.dev/slices#Concat) here:

```suggestion
	lines := slices.Concat(before, desired, after)
```

rdd/pkg/util/managedlines/managedlines.go
```diff
@@ -0,0 +1,137 @@
...
+
+	for len(lines) > 0 && lines[0] == "" {
+		lines = lines[1:]
+	}
```
mook-as, line 175: Hmm, playing around:

```suggestion
	if i := slices.IndexFunc(lines, func(l string) bool { return l != "" }); i > 0 {
		lines = lines[i:]
	}
```
Nope, not any better. Keep yours.

rdd/pkg/util/managedlines/managedlines.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		if !changed || target == "" {
+			return nil
+		}
+		return atomicfile.Write(path, []byte(target), fileMode)
+	case statErr != nil:
+		return statErr
+	}
+
+	current, err := os.ReadFile(path)
```
mook-as, line 48: That means the file can change under us while we compute. But I don't think there's anything we can do about that, and it's not like the old code was any better.

rdd/pkg/util/managedlines/managedlines_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+const (
+	startMarker = "### MANAGED BY RANCHER DESKTOP 2 START (DO NOT EDIT)"
+	endMarker   = "### MANAGED BY RANCHER DESKTOP 2 END (DO NOT EDIT)"
```
mook-as: We should use markers that obviously do not look like the real ones for testing, I think.

rdd/pkg/util/managedlines/managedlines.go
```diff
@@ -0,0 +1,137 @@
...
+// startMarker and endMarker in the file at path. The write is atomic, and text
+// outside the block isn't touched.
+//
+// If removing the block leaves a regular file empty, we delete it, since we were
+// the ones who created it. Symlinks are always rewritten in place so the user's
+// link stays put, even if that means writing an empty file.
+func Manage(path, startMarker, endMarker string, managedLines []string, present bool) error {
+	desired := desiredLines(startMarker, endMarker, managedLines, present)
+
+	info, statErr := os.Lstat(path)
```
mook-as: Do we need to stat first? Can't we just do `os.IsNotExist` on the `err` returned by `os.ReadFile`? We can just do the stat in the case the desired output is empty, I think.

mook-as (changes requested):

rdd/pkg/controllers/app/pathmanagement/controllers/pathlist.go
```diff
@@ -0,0 +1,56 @@
...
+	updated := strings.Join(kept, ";")
+	return updated, updated != current
+}
+
+// normalizeWindowsPathEntry rewrites a Path entry so slash direction and a
+// trailing separator stop mattering: C:/foo/bin and C:\foo\bin\ both come out
+// as C:\foo\bin. filepath.Clean would handle this on Windows, but off Windows
+// it leaves backslash paths alone, and these tests run off Windows. Callers
+// still compare case-insensitively.
+func normalizeWindowsPathEntry(p string) string {
```
mook-as, line 96: ❤️

rdd/pkg/controllers/app/pathmanagement/controllers/pathlist_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	appv1alpha1 "github.com/rancher-sandbox/rancher-desktop-daemon/pkg/apis/app/v1alpha1"
+)
+
+func TestComputeWindowsPath(t *testing.T) {
```
mook-as, line 15: Would be good to test that we don't touch existing duplicate paths (`C:\foo;c:\foo`)

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_unix.go
```diff
@@ -0,0 +1,156 @@
...
+// login-shell files (only the first existing one carries the block), the bash
+// and zsh rc files, the csh rc files, and the fish config.
+func applyPosix(home, configHome, binDir string, strategy appv1alpha1.AddPathStrategy, start, end string) error {
+	present := strategy != appv1alpha1.AddPathManual
+	posixLine := posixPathLine(binDir, strategy)
+
+	// Bash login files, in precedence order. Only the first existing file gets the
+	// block (so PATH isn't extended twice); the rest have it removed.
+	loginFiles := []string{".bash_profile", ".bash_login", ".profile"}
+	if present {
```
mook-as, line 65: Ignore this, just trying to see if I can simplify.

```go

pending := present
for _, name := range loginFiles {
  p := filepath.Join(home, name)
  exists, err := …
  if !exists {
    continue
  }
  err = managedLines.Manage(p, start, end, []string{posixLine}, pending)
  if err != nil {
    return err
  }
  pending = false
}
if pending {
  p := filepath.Join(home, loginFiles[0]
  err := managedlines.Manage(p, start, end, []string{posixLine}, true)
  if err != nil {
    return err
  }
}

```

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_windows.go
```diff
@@ -0,0 +1,105 @@
...
+// Loaded once and shared across broadcasts. LazyDLL/LazyProc resolve on first
+// use, so declaring them here costs nothing if we never broadcast.
+var (
+	user32             = windows.NewLazySystemDLL("user32.dll")
+	sendMessageTimeout = user32.NewProc("SendMessageTimeoutA")
+)
+
+// envSubKey is the HKCU subkey that holds the user Path. It's a var so tests can
+// point at a throwaway key instead of the real one.
+var envSubKey = `Environment`
```
mook-as, line 40: Wouldn't making the root (`registry.CURRENT_USER`) configurable be better? Specifically, I'm thinking about using temporary hives via [`RegLoadAppKey`](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regloadappkeyw) because it auto-GCs, so we just need to give it a file in [`t.TempDir()`](https://pkg.go.dev/testing@go1.27.0#T.TempDir) that doesn't exist.

That's probably overkill, though… so maybe not.

jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260828-131216-pr-638.html

rdd/pkg/controllers/app/pathmanagement/controllers/pathmanagement_windows.go
```diff
@@ -0,0 +1,105 @@
...
+	if changed {
+		broadcastEnvironmentChange()
+	}
+	return nil
+}
+
+// writeUserPath applies strategy to the user Path in the registry and reports
+// whether it changed. It's separate from apply, and skips the broadcast, so a
+// test can run it against a throwaway key.
+func (r *PathManagementReconciler) writeUserPath(strategy appv1alpha1.AddPathStrategy) (bool, error) {
```
mook-as: Hmm, I guess testing the broadcast means making a (invisible) window and checking its messages… which is kind of annoying, because golang doesn't have good support for that.

mook-as (dismissed):
The only comment that might need to be addressed is the extra test case, I think.

jandubois:
Updated AI review: https://jandubois.github.io/rancher-desktop-2/20260831-152634-pr-638.html

The skill now requires a reproducer to justify a Critical or Important categorization. It seems to have eliminated 2 suggestions from the previous review, but that cannot be explained by the new rules. I also noticed that I should require that the recommended fix will let switch the reproducer from fail to pass. Otherwise a broken reproducer doesn't prove anything.

Anutrix:
Minor suggestion: On linux, can we make a backup copy with same *shrc filename suffixed with `.backup-pre-rd2` before any changes? Like how ohmyzsh does.

Nino-K:
> Minor suggestion: On linux, can we make a backup copy with same *shrc filename suffixed with `.backup-pre-rd2` before any changes? Like how ohmyzsh does.

It could be a good idea, but IMHO I’d leave it alone. Our situation is a bit different from `oh-my-zsh` in a way that makes backups useful there but probably unnecessary for us.

Our edits are already non-destructive and reversible. `oh-my-zsh` rewrites the entire `.zshrc`, so its `.pre-oh-my-zsh` copy is basically the only way to get back to the previous state.

In our case, we only insert a fenced block (`### MANAGED BY RANCHER DESKTOP 2 START/END`), and we write it using a temp file + atomic rename, so there’s no risk of leaving the file partially written. With `addPath=manual`, we can also remove exactly what we added. The undo is essentially built in.

Also, this is a reconcile loop rather than a one-time installer. `oh-my-zsh` creates its backup once during install.sh, whereas we reconcile repeatedly and idempotently. Adding backups here would mean we’d need to be careful about when to create one — only on the first actual change, never overwrite an existing backup, skip no-op reconciles, and avoid backing up files that we created ourselves.

That adds quite a bit of complexity for a backup that, given the points above, we’d probably never actually need to restore.

jandubois:
New AI review: https://jandubois.github.io/rancher-desktop-2/20260901-145208-pr-638.html

jandubois:
I didn't initially understand the challenge we face on Windows, that we don't have a fenced block to remove.

I think when the setting is "manual", we should not remove the directory if it is in the middle of the `PATH`, only if it is at the front or the end. "manual" is supposed to allow the user to precisely pick the location, and by always removing our directory, we remove that choice.

The documentation will have to explain that on Windows you can't put the directory at the front or back in "manual" mode; you have to use "front" or "back" for that. This still retains complete flexibility for the user, even though it may be slightly confusing.

EDIT: I think we should still remove the directory in the middle of the `PATH` as part of `rdd svc delete`.

Nino-K:
> I didn't initially understand the challenge we face on Windows, that we don't have a fenced block to remove.
> 
> I think when the setting is "manual", we should not remove the directory if it is in the middle of the `PATH`, only if it is at the front or the end. "manual" is supposed to allow the user to precisely pick the location, and by always removing our directory, we remove that choice.
> 
> The documentation will have to explain that on Windows you can't put the directory at the front or back in "manual" mode; you have to use "front" or "back" for that. This still retains complete flexibility for the user, even though it may be slightly confusing.
> 
> EDIT: I think we should still remove the directory in the middle of the `PATH` as part of `rdd svc delete`.

This makes sense as it's the approach I originally had, and it evolved to this through all the AI reviews. In my original commit, `computeWindowsPath` always dropped our binDir entry for every strategy, at any position; the comment even said: "It always drops any existing binDir entry." So under manual, it removed the middle entry too.

jandubois:
> In my original commit, `computeWindowsPath` always dropped our binDir entry for every strategy, at any position; the comment even said: "It always drops any existing binDir entry." So under manual, it removed the middle entry too.

I'm confused. Isn't that what it still does? I'm arguing that it shouldn't remove the path in the middle, unless it is a `rdd svc delete` situation.

jandubois:
AI claims the fixes for the previous round introduced new issues. 😢 

https://jandubois.github.io/rancher-desktop-2/20260902-135324-pr-638.html

jandubois:
Updated AI review: https://jandubois.github.io/rancher-desktop-2/20260903-153843-pr-638.html

It includes 2 new findings. They are real edge cases. They would return incorrect results, but I don't see them dropping any other profile data. But maybe we should still address them anyways.

jandubois:
Yet another AI review: https://jandubois.github.io/rancher-desktop-2/20260904-113153-pr-638.html

I think only I2 should be addressed. Claude wants some others to "ride along", but up to you to ignore:


> I1 ships new. I labelled it "pre-existing, not a regression," which is true relative to round 5 — but managedlines doesn't exist at the merge-base (git cat-file -e 86ef261d8:...managedlines.go fails), so the whole helper is new in this PR. Nothing on main has this behaviour today; it arrives with the merge. That doesn't make it I2-class — the trigger is a hand-edited stray marker — but don't read "pre-existing" as "already in production."
>
> Worth riding along with the I2 fix (minutes, not hours):
>
> - S1 — permanentPathError's type assertion → errors.As. Two lines, and whoever fixes I2 is already inside that function. Do them together or the second one won't happen.
> - S3, S8 — two stale statements this round created: app_types.go:72 still says a failed edit uses reason Failed (the round added Malformed), and the rewritten comment block reintroduced rdd set addPath=, a key that doesn't resolve.
> - S9 — Valide values. It's @mook-as's own one-token suggestion block, unapplied through six rounds.

jandubois (approved):
Thanks, LGTM

PR #640, opened 2026-08-17
Containers: implement ImagePullRequest

### LINKED ISSUE #630: Controllers: Containers: Implement `ImagePullRequest`
We need to implement `ImagePullRequest` so the extensions work can use it to pull images.

See description: https://github.com/rancher-sandbox/rancher-desktop-2/blob/f3a78a074729f9227975a0ece715ebea90c16089/rdd/docs/design/api_containers.md#pull-image

Blocks #610

mook-as:
This creates a reconciler for ImagePullRequest objects (causing a docker pull), as well as validation around that, and a BATS test to ensure everything works correctly.

Fixes #630 (for moby only).


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260831-135551-pr-640.html

I've updated the skill to require a "reproducer" that shows the actual failure before the agent can claim a finding is "Critical" or "Important" (or provide a good explanation why the issue cannot be tested on the sandbox).

This makes the report more verbose, but hopefully leads to more realistic categorization. And if you agree, it already gives you a test skeleton. Let me know what you think!

mook-as:
Rebased, mostly addressed.

I2: Ignored, but the other path is now covered.
S3: That's the upstream API; if it's dumb, it may get fixed upstream in the future.
S11: It's staying like that for a bit, might as well; we're still v1alpha1.

jandubois:
New AI review: https://jandubois.github.io/rancher-desktop-2/20260902-005206-pr-640.html

mook-as:
I've addressed most of the important issues; the suggestions were mostly ignored as they were not useful.

jandubois (dismissed):
Thanks, LGTM

Final AI review has one issue that _may_ be worth looking at: https://jandubois.github.io/rancher-desktop-2/20260908-111810-pr-640.html

jandubois:
Branch also needs to be rebased to fix merge conflict

mook-as:
I've added a comment, but I1 doesn't really make sense: the engine needs to be up to check if the namespace is valid.  If the engine isn't up, the user not being able to create image pull requests is not any different from not being able to create containers, etc.  Additionally, we should still allow previously-created (but unreconciled, even though they must have passed the validator) image pull requests to persist.

jandubois (approved):
Thanks, LGTM

The namespace thing is not really important, but it is still weird that an `ImagePullRequest` without a namespace will be queued, but one that specifies the namespace will be rejected (when the engine is not running). Since there aren't namespaces in moby, both requests are otherwise equivalent.

But making ImagePullRequests before the container engine is ready is not something we should worry about.

PR #651, opened 2026-08-29
RDD: API docs: Create new ComposeProject resource

mook-as:
This is meant to assist with running `docker compose`; this is necessary to implement extensions.

This PR only includes the API documentation; my current work-in-progress branch is https://github.com/mook-as/rd2/tree/wip/rdd/rdx/compose/initial (which implements things enough to have a BATS test).


mook-as:
Design updated (per offline conversation) to use `ComposeUpRequest` to create `Compose` objects; `docker compose down` is trigger by deleting the `Compose` object (and using a finalizer).

The work-in-progress implementation is now at https://github.com/mook-as/rd2/commits/wip/rdd/rdx/compose/v2/ (I've left the previous iteration around as reference; will delete everything once it's all merged)

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+    uid: 239ef6a0-63bd-4e87-9a7e-c5d4435ddcd4
+  conditions: []
+```
+
+- **metadata.name**: This name must be constructed by the following:
+  - Take `status.namespace` followed by a literal slash (`/`), then lower-cased
+    `status.name`.
+  - Take the SHA256 of the above string, as a lower case hexadecimal string.
+- **status.namespace**: The container namespace; refers to a
+  [`ContainerNamespace`](#namespaces) object.
```
jandubois: It is unclear if the field contains the namespace name itself, or the name of the `ContainerNamespace` resource. They seem to be the same right now, but that doesn't really work because `containerd` namespaces are case sensitive, and Kubernetes resource names are not.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+- **status.name**: The compose project name.
+- **status.workingDir**: Optional; the compose project directory on the host
+  (i.e. relative to where the RDD process runs).
```
jandubois: I don't think resources should store relative path names. The concept doesn't make sense because the working directory of the daemon and the API client will typically be different.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+- **status.configs**: Optional; the list of compose files used to create the
+  project.  Relative to `status.workingDir`, which means it's also a path on the
```
jandubois, line 680: This is fine because now the relative pathnames are anchored by a absolute directories specified by the same resource.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  `name` is the resource type (`Container`, `Volume`, `Image`), followed by a
+  slash (`/`), followed by the object name (i.e. `.metadata.name`).  Each `name`
+  must be unique.  The `uid` is the object UID (i.e. `.metadata.uid`), used to
+  track when the object has been recreated.
+- **status.conditions**: The normal status conditions; see [below](#status-conditions)
+
+`Compose` objects cannot be manually created; they can ony be created using the
+[`ComposeUpRequest`](#composeuprequest) mechanism, or by the reconciler noticing
+compose project references in resources (containers, volumes, etc.) via the
+`com.docker.compose.project` label on the resources.  In the latter case, the
```
jandubois: It is not specified here, but I would expect the reconcilers to create a `ComposeUpRequest` and not create the projects directly. But I don't understand the use-case, so maybe that is not practical? Either way, it would be good to be explicit in the spec what to expect.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+    <td>Running<td>False<td><tt>docker compose up</tt> is running.
+<tr><td>Errored<td>True<td>Error encountered running <tt>docker compose up</tt>.
+<tr><td>Succeeded<td>True<td><tt>docker compose up</tt> succeeded.
+</table>
+
+The `ComposeUpRequest` object will be automatically reaped some time after the
+`Settled` status condition transitions to `True` (whether it was successful or
+not).
+
+#### Delete `Compose`
```
jandubois: This section should be moved up to the `Compose` docs. Right now it ends up being part of the `ComposeUpRequest` docs.

I understand that in some way deleting the Compose object is similar to a `ComposeDownRequest`, but structurally it belongs in the `Compose` resource docs.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+
+Deleting the `Compose` object will cause `docker compose down` to be run.
+Because `status.workingDir` and `status.configs` may not be available, this will
+only be able to delete resources with the correct labels.  The `Compose` object
+itself will be deleted once that succeeds (because `HasMembers` will become
+`False` at that point).
+
+Note: this may leave behind resources created by `docker swarm`.
```
jandubois: Does `docker compose` create swarm resources?

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+- **spec.workingDir**: The compose project directory on the host (i.e. relative
+  to where the RDD process runs).  Used to look up any files needed.
+- **spec.configs**: Optional; the list of compose files used to create the
+  project.  Relative to `spec.workingDir`, which means it's also a path on the
+  host.  Defaults to `Composefile`.
+- **status.conditions**: The normal status conditions; see [below](#status-conditions-1)
+
+##### Status Conditions
+
+The following status conditions are defined:
```
jandubois, line 774: These definitions are incompatible with the ones we define at the top of the document:

> For the `*Request` resources, they use the `Complete` and `Failed` conditions to express state; those are mutually exclusive (only one of the two can be set to True at once). Once either is set to `True`, the request object is considered to be in a terminal state and will be removed after some timeout. This will be at least one minute (so the caller can read any data), but the precise timing is unspecified.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+The `ComposeUpRequest` object will be automatically reaped some time after the
+`Settled` status condition transitions to `True` (whether it was successful or
+not).
+
+#### Delete `Compose`
+
+Deleting the `Compose` object will cause `docker compose down` to be run.
+Because `status.workingDir` and `status.configs` may not be available, this will
+only be able to delete resources with the correct labels.  The `Compose` object
+itself will be deleted once that succeeds (because `HasMembers` will become
```
jandubois: I assume this will use the same finalizer mechanism as all the container objects, using `engine.rancherdesktop.io/mirror` as defined in "Terminology" at the top. If different, please specify.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+`Compose` object will eventually be deleted after `HasMembers` transitions to
+`False`.
+
+### Status Conditions
+
+The following status conditions are defined:
+
+<table>
+<tr><th>Type<th>Reason<th>Status<th>Description
+<tr><td rowspan=3>HasMembers
```
jandubois, line 714: I was wondering why we need this, given that it provides no information beyond `len(status.members) == 0`. All I can think of is that this provides the timestamp to start the countdown for deleting the resource. Or is it useful for anything else?

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +452,131 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+`Settled` status condition transitions to `True` (whether it was successful or
+not).
+
+#### Delete `Compose`
+
+Deleting the `Compose` object will cause `docker compose down` to be run.
+Because `status.workingDir` and `status.configs` may not be available, this will
+only be able to delete resources with the correct labels.  The `Compose` object
+itself will be deleted once that succeeds (because `HasMembers` will become
+`False` at that point).
```
jandubois: Note that `docker compose down` will not delete Image and Volume members, so `HasMembers` may never become true. Maybe it needs to be `HasContainers`.

rdd/docs/design/api_containers.md
mook-as: Oof, good point. And it's not like we can dictate that users never create confusing things.

Let's say I update everything to say it's the namespace name (and _not_ the name of the `ContainerNamespace`)?  And add a blurb in `ContainerNamespace` to have a way to deterministically generate the k8s name, where the lower-case version just maps to itself; I'll need to think about that for a bit I think.

rdd/docs/design/api_containers.md
mook-as: Ah, I should rephrase that; it's relative to the _host_, but still absolute path names.  So `/Users/Bob/Documents/compose`, instead of `Documents/compose`.

rdd/docs/design/api_containers.md
mook-as: I'll double check; but seeing resources with the compose labels will create the `Compose` directly, because we can't recreate the `docker compose up` command from just the resources we observe.

When we get a `ComposeUpRequest`, that runs `docker compose up`; that creates things (like containers) with the labels, which causes the reconcile to create `Compose`objects.

rdd/docs/design/api_containers.md
mook-as: Yeah, it's basically there for timestamp; we can just set it to `Unknown` on any membership change, which implicitly clears the timer (and we can calculate the new `HasMembers` later).

rdd/docs/design/api_containers.md
mook-as: Thanks, I'll go take a look and adjust; it's been too long since I wrote that bit, and I forgot.

That bit has already been updated (and merged), so it's now `Settled` and `Failed`. Switching to that.

rdd/docs/design/api_containers.md
mook-as: Will do.

rdd/docs/design/api_containers.md
mook-as: Oh, I was using cleanup finalizer from `base`; I'll go double check what's more appropriate.

rdd/docs/design/api_containers.md
mook-as: Oh, I was thinking `docker compose down --rmi --volumes`, but maybe that's not desired?  In that case, yeah, it would need to only track containers…

rdd/docs/design/api_containers.md
mook-as: No, but `docker swarm` uses (IIRC) compose under the hood, so it creates things with compose labels.

mook-as (commented):
Thanks, I'll go make changes based on the comments.

mook-as:
I've left it scraping containers + images + volumes (and making delete handle all of them); if you prefer we only monitor containers (and making delete just delete containers), I can switch to that too. I don't think either is better than the other, API-wise.

rdd/docs/design/api_containers.md
```diff
@@ -89,17 +89,38 @@ directly, without calling back into the engine.
 only useful when using the `containerd` backend; when using `dockerd`, the only
 valid instance will have a name of `moby`, and it cannot be modified in any way.
 
-`ContainerNamespace` objects only have the default metadata, since they
-currently do not need anything else.
+As container namespaces are a superset of Kubernetes object names (due to the
```
jandubois: They are disjoint sets. Kubernetes forbids uppercase and containerd forbids `a--b`.

rdd/docs/design/api_containers.md
```diff
@@ -89,17 +89,38 @@ directly, without calling back into the engine.
...
+lack of case sensitivity in the latter), the object name must be
+deterministically generated from the former.  The algorithm is:
+- Any instances of dash (`-`) are replaced with zero follow by dash (`0-`).
+- Any instances of underscore (`_`) are replaced with one followed by dash
+  (`1-`).
+- Convert any upper case characters to the lower case equivalent followed by a
+  dash (`A` becomes `a-`).
+- If the converted name end with a dash (`-`), append a dash and a z (`-z`).
```
jandubois: Still insufficient. You must also forbid dashes in front of dots. So `fooA.bar` would become `fooa-.bar`, which is invalid in Kubernetes.

I would suggest to use the same kind of rule this document already describes for images and volumes: use the real name, if a valid object name, otherwise use a prefix (e.g. `cns-` here) followed by the sha256 of the real name.

rdd/docs/design/api_containers.md
```diff
@@ -89,17 +89,38 @@ directly, without calling back into the engine.
...
+  name: k8s.io
+status:
+  labels: {}
 ```
 
+- **metadata.namespace**: the Kubernetes namespace; this must be the same value
+  as the [App](api_app.md#app-object) resources's `spec.namespace` field.
+- **metadata.name**: The Kubernetes object name; this is generated from
+  `status.name` using the algorithm specified above.
+- **spec.name**: The container namespace; immutable after creation.
```
jandubois:
```suggestion
- **status.name**: The container namespace; immutable after creation.
```

(see also line above)

rdd/docs/design/api_containers.md
```diff
@@ -150,17 +169,48 @@ status:
...
   lastAction:
     action: pause
     state: Succeeded
     observedAt: "2026-04-15T10:30:00Z"
     completedAt: "2026-04-15T10:30:00Z"
 ```
 
+- **metadata.namespace**: the Kubernetes namespace; this must be the same value
+  as the [App](api_app.md#app-object) resources's `spec.namespace` field.
+- **metadata.name**: The container ID, in lower case hexidecimal.
```
jandubois: Should be container id, or `ctr-` + sha256

rdd/docs/design/api_containers.md
```diff
@@ -150,17 +169,48 @@ status:
...
+- **status.name**: The container name.
+- **status.namespace**: The containerd namespace; same as the `spec.name` of a
+  [`ContainerNamespace`](#namespaces) object.
+- **status.path**: The path to the executable for PID 1 in the container.
+- **status.args**: The arguments to the executable for PID 1 in the container.
+- **status.image**: Image ID; corresponds to [`Image`](#images) object's
+  `.status.id` field.
+- **status.ports**: Exposed ports, similar to moby engine API.
+- **status.labels**: Container labels.
+- **status.status**: One of `created`, `running`, `paused`, `restarting`,
```
jandubois:
```suggestion
- **status.status**: One of `created`, `running`, `pausing`, `restarting`,
```

Also has `unknown`, which is the default.

rdd/docs/design/api_containers.md
```diff
@@ -150,17 +169,48 @@ status:
...
+- **status.error**: Error message when running the container.
+- **status.createdAt**: Time when the container was created.
+- **status.startedAt**: Time when the container started.
+- **status.finishedAt**: Time when the container exited.
+- **status.lastAction**: Outcome of the most recent action; persists until the
+  next action overwrites it, regardless of any observable state changes (e.g. a
+  direct `docker stop`) in between.
+- **status.lastAction.action**: The last observed action.
+- **status.lastAction.state**: The result of the last action; if it is still in
+  progress, an empty string.
```
jandubois: It is either `Succeeded` or `Failed`.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
 engine side volume.
 Webhooks will be needed for validation to reject deleting volumes that are in
 use.
+
+## Compose Projects
+
+`Compose` objects do not reflect actual container engine objects; instead, they
```
jandubois: I know I suggested using `Compose` as a short name, but it just feels wrong when I read it. The object kind should be a thing, and a `Compose` is not a thing. So I would prefer to go back to your original name of `ComposeProject`.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+reflect `docker compose` projects.
+
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: Compose
+metadata:
+  name: foo
+  namespace: rancher-desktop
+status:
+  namespace: k8s.io
```
jandubois: We should not use the `k8s.io` namespace for any examples. It must only be managed by kubelet.

```suggestion
  namespace: moby
```

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  configs: []
+  members:
+  - name: Container/8eb6f2cf72b6616aa743cf9187f350af84c9749dab65474db2530f26745d2ef3
+    uid: 239ef6a0-63bd-4e87-9a7e-c5d4435ddcd4
+  conditions: []
+```
+
+- **metadata.name**: This name must be constructed by the following:
+  - Take `status.namespace` followed by a literal slash (`/`), then lower-cased
+    `status.name`.
```
jandubois: Why lowercasing the project name? I believe compose only allows lowercase names. But if it doesn't, the sha256 will hide the case anyways. Calling lowercase explicitly would fold `Foo` → `foo`. Either it is disallowed to begin with, then lowercasing is a no-op, or it is allowed, then lowercasing is a bug.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  - Take the SHA256 of the above string, as a lower case hexadecimal string.
+- **status.namespace**: The containerd namespace; same as the `spec.name` of a
+  [`ContainerNamespace`](#namespaces) object.
+- **status.name**: The compose project name.
+- **status.workingDir**: Optional; the compose project directory on the host on
+  which the RDD process runs, as an absolute path.
+- **status.configs**: Optional; the list of compose files used to create the
+  project.  Relative to `status.workingDir`, which means it's also a path on the
+  host.
+- **status.members**: A list of objects that are part of this project.  The
```
jandubois: I think we should only track containers, so this field can be `status.containers`. I don't think we can track images by compose project because they may be shared by containers outside the project. I think a ComposeProject should be deleted when it has no containers left, and recreated when containers come back. So it will no longer be the same Kubernetes object, but it will have the same name as before.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  `name` is the resource type (`Container`, `Volume`, `Image`), followed by a
+  slash (`/`), followed by the object name (i.e. `.metadata.name`).  Each `name`
+  must be unique.  The `uid` is the object UID (i.e. `.metadata.uid`), used to
+  track when the object has been recreated.
+- **status.conditions**: The normal status conditions; see [below](#status-conditions)
+
+`Compose` objects cannot be manually created; they must only be created by the
```
jandubois: Maybe we should write "should not" because we have no mechanism actually preventing the user from creating them.

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+    <td>Found<td>True<td>Objects matching this project were found.
+<tr><td>Deleted<td>False<td>The last object for this project was deleted; this project will be reaped.
+<tr><td>Calculating<td>Unknown<td>Action is being processed.
+</table>
+
+### Delete `Compose`
+
+The `engine.rancherdesktop.io/mirror` finalizer as described above is used to
+monitor `Compose` objects being deleted.  Deleting the `Compose` object will
+cause `docker compose down` to be run.  Any associated volumes and images will
```
jandubois:
```suggestion
cause `docker compose down --remove-orphans` to be run.  Any associated volumes and images will
```

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+also be deleted (i.e. using the equvialent of
+`docker compose down --rmi all --volumes`).  Because `status.workingDir` and
+`status.configs` may not be available, this will only be able to delete
+resources with the correct labels.  The `Compose` object itself will be deleted
+once that succeeds (because `HasMembers` will become `False` at that point).
```
jandubois, line 729: Should we specify what happens when `compose down` fails? Or is this obvious (logging an error and retrying with exponential backoff)?

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+#### `ComposeUpRequest`
+
+```yaml
+apiVersion: containers.rancherdesktop.io/v1alpha1
+kind: ComposeUpRequest
+metadata:
+  name: foo
+  namespace: rancher-desktop
+spec:
+  namespace: k8s.io
```
jandubois:
```suggestion
  namespace: moby
```

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  conditions: []
+```
+
+Creating a `ComposeUpRequest` will trigger `docker compose up` and creation of
+a `Compose`.
+
+- **metadata.name**: The name must be constructed in the same was as a
+  [`Compose`](#compose-projects) resource, based on `spec.namespace` and
+  `spec.name`.  That is, the resulting `Compose` object will have the same name
+  as the `ComposeUpRequest`.
```
jandubois: This means on a failed `ComposeUpRequest` we have to delete the failed object before we can retry. Would it make sense to use the same naming mechanism we use for containers, so `generateName` create the name and reports it in status?

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+- **status.namespace**: The containerd namespace; same as the `spec.name` of a
```
jandubois:
```suggestion
- **spec.namespace**: The containerd namespace; same as the `spec.name` of a
```

rdd/docs/design/api_containers.md
```diff
@@ -452,3 +502,137 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  [`ContainerNamespace`](#namespaces) object.
+- **spec.name**: The compose project name.
+- **spec.workingDir**: The compose project directory on the host (i.e. relative
+  to where the RDD process runs).  Used to look up any files needed.
+- **spec.configs**: Optional; the list of compose files used to create the
+  project.  Relative to `spec.workingDir`, which means it's also a path on the
+  host.  Defaults to `Composefile`.
```
jandubois: There is no single default filename. Compose looks for multiple candidates.

rdd/docs/design/api_containers.md
mook-as: Ugh, right. I looked it up:

Kubernetes: each part is lower-alphanumeric or dash, must not start or end with dash; the parts are joined by dot.
https://github.com/kubernetes/kubernetes/blob/40f11e7981d9f5b3d284b22ef5248ced7b78b2c9/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L184

Docker compose: lower case alphanumeric or underscore or dash; must start with alphanumeric.
https://github.com/compose-spec/compose-go/blob/11feead015fc8741906d9a69496556186a8e1364/loader/loader.go#L780-L785

Containerd namespaces, etc.: parts of (upper or lower case) alpha-numeric, joined by any of dot, dash, underscore.
https://github.com/containerd/containerd/blob/f6132dbe1f482cbe0aebc4bd3d8d7a184fb4a2aa/pkg/identifiers/validate.go#L34-L43

rdd/docs/design/api_containers.md
mook-as: Oh, I totally forgot that we went with that format.

rdd/docs/design/api_containers.md
mook-as: Ah, right, because we'd need a `ContainerNamespaceCreateRequest` to make new namespaces, because of mirroring issues…

rdd/docs/design/api_containers.md
mook-as: … Yes, this is the container ID, which is itself a lower-case hex string.

rdd/docs/design/api_containers.md
mook-as: No; we use `generateName` for containers because we can not know the `metadata.name` (i.e. the container ID) ahead of time.  We know the compose project name ahead of time.

If there's a failing `ComposeUpRequest`, then the user would probably need to do something to resolve the issue, so making them delete it to trigger reconcile should make sense.

Alternatively, we can implement some sort of automatic retry instead? So it would be Settled=False,Failed=True?

mook-as (commented):
Docs updated to commit a23d62a27

Test implementation at 09d969c35

rdd/docs/design/api_containers.md
```diff
@@ -555,3 +640,148 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+<tr><td>Calculating<td>Unknown<td>Action is being processed.
+</table>
+
+### Delete `ComposeProject`
+
+The `engine.rancherdesktop.io/mirror` finalizer as described above is used to
+monitor `ComposeProject` objects being deleted.  Deleting the `ComposeProject`
+will cause `docker compose down --remove-orphans` to be run.  Any associated
+volumes and images will also be deleted (i.e. using the equivalent of
+`docker compose down --rmi all --volumes`).  Because `status.workingDir` and
```
jandubois: I feel uneasy about deleting volumes by default. The user may not expect this and have e.g. a persistent psql data store that they expect to still be there when they run `docker compose up` again.

rdd/docs/design/api_containers.md
```diff
@@ -555,3 +640,148 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+`HasMembers` status condition would also be set to `True` to indicate that
+containers have been detected.
+
+Both `status.workingDir` and `status.configs` are set from observed containers;
+if multiple containers are part of the same project, but they disagree on such
+fields, it is undefined which will be used.  As such, these are informational
+only.
+
+Associated containers being deleted would similarly update `status.containers`;
+once the last item has been removed, `HasMembers` would be set to `False`.  The
```
jandubois, line 704: So when the VM is down, the containers will be deleted by the reconciler. This in turn will trigger the compose project to be deleted. This obviously fails while the VM is down, but when it restarts, it will succeed, even if the compose project has auto-restart enabled.

And if we automatically delete images and volumes, we may have destroyed a running/working deployment.

rdd/docs/design/api_containers.md
```diff
@@ -555,3 +640,148 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  - If the candidate name is a valid Kubernetes name (that is, runs of
+    lower-case alphanumeric characters or dash, but does not start or end with
+    dash; each run is joined by a dot), then use it as `metadata.name`.
+  - Otherwise, this is `cmp-` followed by the lower-case SHA-256 hash of the
+    candidate name.
+- **status.namespace**: The containerd namespace; same as the `status.name` of a
+  [`ContainerNamespace`](#namespaces) object.
+- **status.name**: The compose project name.
+- **status.workingDir**: Optional; the compose project directory on the host on
+  which the RDD process runs, as an absolute path.
```
jandubois, line 678: Maybe add a note that this directory may get deleted by the user while the project is running. A specified but not-existing directory must not block deleting the compose project.

rdd/docs/design/api_containers.md
```diff
@@ -555,3 +640,148 @@ Delete the `Volume` object; finalizers will cause deletion of the container
...
+  conditions: []
+```
+
+Creating a `ComposeUpRequest` will trigger `docker compose up` and creation of
+a `ComposeProject`.
+
+- **metadata.name**: The name must be constructed in the same way as a
+  [`ComposeProject`](#compose-projects) resource, based on `spec.namespace` and
+  `spec.name`.  That is, the resulting `ComposeProject` object will have the
+  same name as the `ComposeUpRequest`.
```
jandubois, line 762: Doesn't this also mean to modify a project you have to make sure you search for and delete any existing ComposeUp objects that may still be waiting for reaping? Because you can't create a new ComposeUpRequest with the same name. I still think a generated name would work better because it becomes "create and forget".

rdd/docs/design/api_containers.md
```diff
@@ -124,18 +124,27 @@ name afterwards.
...
+status:
+  name: moby
+  labels: {}
 ```
 
+- **metadata.namespace**: the Kubernetes namespace; this must be the same value
+  as the [App](api_app.md#app-object) resources's `spec.namespace` field.
+- **metadata.name**: As container namespaces may not be valid Kubernetes object
+  names, this is `cns-` followed by the lower case SHA-256 hash of the container
+  namespace name.
```
jandubois: I thought we would follow the same rule as for containers: use the name directly if it is valid, or replace with `cns-`+sha only when it is not.

jandubois (changes requested):
Still some change requests.

I still have some other minor issues, but I would prefer to get this merged now, and maybe do a doc review pass as a separate PR.

rdd/docs/design/api_containers.md
mook-as: Good point. Since (from the other comments) I'll now just _only_ deal with containers (ignoring images and volumes), I'll just add a check to skip the `docker compose down` when `HasMembers=False`.

rdd/docs/design/api_containers.md
mook-as: Yeah, I've changed things to only delete the containers (i.e. no flags).

rdd/docs/design/api_containers.md
mook-as: Comment added (on the next line).

rdd/docs/design/api_containers.md
mook-as: Ah, you're thinking of the case of `ComposeUpRequest` created → `ComposeProject` created → want to re-create `ComposeUpRequest` before the first one has been reaped?

I think we just need to check the `observedGeneration` of the condition, and compare it against the `generation` before reaping.  So in that case the user just needs to do a create-or-update, and things will work out fine.  That would mean we'll need to make `spec.*` mutable, though.

I really want to avoid using generated names, because doing _that_ means we can have two independent `ComposeUpRequests` in flight simultaneously.  While that's actually possible on a machine (just spawn two terminals), I don't think the results are going to be good.

rdd/docs/design/api_containers.md
mook-as: Did that, in the section describing deletions.

mook-as (commented):
Updated to e2b85ab89.

Test implementation updated to eb63ae8ac.

jandubois (approved):
Thanks, LGTM

PR #652, opened 2026-08-29
RDD: BATS: Better WSL detection

mook-as:
Only treat the environment as WSL when we have working WSL interop; as that is needed for running against the host Windows.  If that does _not_ exist, just treat it as Linux; this can be used, for example, in containers.

### DIFF
--- a/rdd/bats/helpers/os.bash
+++ b/rdd/bats/helpers/os.bash
@@ -11,7 +11,15 @@ Darwin)
     OS=darwin
     ;;
 Linux)
-    if grep --quiet WSL2 </proc/sys/kernel/osrelease; then # spellchecker:ignore
+    if is_wsl; then
+        if [[ ! -e /proc/sys/fs/binfmt_misc/WSLInterop ]]; then
+            # WSLInterop is missing; we will die later because we can't spawn
+            # Windows processes.  Die early for better debugging.
+            # To run Linux tests on Windows, run in a VM or a container.
+            # shellcheck disable=SC2016 # We want to print the literal command, not expand it.
+            echo 'WSL detected, but WSL interop is missing.  Try `wsl --shutdown`.' >&2
+            exit 1
+        fi
         OS=windows
     else
         OS=linux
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
I think this test is rather brittle (see https://github.com/rancher-sandbox/rancher-desktop/issues/10049).

If you want to differentiate between running inside a WSL distro vs running inside a container, I would suggest checking if `/bin/wslpath` is a symlink to `/init`, which is true in every WSL distro, but not in a container:

```console
jan@DESKTOP-R0D3F8V:~$ ls -l /bin/wsl*
lrwxrwxrwx 1 root root 5 May 11 12:04 /bin/wslinfo -> /init
lrwxrwxrwx 1 root root 5 May 11 12:04 /bin/wslpath -> /init
```

```console
PS C:\Users\jan> docker run -it busybox sh -c "ls -l /bin/wsl*"
ls: /bin/wsl*: No such file or directory
```

mook-as:
I will switch to that, but I want to point out that if the `WSLInterop` binfmt_misc is not installed, then the tests will fail (because it will fail to run all the executables).

But I guess it's better to die early in that case? After all, it's a known-broken config…

jandubois:
I just realized that we don't really support running BATS in WSL2 for RD2, so this may all be moot, at least right now.

One problem is that interop can be lost while BATS is running. So just testing for it upfront is not enough. I think I want to postpone "BATS on WSL2" until "later"...

mook-as:
Yeah, this actual thing I want to enable with this PR is actually just running BATS _in a container_, as if it's on a Linux host.  It's not possible to do that as if it's a Windows host anyway (since that would require WSL interop, which isn't going to work because container).  But _some_ kind of BATS in a container can be useful.

jandubois (approved):
I had to double-check that `-ef` resolves symlinks before comparing inodes. Strictly speaking the `is_wsl` test would also pass if `/bin/wslpath` was a hardlink instead of a symlink, so the comment isn't exactly correct, but good enough.

Thanks, LGTM

PR #654, opened 2026-09-04
Bump embedded Kubernetes to 1.37.0 and controller-runtime to 0.25.0

jandubois:
The daily Dependabot scan for `/rdd` has failed since 2026-09-02, when a 7-day cooldown released k8s 1.37.0 (published 08-26). It bumps `k8s.io/kubernetes` alone while the 35 staging `replace` directives stay put, which go cannot resolve, so the scan exits 1 and the version has to move by hand.

controller-runtime tracks Kubernetes minors one for one, so v0.25.0 goes with 1.37 the way v0.24.1 went with 1.36.

kine has to move too: 1.37's apiserver requires etcd 3.7, and kine v1.14.2 does not compile against its regenerated protobuf API. v0.17.0 is the release built for etcd 3.7. It reads as a downgrade because v1.14.2 is a stale September 2025 tag that semver-sorts above kine's maintained v0.x line, which is also why nothing ever proposed the move. The dependabot ignore that was waiting on this goes away. v0.17.0 also needs go 1.26.5, which raises the directive and so the toolchain CI installs.

`ServiceAccountTokenGetter` gained two webhook getters, which ours declines the way it already declines pods and nodes. `BuildGenericConfig` and `CreateConfig` each grew a parameter, both passed `nil` to match k8s 1.37's own `samples/generic/server` that this code follows. Everything under `pkg/apis` is `applyconfiguration-gen` output.


jandubois:
Converting to draft. controller-runtime v0.25.0 does not compile for `GOOS=windows`, so `RDD Unit tests (windows-latest)` and `Lint RDD (windows-latest)` both fail. (`Lint RDD (macos-latest)` shows CANCELLED from fail-fast, not a failure of its own.) Everything else is green, including all twelve BATS suites and Package on all four targets.

In `sigs.k8s.io/controller-runtime/pkg/internal/testing/process`, `signal_windows.go:26` declares `func signalProcess` while `process.go:132` declares `var signalProcess = signalProcessImpl`. The name collides and `signalProcessImpl` is undefined, so every package importing `pkg/envtest` fails to build on Windows. Ours are `rdd/pkg/controllers/base/finalizer_test.go` and `rdd/pkg/service/controllers/discovery_test.go`.

```
signal_windows.go:26:6: signalProcess redeclared in this block
        process.go:132:5: other declaration of signalProcess
process.go:132:21: undefined: signalProcessImpl
```

Two upstream PRs collided. kubernetes-sigs/controller-runtime#3519 (merged 2026-06-13) added `signal_windows.go` with `func signalProcess`; kubernetes-sigs/controller-runtime#3523 (merged 2026-06-17) added the `signalProcessImpl` indirection so tests can stub the signal call, renaming the function in `signal_unix.go` and `signal_other.go` but not in the Windows file. Every controller-runtime workflow runs on `ubuntu-latest` with no `GOOS=windows` cross-build, so nothing caught it.

`main` and `release-0.25` are both still broken, v0.25.0 is the newest tag, and no upstream issue exists. The fix is a one-identifier rename. This PR needs either that fix released as v0.25.1, or a temporary `replace` in `rdd/go.mod` pointing at a fork.

jandubois:
kubernetes-sigs/controller-runtime#3586 fixes the Windows compile break that blocks this. One identifier: `signal_windows.go` declares `signalProcess` as a function while `process.go` declares it as a var holding `signalProcessImpl`, so anything importing `pkg/envtest` fails to build for `GOOS=windows`. That is what reddens RDD Unit tests and Lint RDD on windows-latest here.

It is waiting on a CNCF CLA signature and a maintainer's `/ok-to-test`. Even merged, it needs a `release-0.25` backport and a v0.25.1 before we can consume it, since v0.25.0 is the newest tag.

The alternative, if that runs long, is a `replace` in `rdd/go.mod` pointing at a fork with the rename, which costs a non-upstream module in `go.sum` and in the license check. Staying in draft until one of the two resolves.

PR #656, opened 2026-09-04
RDD: Bump to go 1.27

mook-as:
All this just to get `strings.CutLast`… 😛


rdd/pkg/controllers/app/app/controllers/app_controller.go
```diff
@@ -439,7 +433,7 @@ func (r *AppReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.Res
 		// ConfigMaps are not watched (no Owns(&corev1.ConfigMap{})), so deleting
 		// one produces no watch event. Requeue explicitly to guarantee the next
 		// reconcile runs rather than relying on implicit LimaVM status activity.
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{RequeueAfter: 1}, nil
```
jandubois, line 436: https://jandubois.github.io/rancher-desktop-2/20260904-184206-pr-656.html points out that this means 1ns, which is no delay at all. But measuring showed that it didn't matter.

Might still be worth using a larger duration, just in case the machine is slow.

jandubois (dismissed):
Thanks, LGTM

I would still change the `RequeueAfter` duration, even though it doesn't seem to make a difference.

rdd/pkg/controllers/app/app/controllers/app_controller.go
mook-as: Yes, `Requeue: true` also meant no delay. I'm just converting it to the non-deprecated version.

mook-as:
Rebased due to merge conflict.

rdd/pkg/controllers/app/app/controllers/app_controller.go
jandubois: > Yes, `Requeue: true` also meant no delay.

No. It is using the default rate limiter, so the first requeue will be after 5ms, and each following one will double the delay until it hits 1000s.

jandubois: https://github.com/kubernetes-sigs/controller-runtime/blob/91e819d2dbfe0f8c6e37cacec3b04c3004afc97a/pkg/internal/controller/controller.go#L498-L510

jandubois: The way I understand it `ctrl.Result{Requeue: true}` was deprecated because exponential backoff should only be used for errors. If you are polling, you should use a constant polling interval.

jandubois (approved):

PR #657, opened 2026-09-05
[DO-NOT-MERGE] RDA: containers: Add a select-all checkbox to project groups

jandubois:
The containers table already groups rows by compose project, but stopping or deleting a whole project meant ticking every row by hand. The group row now has its own checkbox that selects every container under it, so the bulk Stop and Delete buttons act on the project.

SortableTable tracks the selection by object identity, so `rows` now hands out the same object for a container across recomputes; otherwise the selection cleared itself whenever the daemon reported any container change. The table also matches rows by key field, because the same row can reach the selection through a different Vue proxy. A dead `else` in the bulk-action layout dereferenced the element it had just found missing; it now hides the "N selected" label instead.

Ported from rancher-sandbox/rancher-desktop#10511, which is still open. If it changes before it merges, this will need a follow-up.

The e2e coverage that came with it builds fixtures with the docker CLI, so it is marked `fixme` like the other docker-CLI blocks in that file; four unit tests cover the group-selection logic and the row-object reuse instead.

### DIFF
--- a/e2e/containers.e2e.spec.ts
+++ b/e2e/containers.e2e.spec.ts
@@ -651,3 +651,97 @@ test.describe.fixme('Container Stats Tab', () => {
     await expect(statsPage.cpuChart).not.toBeVisible();
   });
 });
+
+test.describe.fixme('Container Compose Group Actions', () => {
+  let electronApp: ElectronApplication;
+  let composeContainerId1: string;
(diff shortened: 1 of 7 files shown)


jandubois:
CleanShot 2026-09-07 at 22 47 02@2x

PR #658, opened 2026-09-05
Add containerd as a first-class engine backend

### LINKED ISSUE #187: Expose whether the container engine supports namespaces
In an effort to push as much of logic as possible to the back end, I would like to have something that exposes whether the selected container engine supports namespaces (so the UI can hide the drop-down as appropriate).

First scratch at the API:

The `App` resource has a new status condition `SupportsNamespaces`, with values `True` (containerd), `False` (moby), or (the default) `Unknown` (e.g. when the container engine hasn't started yet, but this should _not_ be used to determine if that container engine is ready; the existing `Running` status condition should be used instead).

Alternatively:

The `App` resource has a `.status.supportsNamespaces` field with a boolean value.

I think I prefer the first option, but I'm open to being persuaded (including to options I haven't thought of).

jandubois:
Adds containerd as a first-class engine backend alongside moby. With `containerEngine.name=containerd` the engine controller now starts a containerd watcher that mirrors namespaces, containers and images into `ContainerNamespace`, `Container` and `Image` resources, dispatches container actions, and forwards mirror deletes back to the engine.

**Please review commit by commit.** Each of the thirteen commits is a single idea, and every one builds and passes unit tests on its own, so the series reads far better than the squashed diff.

The commits add the forwarded socket, the watcher with namespace mirrors, the sweep on a backend switch, container and image mirrors, container actions, the container's stop signal, finalizer-forwarded deletes, `status.supportsNamespaces`, which closes #187, the stop timeout, the status fields containerd has no direct source for, the mirror field docs, and an explicit cancellation arm in the stop grace period.

Five things are out of scope, so you do not have to flag them. On Windows containerd still reports `NotApplicable`, since nothing serves the named pipe yet. There are no `Volume` mirrors, because containerd has no volume concept. Container logs return 501 rather than streaming, and passthrough waits on the rdd-guest transport. An `ImagePullRequest` fails with reason `PullFailed`, since the containerd watcher does not pull images yet. The GUI does not read `status.supportsNamespaces` yet, so its namespace selector stays hidden until the renderer is wired up.

Every commit builds for host, `GOOS=windows` and `GOOS=linux` and passes unit tests standalone, every commit that changes a bats file ran that suite against a live VM at that commit, and `make lint-rdd`, `lint-bats` and `check-ltag` are clean with no generated-file drift. The stop timeout and the grace period's cancellation arm have no tests of their own.


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260906-134524-pr-658.html

Note: review includes the first four commits which have been split out into #659

rdd/pkg/controllers/app/engine/controllers/container_actions_containerd.go
```diff
@@ -0,0 +1,646 @@
...
+	// for, and the moby path reports that as success, so the mirror must not
+	// say Failed either. Return without waiting on exitCh, because a Kill
+	// that could not find the task is no guarantee an exit status is coming.
+	if err := task.Kill(nsCtx, stopSignal(nsCtx, log, ctr)); err != nil {
+		if !errdefs.IsNotFound(err) {
+			return fmt.Errorf("failed to signal task: %w", err)
+		}
+		return nil
+	}
+	select {
```
Nino-K, line 317: I think this select also needs `case <-nsCtx.Done():` otherwise it would ignore context cancellation.

rdd/pkg/controllers/app/engine/controllers/engine_reconciler.go
```diff
@@ -166,40 +179,79 @@ func (r *EngineReconciler) Reconcile(ctx context.Context, _ ctrl.Request) (ctrl.
 	r.apiNamespace = app.GetResourceNamespace()
```
Nino-K: This one is not done by this PR, but I thought I should point it out. The `r.apiNamespace` is written without any synchronization, but `HandleLogs` reads it at line 79 above from the passthrough goroutine. It looks like there might be some data race going on here.

rdd/pkg/controllers/app/engine/controllers/container_actions_containerd.go
jandubois: Good find, added in https://github.com/rancher-sandbox/rancher-desktop-2/commit/5542a3ca0665c451e166a71f50779e4ad035343d

rdd/pkg/controllers/app/engine/controllers/engine_reconciler.go
jandubois: This was already fixed by #640 which is now in our history since the rebase.

jandubois:
BATS on Windows was flaky after the rebase; the flake is fixed by #754.

Nino-K (dismissed):

Nino-K (approved):

PR #659, opened 2026-09-06
Extract the shared mirror client from the Docker watcher

jandubois:
Groundwork for the containerd backend (#658) that touches only the moby path. The three Kubernetes-side helpers move out of the Docker watcher into a `mirrorClient` that it embeds, the action-annotation bats helpers move from the moby suite into `helpers/controller.bash` for any engine suite to share, a comment on `dispatchContainerAction` stops claiming a default branch that does not exist, and the moby suite setup captures the App namespace with `run -0`. No behavior change.

### DIFF
--- a/rdd/bats/tests/32-app-controllers/engine-docker.bats
+++ b/rdd/bats/tests/32-app-controllers/engine-docker.bats
@@ -486,32 +486,8 @@ assert_single_mirror() { # <image-id> <expected-tag>
 }
 
 # --- Container actions via annotation ---
-
-# assert_action_consumed reports success once the reconciler has
-# removed the action annotation, which is how we know the dispatch
-# has completed.
-assert_action_consumed() {
-    local cid=$1
-    run -0 rdd ctl get container "${cid}" --namespace="${RDD_NAMESPACE}" \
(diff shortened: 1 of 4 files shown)


rdd/bats/tests/32-app-controllers/engine-docker.bats
```diff
@@ -45,7 +45,8 @@ local_setup_file() {
     # Mirror resources live in App.spec.namespace. Override RDD_NAMESPACE
     # to whatever the App was created with so the test queries the same
     # namespace the engine controller uses, regardless of CRD defaults.
-    RDD_NAMESPACE=$(rdd ctl get app app -o jsonpath='{.spec.namespace}')
+    run -0 rdd ctl get app app -o jsonpath='{.spec.namespace}'
```
Nino-K: This was previously only capturing the stdout, but now it captures both `stderr` and `stdout`, so if the command ever errors, it would leak into the `RDD_NAMESPACE`.

jandubois: `run -0` means the command will fail, but you are right that we should use `--separate-stderr` or whatever it is called. Will change.

Nino-K (approved):

PR #660, opened 2026-09-06
Stop requiring createdAt on Volume mirrors

jandubois:
The Volume CRD required `status.createdAt`, so an unparsable engine timestamp dropped the field and the API server rejected the whole status apply, leaving a mirror with a finalizer and no status.

The first commit substituted the epoch. The second reverses that approach: `createdAt` was never meant to be required, and the epoch reaches the UI as a real date. The requirement arrived when the struct was `VolumeSpec`, where every field was immutable, and outlived both the rename to `VolumeStatus` and the removal of the immutability rules. Container and Image already declare `createdAt` optional, and the mock container reconciler never sets one.

Every optional timestamp now has `,omitempty,omitzero`, matching `ObservedAt` and `CompletedAt`. Without `omitzero` an unset `metav1.Time` marshals as null and depends on the API server pruning it before validation; `omitempty` alone is a no-op on structs.

The Volumes page already guards on `status.createdAt`, so an absent value renders as an empty Created column. The generated `pkg/rdd-client` model still declares the field non-optional; regenerating needs a host dockerd, so that is not in this PR.

### DIFF
--- a/rdd/pkg/apis/containers/v1alpha1/timestamps_test.go
+++ b/rdd/pkg/apis/containers/v1alpha1/timestamps_test.go
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package v1alpha1
+
+import (
+	"encoding/json"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestOptionalTimestampsAreOmitted guards the omitzero tag on every optional
+// timestamp. omitempty alone does not omit a struct, so without omitzero an
+// unset metav1.Time reaches the API server as null.
+func TestOptionalTimestampsAreOmitted(t *testing.T) {
+	tests := []struct {
+		name   string
+		status any
+		fields []string
+	}{
+		{"ContainerStatus", ContainerStatus{}, []string{"createdAt", "startedAt", "finishedAt"}},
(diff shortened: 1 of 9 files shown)


Nino-K (commented):
Minor style nit, non-blocking: I’d consider hoisting the logger to the top of `applyVolume` instead of creating it inline in the `else` block:

```go
log := logf.FromContext(ctx).WithName("docker-watcher")
```

`syncAllVolumes` already does this, so it would also keep things consistent.

The inline logger call here also reads a little awkward since the args are indented at the same level as `V(1).Info`, which makes it harder to see:

```go
logf.FromContext(ctx).WithName("docker-watcher").
    V(1).Info("Reporting the epoch as the volume creation time",
    "volume", vol.Name, "created", vol.CreatedAt)
```

With a local logger, it becomes a bit cleaner:

```go
log.V(1).Info("Reporting the epoch as the volume creation time",
    "volume", vol.Name, "created", vol.CreatedAt)
```

rdd/pkg/apis/containers/v1alpha1/timestamps_test.go
```diff
@@ -0,0 +1,52 @@
...
+func TestOptionalTimestampsAreOmitted(t *testing.T) {
+	tests := []struct {
+		name   string
+		status any
+		fields []string
+	}{
+		{"ContainerStatus", ContainerStatus{}, []string{"createdAt", "startedAt", "finishedAt"}},
+		{"ContainerLastAction", ContainerLastAction{}, []string{"observedAt", "completedAt"}},
+		{"ImageStatus", ImageStatus{}, []string{"createdAt"}},
+		{"VolumeStatus", VolumeStatus{}, []string{"createdAt"}},
```
mook-as, line 28: I'm not sure specifically testing this set of properties is useful. I guess it's fine…

mook-as (approved):

PR #661, opened 2026-09-07
Drop node-gyp and the last native addon

jandubois:
`native-reg` was the app's only native addon, kept for a single HKLM read. Removing it takes node-gyp, `node-gyp-build`, Python, setuptools and the Visual Studio setup with it.

There is a bug behind the cleanup: `native-reg` has been unbuildable from source since node-addon-api 8, because its `binding.gyp` uses the deprecated `require('node-addon-api').gyp`, which resolves to a target that defines nothing. Windows x64 escapes it only because `@electron/rebuild` finds a prebuilt binary; arm64 does not.

The last commit is the one CI never exercised, since it always passed `-SkipVisualStudio`. What the green build does show is `@electron/rebuild` reporting no native module to rebuild at all, so nothing still wants MSBuild. Drop that commit if you would rather keep VS in the developer setup.

### DIFF
--- a/pkg/rancher-desktop/main/update/MSIUpdater.ts
+++ b/pkg/rancher-desktop/main/update/MSIUpdater.ts
@@ -1,4 +1,4 @@
-import { spawn } from 'child_process';
+import { spawn, spawnSync } from 'child_process';
 import path from 'path';
 
 import { AllPublishOptions, newError } from 'builder-util-runtime';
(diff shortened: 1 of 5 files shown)


pkg/rancher-desktop/main/update/MSIUpdater.ts
```diff
@@ -140,27 +139,16 @@ export default class MsiUpdater extends NsisUpdater {
    * shouldElevate indicates whether we need elevation to install the update.
    */
   protected get shouldElevate(): boolean {
```
mook-as, line 141: I think eventually we need to this to a golang library; not sure yet if that will ship with RDA or RDD.  Either way, I'd like all significant logic to be not in TypeScript, so y'all wouldn't be as averse to work on it.

Alternatively, just have the installer drop a marker file when doing admin-installs… that seems simpler.

mook-as (approved):

Issue #434, opened 2026-06-11
Linux/arm64 and Windows/arm64 support


mook-as:
### Problem Description

I believe the current setup should support Linux/arm64 as well; investigate if we can ship support for it.

### Proposed Solution

We should support it in CI:
- Packaging: produce Linux/arm64 rdd binaries, as well as rda zip files.
- Packaging: see what we need to fix in the rpm/deb/AppImage configuration

### Additional Information

I do not believe we currently have access to appropriate hardware to test this, other than maybe the CI runners.

jandubois:
Linux/arm64 needed nothing beyond a matrix entry: every dependency already publishes arm64, and the packages build. Windows/arm64 needed four separate fixes, three of them outside the packaging code.

## The pull requests

| PR | What it does |
| --- | --- |
| #661 | Removes `native-reg`, the last native addon, by reading the one HKLM value it was there for with `reg.exe`. That takes node-gyp, `node-gyp-build`, the Python and setuptools install, and the Visual Studio setup with it. |
| #662 | Adds the aarch64 packaging legs on `ubuntu-22.04-arm` and `windows-11-arm`, builds the Windows installer for the target architecture, and gives the Linux zip and the MSI architecture-specific names so a release can carry both. |
| #663 | Adds the arm64 guest image to the WSL2 template. Without it Rancher Desktop on Windows/arm64 has nothing to boot. |
| #664 | Switches steve to the same platform matrix every other dependency uses, so a platform vanishing upstream fails loudly instead of dropping out of the manifest. |
| #665 | Stops the x86_64 smoke test picking up an arm64 asset once releases carry both, and fetches yq for the runner's architecture. |
| #666 | Lets the smoke test install and launch a build from a Package workflow run rather than only from a release. |
| #668 | Makes the Package workflow's `sign` input work. `github.event.inputs` holds strings, so `false` read as true and the signing tests always ran. |

## Merge order

The forks cannot be stacked, because GitHub will not base a pull request from a fork on another fork branch. Two of these therefore carry the commits of the one below them, and need a rebase once it lands.

1. **#661** first. It stands alone and fixes a bug on every architecture: `native-reg`'s from-source build has been broken since node-addon-api 8, and x64 escapes it only because a prebuilt binary exists.
2. **#662** next. Rebase it onto `main` first, which drops #661's three commits, then take it out of draft. It is a draft only because it currently shows them.
[...]

PR #662, opened 2026-09-07
Package for Linux and Windows on arm64

jandubois:
Adds the aarch64 packaging legs on `ubuntu-22.04-arm` and `windows-11-arm`, with the installer and naming fixes they need. Closes the first bullet of #434.

All six legs green in my fork, producing an arm64 MSI, Windows zip and Linux zip beside unchanged x64 names.

CI builds these but cannot run them; #434 has the evidence.


.github/actions/yarn-install/action.yaml
```diff
@@ -31,6 +31,13 @@ runs:
     shell: powershell
     run: .\scripts\windows-setup.ps1 -SkipTools
 
+  # The mingw gcc on the Windows arm64 runners targets x86_64, so cgo cannot
```
mook-as: That's odd, but okay…

mook-as (approved):
I grabbed the arm64 zip to test; every exe/dll seems to be aarch64.

PR #664, opened 2026-09-07
Resolve steve assets from the platform matrix

jandubois:
Verified both directions against the real releases. At v0.1.0-beta9.3 the six assets come out with checksums identical to those in `dependencies.yaml` today, so the rewrite changes nothing. At v0.1.0-beta9.2 it now fails loudly, with a 404 on the absent `steve-windows-arm64.tar.gz`.

That silent case is how the missing Windows build in #662 went unnoticed until someone packaged for arm64.

Needs #662's bump to v0.1.0-beta9.3 to merge first.

### DIFF
--- a/scripts/dependencies/tools.ts
+++ b/scripts/dependencies/tools.ts
@@ -1,6 +1,5 @@
 import path from 'path';
 
-import { defined } from '@/pkg/rancher-desktop/utils/typeUtils';
 import {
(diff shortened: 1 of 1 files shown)


mook-as (approved):
I can't parse your pull request description (it's a bit too dense), but the actual code change makes sense.

PR #666, opened 2026-09-07
CI: Smoke test a build from the Package workflow

jandubois:
Stacked on #665, whose commits it carries, because GitHub will not base a pull request from a fork on another fork branch. It also expects the arch-suffixed artifact names from #662.

Verified against a real Package run rather than reasoned about: all four installer legs installed and launched the CI build, the three release-only jobs skipped, and the guard that force-fails a release smoke test with signature checks off correctly stayed out of the way.

The Windows leg covered more than itself. Those artifacts came from a run built on #662, so the MSI that installed and launched was produced by its changed installer code.

### DIFF
--- a/.github/workflows/smoke-test.yaml
+++ b/.github/workflows/smoke-test.yaml
@@ -89,22 +89,28 @@ jobs:
         path: |
           *.x86_64.dmg
           *.x86_64.dmg.sha512sum
-    - name: Upload Windows artifacts
+    # The aarch64 exclusions keep an arm64 asset out of the x86_64 bundle once
+    # releases carry both.  `smoke-test.sh` installs the first archive it finds,
+    # and `aarch64` sorts first, so without them the x86_64 legs would install
+    # the arm64 build.  Add aarch64 bundles here when releases publish them.
+    - name: Upload Windows x86_64 artifacts
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       with:
-        name: application-win32.zip
+        name: application-win32-x86_64.zip
         if-no-files-found: error
         path: |
           *.msi
           *.msi.sha512sum
-    - name: Upload Linux artifacts
+          !*aarch64*
+    - name: Upload Linux x86_64 artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yaml
```diff
@@ -73,6 +83,38 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
         RELEASE_TAG: ${{ needs.find-release.outputs.release-tag }}
+    - name: Download Package workflow artifacts
+      if: inputs.package-run != ''
+      run: |
+        set -o xtrace
+        # Each artifact contains one file, already under its release asset
+        # name.  A CI build ships no checksum, so make one.  It says nothing
```
mook-as: Sigh. _[verb] no [noun]_ again.

.github/workflows/smoke-test.yaml
```diff
@@ -73,6 +83,38 @@ jobs:
       env:
         GH_TOKEN: ${{ github.token }}
         RELEASE_TAG: ${{ needs.find-release.outputs.release-tag }}
+    - name: Download Package workflow artifacts
```
mook-as: It would be better to use [`actions/download-artifact`](https://github.com/actions/download-artifact) with a custom [run-id](https://github.com/actions/download-artifact#download-artifacts-from-other-workflow-runs-or-repositories) instead.  We'd need to create the checksums in a step after, but we'd get a better description of what's going on.

.github/workflows/smoke-test.yaml
```diff
@@ -443,7 +488,7 @@ jobs:
     name: Flag that signing checks were skipped
     needs: [smoke-test-installer, smoke-test-repository, smoke-test-appimage]
     runs-on: ubuntu-latest
-    if: inputs.skip-signing-checks
+    if: inputs.skip-signing-checks && inputs.package-run == ''
```
mook-as: Why would `skip-signature-checks` be on? That seems like a silly combination.

Or did you want `||` instead of `&&`?

mook-as (changes requested):

PR #668, opened 2026-09-07
CI: Make the Package workflow sign input work

jandubois:
Dispatching the Package workflow with `sign: false` ran the signing tests regardless; I watched both succeed in a run dispatched with `--field sign=false`.

It costs two extra jobs on a manual dispatch, which is presumably why it went unnoticed.


mook-as (approved):
Well, it's not any _more_ broken, so why not :D

Issue #708, opened 2026-09-08
Implement a mechanism to pass embedded image to LimaVM


jandubois:
Part of #449. Depends on #711, which downloads and verifies the image at build time.

## Why

RD2 should bundle everything it needs to run containers, so that it installs and works on an airgapped host. Creating a VM today downloads the distro image from a GitHub release, unless Lima's cache already has it, and every distro bump means hand-editing the version and digest in `lima-images-{unix,wsl2}.yaml`, separately from the guest manifest that rddepman maintains. So this issue embeds the image in `rdd`, which extracts it for Lima when it creates the VM.

It covers the distro as downloaded; the overlay (#450) comes next and changes only how the build produces the embedded file. Kubernetes still downloads k3s at runtime unless its cache is pre-populated, which is outside this issue.

## Design

Lima needs no change. On hosts without xz, the limavm controller already writes `<instance dir>/image` between `limainstance.Create` and `limainstance.Prepare`, and `Prepare` skips its own download when that file exists. The WSL2 driver does the same when `basedisk` exists, and rdd hardlinks `basedisk` to `image`.

- **Template.** The app templates list one image, `embedded:///distro.raw.xz`, or `embedded:///distro.tar.xz` for WSL2, with no `arch` and no `digest`. Lima fills in the host arch and accepts any location containing `://` without checking it.
- **Embedding.** A new `rdd/pkg/embedded` package holds the compressed image as a string. Built with the `with_distro` tag, it embeds `distro.raw.xz`, or `distro.tar.xz` on Windows. Without the tag it's empty, so `go test` and lint need no image.
- **Extraction.** For an `embedded:` location, the controller decompresses the embedded image into `<instance dir>/image` with rdd's pure-Go xz decoder, whether or not the host has xz. An `rdd` binary built without the tag refuses to create the VM and says why.
- **Build.** `make build-rdd` and `make build-lima-controller` run `download-guest-deps` (#711) and build `bin/rdd` and `bin/lima-controller` with `with_distro`. `make test`, `make lint-rdd` and the other standalone controller binaries stay untagged.

LimaVMs whose template names an `https://` image still go through Lima's downloader. The bats Lima tests boot the embedded image, as the App does.

## Accepted costs

- `rdd` grows by the compressed image, which for v0.2.8 is 207 MiB on arm64 macOS, 250 MiB on amd64 macOS and Linux, and 177 MiB on Windows.
[...]

PR #711, opened 2026-09-09
rdd: add the guest dependency downloader

### LINKED ISSUE #709: Create downloader for guest dependencies
The rdd binary is going to embed the distro image (#708), so the build needs the image on disk before `go build` runs. Postinstall cannot put it there, because CI's shared yarn-install action runs `make build-rdd` before `yarn install`, and the bats workflow never runs yarn at all. So `make build-rdd` has to stage the image itself, with nothing from node.

That needs a Go downloader that picks the asset for the build target from `rdd/dependencies.yaml` and verifies the bytes against the checksum rddepman records.

jandubois:
Fixes #709.

Adds `download-guest-deps`, which stages the distro image for the rdd build to embed. It picks the asset for the build target from `rdd/dependencies.yaml` and verifies the bytes against the checksum rddepman recorded. Nothing calls it yet; the Makefile rule and the embed come with #708.

The second commit corrects rddepman's nerdctl note, which claimed the build already stages nerdctl.

To try it, follow "Trying it out" in `rdd/cmd/download-guest-deps/README.md`.

### DIFF
--- a/rdd/pkg/guestdeps/stage.go
+++ b/rdd/pkg/guestdeps/stage.go
@@ -1,0 +1,586 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-FileCopyrightText: The Rancher Desktop Authors
+
+package guestdeps
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"math"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync/atomic"
+	"time"
+)
+
+// downloadAttempts bounds the retries of a transfer that drops, stalls or is
+// throttled partway through a several-hundred-megabyte asset. Anything longer
+// is an outage a re-run has to cover.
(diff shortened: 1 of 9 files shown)


jandubois:
I've manually tested the PR using the instructions in the README.md.

rdd/pkg/guestdeps/stage.go
```diff
@@ -0,0 +1,586 @@
...
+		if delay, ok := retryAfter(resp.Header.Get("Retry-After"), time.Now()); ok {
+			return 0, &retryAfterError{err: err, delay: delay}
+		}
+		// Otherwise a 4xx means the manifest names something the server does not
+		// have, which a retry cannot conjure up. The two exceptions are about the
+		// request itself, so repeating it can succeed.
+		if resp.StatusCode >= 400 && resp.StatusCode < 500 &&
+			resp.StatusCode != http.StatusTooManyRequests &&
+			resp.StatusCode != http.StatusRequestTimeout {
+			return 0, &fatalError{err}
```
Nino-K, line 404: If a bare `403` really is a throttle GitHub uses on these downloads, I don't think it should be unconditionally fatal. There might be a few options: treat `403` the same as `429` and let it retry, or fall back to `x-ratelimit-reset` when `Retry-After` is missing.

Nino-K (approved):

PR #737, opened 2026-09-09
E2E: Update configuration

mook-as:
- Use multiple workers on developer machines; this is possible because we use a different RDD instance per test file anyway.
- Forbid `test.only` in CI.
- Add the `github` built-in reporter in CI.

Sample run: https://github.com/mook-as/rd2/actions/runs/34418158732 — there's a new annotation (from the `github` reporter, I believe).


jandubois (approved):
Thanks, LGTM

PR #738, opened 2026-09-10
BATS: retry deleting executables

mook-as:
On Windows, `rdd.exe` may be locked if Windows Defender etc. is scanning the file.  Wrap the `rm -rf` in `try` in case we can successfully delete it after a bit; this would normally not be required.

This is seen in https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/34408367602/job/102656722304


jandubois (approved):
Thanks, LGTM

PR #739, opened 2026-09-10
CI: use self-repository syntax for uses:

mook-as:
GitHub now supports a new syntax to indicate that a reusable workflow comes from the same repository; it does not require an explicit checkout before use.  Switch to that syntax to avoid the potential of using the incorrect workflow.

See: https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/


jandubois (approved):
Thanks, LGTM

Issue #740, opened 2026-09-10
Revert #729: @rancher/components alpha.4 breaks badge colours and drops RadioButton attributes


jandubois:
PR #729 bumped `@rancher/components` from 0.3.0-alpha.1 to 0.3.0-alpha.4 and merged on 2026-09-09. We should revert it. It breaks two things visibly, and the release it came from is a snapshot that upstream has since moved 69 commits past with no sign of publishing again.

## Regression 1: info, warning and error badges lose their pill

alpha.4 reworked BadgeState from a bordered pill to a filled one, and the new rules reference CSS variables we don't define. All nine of these are absent from this repo. `--on-info-banner`, `--info-badge`, `--info-banner`, `--on-error-banner`, `--error-badge`, `--error-banner`, `--on-warning-banner`, `--warning-badge`, `--warning-banner`.

alpha.1 shipped `.badge-state.bg-info { border-color: var(--info) }`. alpha.4 ships `.badge-state.bg-info { color: var(--on-info-banner); background: var(--info-badge, var(--info-banner)) }` and no longer sets `border-color`, so the base rule's `border: 1px solid transparent` stands. With both variables undefined, both declarations are invalid at computed-value time, so the badge gets no background and no border. Computed styles, measured in headless Chrome against our own theme values:

```
bg-info  alpha.1  color=rgb(20,20,25)  background=transparent  border=rgb(61,152,211)
bg-info  alpha.4  color=rgb(20,20,25)  background=transparent  border=transparent
```

The badge renders as bare text. `bg-error` and `bg-warning` fail the same way. `bg-success` still renders, because its fallbacks (`var(--success-badge, var(--success))` and `var(--on-success-banner, var(--success-text))`) reach variables we do define, though it changes from outlined to filled.

This hits the error badge in `Nav.vue:16` and, in `ImagesScanResults.vue`, the critical, high and medium summary badges at `:151`, `:156` and `:161`, plus the per-row severity badge at `:180` whose colour comes from `SEVERITY_MAP` at `:6-23`. The low and total-issues badges use `bg-darker` and the default, so those two still render.

## Regression 2: RadioButton drops every attribute we pass it

`RadioButton` gained `inheritAttrs: false` in alpha.4 and never re-binds `$attrs`. Mounting both published bundles with identical props:

```
0.3.0-alpha.1  root class: "radio-container locked-radio my-custom-class"  data-testid: "rd-radio"
[...]

PR #742, opened 2026-09-10
Revert "build(deps-dev): bump @rancher/components"

mook-as:
This reverts commit 5e004f71c9af86a39283420a1eacec307ac13514.

Also update dependabot configuration to never update that component; it currently does not appear to be worth bumping.  For deatails, see https://github.com/rancher-sandbox/rancher-desktop-2/issues/740


jandubois (approved):
Thanks, LGTM

PR #745, opened 2026-09-10
rddepman: port the buildx checksum and tachometer fixes from rancher-desktop

jandubois:
Ports two rancher-desktop PRs that unblock rddepman:

- rancher-sandbox/rancher-desktop#10747 verifies the buildx macOS and Windows binaries against `checksums-signed.txt`.
- rancher-sandbox/rancher-desktop#10882 bumps tachometer to 0.2.6, because its publisher deleted the 0.1.1 tag, and adds the MUI theme options 0.2.6 needs.

Without them, rddepman currently fails on the buildx bump, and after that at the next extension bump.

CI doesn't run rddepman; a local `RD_DEPMAN_LOCAL_CHANGES=1` run on this branch resolved all eight pending bumps.

### DIFF
--- a/pkg/rancher-desktop/assets/extension-data.yaml
+++ b/pkg/rancher-desktop/assets/extension-data.yaml
@@ -124,23 +124,32 @@
   publisher: Epinio by Krumware and SUSE
   short_description: Push from source to Kubernetes in one step
 - slug: julianb90/tachometer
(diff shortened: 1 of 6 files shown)


mook-as (approved):
Yeah, this would have made more sense as two separate PRs.

PR #753, opened 2026-09-10
rddepman: quote release notes without pinging upstream

jandubois:
rddepman quotes upstream release notes in its bump PRs, so GitHub notified everyone those notes mentioned and added a backlink to every issue they referenced.

This turns mentions into code and links issue references through redirect.github.com, which GitHub renders as a plain link.

It also quotes only the newest release's notes and links the notes of each older release, because GitHub rejects a pull request body over 65,536 characters and the nerdctl bump from 2.2.2 to 2.3.5 exceeds it (https://github.com/rancher-sandbox/rancher-desktop-2/actions/runs/34683522022).

### DIFF
--- a/scripts/lib/__tests__/release-notes.spec.ts
+++ b/scripts/lib/__tests__/release-notes.spec.ts
@@ -1,4 +1,4 @@
-import { quoteReleaseNotes } from '../release-notes';
+import { formatReleaseNotes, quoteReleaseNotes, Release } from '../release-notes';
 
 describe('quoteReleaseNotes', () => {
   /** A mention in its quoted form. */
@@ -53,5 +53,29 @@ describe('quoteReleaseNotes', () => {
     'https://github.com/other/project/compare/v1.0.0...v1.1.0',
   ])('leaves %j unchanged', (body) => {
     expect(quoteReleaseNotes(body, 'upstream', 'project')).toEqual(body);
+  });
+});
+
+describe('formatReleaseNotes', () => {
+  /** A release of upstream/project. */
+  function release(version: string, body: string): Release {
+    return {
+      name: version, tag_name: version, html_url: `https://github.com/upstream/project/releases/tag/${ version }`, body,
+    };
+  }
+
+  it('quotes the notes of the newest release and links the older ones', () => {
+    const releases = [release('v1.1.0', 'Older notes'), release('v1.2.0', 'Fix #13')];
+
+    expect(formatReleaseNotes(releases, 'v1.0.0', 'upstream', 'project')).toEqual([
+      '## v1.1.0 (v1.1.0)',
+      '[Release notes](https://github.com/upstream/project/releases/tag/v1.1.0)',
+      '[Compare between v1.0.0 and v1.1.0](https://github.com/upstream/project/compare/v1.0.0...v1.1.0)',
+      '',
+      '## v1.2.0 (v1.2.0)',
+      'Fix <a href="https://redirect.github.com/upstream/project/issues/13">#13</a>',
+      '[Compare between v1.1.0 and v1.2.0](https://github.com/upstream/project/compare/v1.1.0...v1.2.0)',
+      '',
+    ].join('\n'));
   });
 });
--- a/scripts/lib/__tests__/release-notes.spec.ts
+++ b/scripts/lib/__tests__/release-notes.spec.ts
@@ -11,6 +11,9 @@ describe('quoteReleaseNotes', () => {
     return `<a href="https://redirect.github.com/${ repository }/issues/${ number }">${ text }</a>`;
   }
 
+  /** A bare pull request URL in its quoted form. */
+  const pullRequest = '<a href="https://redirect.github.com/other/project/pull/56">other/project#56</a>';
+
   it.each([
     ['* Fix it by @someone in #12', `* Fix it by ${ mention('someone') } in ${ link('upstream/project', 12) }`],
     ['Thanks (@another-person)', `Thanks (${ mention('another-person') })`],
@@ -25,10 +28,14 @@ describe('quoteReleaseNotes', () => {
     ['Thanks _@someone_', `Thanks _${ mention('someone') }_`],
     ['See _other/project#34_', `See _${ link('other/project', 34, 'other/project#34') }_`],
     ['Fixes _#12_ and __GH-13__', `Fixes _${ link('upstream/project', 12) }_ and __${ link('upstream/project', 13, 'GH-13') }__`],
-    ['in https://github.com/other/project/pull/56', 'in https://redirect.github.com/other/project/pull/56'],
-    ['in http://github.com/other/project/pull/56', 'in https://redirect.github.com/other/project/pull/56'],
-    ['_https://github.com/other/project/pull/56_', '_https://redirect.github.com/other/project/pull/56_'],
+    ['in https://github.com/other/project/pull/56', `in ${ pullRequest }`],
+    ['in http://github.com/other/project/pull/56', `in ${ pullRequest }`],
+    ['_https://github.com/other/project/pull/56_', `_${ pullRequest }_`],
     ['[fixes #56](https://github.com/other/project/pull/56)', '[fixes #56](https://redirect.github.com/other/project/pull/56)'],
+    ['[56]: https://github.com/other/project/pull/56', '[56]: https://redirect.github.com/other/project/pull/56'],
+    ['<https://github.com/other/project/pull/56>', '<https://redirect.github.com/other/project/pull/56>'],
+    ['<a href="https://github.com/other/project/pull/56">fix</a>', '<a href="https://redirect.github.com/other/project/pull/56">fix</a>'],
+    ['https://github.com/other/project/pull/56/files', 'https://redirect.github.com/other/project/pull/56/files'],
   ])('rewrites %j', (body, expected) => {
     expect(quoteReleaseNotes(body, 'upstream', 'project')).toEqual(expected);
   });
--- a/scripts/lib/__tests__/release-notes.spec.ts
+++ b/scripts/lib/__tests__/release-notes.spec.ts
@@ -1,0 +1,38 @@
+import { quoteReleaseNotes } from '../release-notes';
+
+describe('quoteReleaseNotes', () => {
+  /** A mention in its quoted form. */
+  function code(handle: string): string {
+    return `<code>@\u200B${ handle }</code>`;
+  }
+
+  /** An issue reference in its quoted form. */
+  function link(repository: string, number: number): string {
+    return `[${ repository }#${ number }](https://redirect.github.com/${ repository }/issues/${ number })`;
+  }
+
+  it.each([
+    ['* Fix it by @someone in #12', `* Fix it by ${ code('someone') } in ${ link('example/tool', 12) }`],
+    ['Thanks (@another-person)', `Thanks (${ code('another-person') })`],
+    ['cc @example/maintainers', `cc ${ code('example/maintainers') }`],
+    ['See other/project#34', `See ${ link('other/project', 34) }`],
+    ['See [#34]', `See [${ link('example/tool', 34) }]`],
+    ['in https://github.com/example/tool/pull/56', 'in https://redirect.github.com/example/tool/pull/56'],
+    ['[fixes #56](https://github.com/example/tool/pull/56)', '[fixes #56](https://redirect.github.com/example/tool/pull/56)'],
+  ])('rewrites %j', (body, expected) => {
+    expect(quoteReleaseNotes(body, 'example', 'tool')).toEqual(expected);
+  });
+
+  it.each([
+    'Write to someone@example.com',
+    'Read https://blog.example.com/@someone/post',
+    '[found by @someone](https://example.com)',
+    'Run `npm install @scope/package`',
+    '```\n@someone fixed #12\n```',
+    'https://example.com/page#12',
+    'Zero&#8203;width',
+    'https://github.com/example/tool/compare/v1.0.0...v1.1.0',
+  ])('leaves %j unchanged', (body) => {
+    expect(quoteReleaseNotes(body, 'example', 'tool')).toEqual(body);
+  });
+});
--- a/scripts/lib/__tests__/release-notes.spec.ts
+++ b/scripts/lib/__tests__/release-notes.spec.ts
@@ -22,6 +22,9 @@ describe('quoteReleaseNotes', () => {
     ['Fixes GH-12', `Fixes ${ link('upstream/project', 12, 'GH-12') }`],
     ['Fixes #12/#13', `Fixes ${ link('upstream/project', 12) }/${ link('upstream/project', 13) }`],
     ['Escaped \\#12 and \\@someone', `Escaped ${ link('upstream/project', 12, '\\#12') } and ${ mention('someone') }`],
+    ['Thanks _@someone_', `Thanks _${ mention('someone') }_`],
+    ['See _other/project#34_', `See _${ link('other/project', 34, 'other/project#34') }_`],
+    ['Fixes _#12_ and __GH-13__', `Fixes _${ link('upstream/project', 12) }_ and __${ link('upstream/project', 13, 'GH-13') }__`],
     ['in https://github.com/other/project/pull/56', 'in https://redirect.github.com/other/project/pull/56'],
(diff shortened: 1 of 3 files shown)


scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
+const quotablePattern = new RegExp([
+  // Code spans, fenced code blocks and link text, copied unchanged, since
+  // GitHub finds no mentions or references inside them.
+  /(?<ticks>`+)[\s\S]*?\k<ticks>/.source,
```
mook-as: Do any of these need `/m`?

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+  /\[[^\]\n]*\](?=\()/.source,
+  // A link to an issue or pull request.
+  /https:\/\/github\.com\/(?<path>[\w.-]+\/[\w.-]+\/(?:issues|pull)\/\d+)/.source,
+  // An issue reference, `#12` or `owner/repo#12`, outside URLs and HTML entities.
```
mook-as: Should we assume `owner/repo#12` is preceded by a space? That is, use `(?<=\s)` instead of `(?<![…])`?

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
+const quotablePattern = new RegExp([
+  // Code spans, fenced code blocks and link text, copied unchanged, since
+  // GitHub finds no mentions or references inside them.
+  /(?<ticks>`+)[\s\S]*?\k<ticks>/.source,
+  /\[[^\]\n]*\](?=\()/.source,
+  // A link to an issue or pull request.
+  /https:\/\/github\.com\/(?<path>[\w.-]+\/[\w.-]+\/(?:issues|pull)\/\d+)/.source,
```
mook-as: This should start at a word boundary? `/\bhttps…`

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+  // An issue reference, `#12` or `owner/repo#12`, outside URLs and HTML entities.
+  /(?<![\w.&/-])(?:(?<repository>[\w-]+\/[\w.-]+))?#(?<number>\d+)\b/.source,
+  // A user or team mention, outside email addresses and URLs.
+  /(?<![\w/])@(?<mention>[a-z\d](?:-?[a-z\d])*(?:\/[\w.-]+)?)(?![\w-])/.source,
```
mook-as: Same here, `(<=\s)` instead of `(?<![…])`?

mook-as: We used `[\w.-]` for org names before; surely user names should be the same? (… see, for example, underscore.)

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+    // The last argument to a replacer is the named groups object.
+    const { path, repository, number, mention } = args.at(-1);
+
+    if (path) {
+      return `https://redirect.github.com/${ path }`;
+    }
+    if (number) {
+      const target = repository ?? `${ owner }/${ repo }`;
+
+      return `[${ target }#${ number }](https://redirect.github.com/${ target }/issues/${ number })`;
```
mook-as: I think the link text should stay the same (as `match`); the links are implicitly to the same repository (i.e. not ours).

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+  /(?<![\w.&/-])(?:(?<repository>[\w-]+\/[\w.-]+))?#(?<number>\d+)\b/.source,
+  // A user or team mention, outside email addresses and URLs.
+  /(?<![\w/])@(?<mention>[a-z\d](?:-?[a-z\d])*(?:\/[\w.-]+)?)(?![\w-])/.source,
+].join('|'), 'gi');
+
+/**
+ * Rewrite upstream release notes for quoting in a pull request body.
+ * GitHub notifies every user a body mentions and adds a backlink to every
+ * issue or pull request it references, so mentions become code and references
+ * link through redirect.github.com, which GitHub renders as a plain link.
```
mook-as:
```suggestion
 * Rewrite upstream release notes for quoting in a pull request body.
 * GitHub notifies every user a body mentions and adds a backlink to every
 * issue or pull request it references; use `redirect.github.com` to
 * construct a link instead, which disables the backlink.
```

I'm not sure how to stuff the link to the docs in here, given that it's ridiculously long: https://github.com/github/docs/blob/fca276503b17a1742e2dca1553288e92498c881b/content/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls.md#avoiding-backlinks-to-linked-references

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function quoteReleaseNotes(body: string, owner: string, repo: string): string {
+  return body.replace(quotablePattern, (match, ...args) => {
+    // The last argument to a replacer is the named groups object.
+    const { path, repository, number, mention } = args.at(-1);
+
+    if (path) {
+      return `https://redirect.github.com/${ path }`;
```
mook-as: Should this also be a manually constructed link, with the href?
`[${ match }](https://redirect.github.com/${ path })`

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+    if (number) {
+      const target = repository ?? `${ owner }/${ repo }`;
+
+      return `[${ target }#${ number }](https://redirect.github.com/${ target }/issues/${ number })`;
+    }
+    if (mention) {
+      // The zero-width space keeps a copy of the rendered text from mentioning anyone.
+      return `<code>@\u200B${ mention }</code>`;
```
mook-as: Probably easier to just plop in HTML: `**@<span>jandubois</span>**` (**@<span>jandubois</span>**)
Or if you want to be fancy, `[**@jandubois**](https://redirect.github.com/jandubois)`.

scripts/rddepman.ts
```diff
@@ -165,7 +166,7 @@ async function getBody(dependency: VersionedDependency, currentVersion: Version,
   let lastVersion = dependency.versionToTagName(currentVersion);
 
   return releaseNotes.map(([, release]) => {
-    const body = release.body?.replace(/(?<!\w)(#\d+)\b/g, (n) => `${ owner }/${ repo }${ n }`) || `Release ${ release.name } does not have release notes.`;
+    const body = release.body ? quoteReleaseNotes(release.body, owner, repo) : `Release ${ release.name } does not have release notes.`;
```
mook-as: Since we're here, consider wrapping.

scripts/lib/__tests__/release-notes.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { quoteReleaseNotes } from '../release-notes';
+
+describe('quoteReleaseNotes', () => {
+  /** A mention in its quoted form. */
+  function code(handle: string): string {
```
mook-as:
```suggestion
  function mention(handle: string): string {
```
Since `code` doesn't really make sense here.

scripts/lib/__tests__/release-notes.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  it.each([
+    ['* Fix it by @someone in #12', `* Fix it by ${ code('someone') } in ${ link('example/tool', 12) }`],
+    ['Thanks (@another-person)', `Thanks (${ code('another-person') })`],
+    ['cc @example/maintainers', `cc ${ code('example/maintainers') }`],
+    ['See other/project#34', `See ${ link('other/project', 34) }`],
+    ['See [#34]', `See [${ link('example/tool', 34) }]`],
+    ['in https://github.com/example/tool/pull/56', 'in https://redirect.github.com/example/tool/pull/56'],
+    ['[fixes #56](https://github.com/example/tool/pull/56)', '[fixes #56](https://redirect.github.com/example/tool/pull/56)'],
```
mook-as: We should use something different from `example/tool` here, to ensure the link isn't getting clobbered by the ambient config.

mook-as (changes requested):
Thanks, that's definitely in the right direction.

I wonder if we can do something with [`marked`](https://www.npmjs.com/package/marked) since we already install it  (for release notes).

jandubois:
> I wonder if we can do something with [`marked`](https://www.npmjs.com/package/marked) since we already install it (for release notes).

I don't think so. Testing with a prototype showed leaked several expressions, passed raw HTML through untouched, and turned the body into HTML. The regex caught all test cases (including 70 real releases from history).

scripts/lib/release-notes.ts
jandubois: No, `/m` only affects `^` and `$`. Are you thinking about `/s`, so we can use `.` instead of `[\s\S]`.

scripts/lib/release-notes.ts
jandubois: No, that breaks for URLs marked as italic: `_https://..._` because `_` is a word character.

scripts/lib/release-notes.ts
jandubois: No, could be surrounded by parens, brackets, or Markdown markup for bold/italics.

scripts/lib/release-notes.ts
jandubois: Also no, for the same reason.

jandubois: That expression is for repo names. User (and org) names on Github are alphanumeric plus dashes. Underscores only exist in "Enterprise Managed Users", which are not usable with public repos.

scripts/lib/release-notes.ts
jandubois: Sure, what is more more regex? 😸

scripts/lib/release-notes.ts
mook-as: Ah, and `\b` is bad because `_owner/repo#12_` might be markdownup, which GitHub actually takes…

This is so complicated.

scripts/lib/release-notes.ts
mook-as: This comment is more about, wouldn't the mention name (i.e. user, org, or group) use the same expression as a repo owner (user or org).

scripts/lib/release-notes.ts
mook-as: The link _text_ is still different from the input.

mook-as (changes requested):
I'm confused; some of the 👍ed comments, like using `@<span>user</span>` seem to be missing. I mostly just want to avoid the use of zwsp…  Please either do that, or say you don't want to (with a reason; pretty much anything would be fine).

I'd still like the test to not use `example/tool`, but not really going to block a review on that.

jandubois:
> I mostly just want to avoid the use of zwsp…

I wanted to do that, but Claude made me change my mind:

> `<a href="https://github.com/someone"><code>@someone</code></a>` with a zero-width space after the `@` is exactly Dependabot's markup (see any Dependabot PR here, such as #735), and the zero-width space keeps a copy of the rendered text from pinging anyone. With `@<span>someone</span>`, the copy is a plain `@someone`, which pings them when pasted into a comment.

scripts/lib/__tests__/release-notes.spec.ts
jandubois: Renamed to `other/project` in 07a18f5, the name the cross-repository case already uses. Since 0454229 the tests quote as `upstream/project`, so they already catch a link built from the quoted repository instead of the URL.

scripts/lib/release-notes.ts
jandubois: And the regex had the same problem. It treated `_` as a word character, so `_#12_`, `_owner/repo#12_`, and `_@someone_` went through unquoted, and the last one would ping. a031842 quotes them. An underscore that is part of a word, as in `snake_#12`, still blocks a match, because GitHub doesn't link that either.

scripts/lib/release-notes.ts
jandubois: They share one now. a031842 moves the login pattern into a `login` constant and uses it for mentions, for the owner in `owner/repo#12`, and for the owner in URLs. That also keeps `_owner/repo#12_` from matching with `_owner` as the owner.

scripts/lib/release-notes.ts
jandubois: ~Simon~Claude says:

> The original URL can't be the link text, because GitHub links it again. In a paragraph it links the URL when it is the text of an `<a>`, and in an HTML block it links the URL when it is the text of a markdown link. 1ff2a6b uses Dependabot's form instead, `<a href="https://redirect.github.com/o/r/pull/12">o/r#12</a>`. That also fixes bare URLs inside `<details>`, where GitHub linked the `o/r/pull/12` inside the redirect URL. A URL in a link destination or reference definition, an autolink, an `href`, or one that goes on past the number stays a bare redirect.

scripts/lib/release-notes.ts
jandubois: Done in 0454229: the link text keeps what the notes said, such as `#12` or `other/project#34`, as Dependabot does.

scripts/lib/release-notes.ts
jandubois: Taken in 0454229, with a clause for mentions. The docs.github.com URL is shorter than the blob link, so it went in as an `@see`.

scripts/rddepman.ts
jandubois: Wrapped in 0454229.

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,108 @@
...
+  // inside `_emphasis_`, but not beside an underscore that is part of a
+  // word, as in `snake_#12`.
+  String.raw`(?<![a-z\d&\\]|[a-z\d]_)\\?(?:(?<repository>${ login }/[\w.-]+)#|#|GH-)(?<number>\d+)(?![a-z\d]|_[a-z\d])`,
+  // A user or team mention, outside email addresses.
+  String.raw`(?<![a-z\d\\]|[a-z\d]_)\\?@(?<mention>${ login }(?:/[\w.-]+)?)(?![a-z\d-]|_[a-z\d])`,
+].join('|'), 'gis');
+
+// Text that ends where only a URL fits: a markdown link destination or
+// reference definition, an autolink, or an HTML attribute.
+const urlOnlyContext = /(?:\][(:]\s*|[<"])$/;
```
mook-as: Should this go inside the `if (path)` block? It appears to be only relevant there. (Because it's a `const` literal, it shouldn't get re-parsed on each invocation, IIRC)  Alternatively, this could be defined before `quotablePattern` as a lookbehind, and just stuck in there: ``String.raw`(?<context>${ urlOnlyContext })?…``

Similarly for the after-URL thing (which you currently have as a literal).

The comment is kind of hard to parse (I think I had to read it half a dozen times). If I were to rephrase, I'd have used:
> Regular expression that matches text that would precede a URL in markup: a markdown link destination, reference definition, autolink, or HTML attribute.

For the _HTML attribute_ case, I'd have expected an `=` somewhere?

scripts/lib/release-notes.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (urlOnlyContext.test(before) || /^[/#?]/.test(after)) {
+        return href;
+      }
+      // In an HTML block, GitHub would link the `owner/repo/pull/12` inside a
+      // bare redirect URL.
+      const [pathOwner, pathRepo, , pathNumber] = path.split('/');
+
+      return `<a href="${ href }">${ pathOwner }/${ pathRepo }#${ pathNumber }</a>`;
+    }
+    // HTML links work inside HTML blocks too, where GitHub parses no markdown.
```
mook-as:
```suggestion
    // HTML links work inside HTML blocks too, where GitHub does not parse markdown.
```

mook-as (changes requested):
Making the copy work is _exactly_ why I wanted to use `<span>`; if you're manually copy/pasting (instead of using the quote functionality), you probably did want to mention something about that specifically.

Don't actually care about that very much, though; requesting changes only because the comment on line 22 was hard to parse.

jandubois:
I was going back and forth about the zero width space, but in the end decided that matching the dependabot behaviour is the right thing.

I've updated (and moved) the comment.

> For the HTML attribute case, I'd have expected an `=` somewhere?

Thanks, that was a real bug, hopefully fixed.

mook-as (approved):

PR #763, opened 2026-09-12
Embed the guest distro image in rdd

### LINKED ISSUE #708: Implement a mechanism to pass embedded image to LimaVM

jandubois:
Fixes #708.

Embeds the guest distro image in `bin/rdd` and `bin/lima-controller` (`with_distro` build tag) and has the lima controller extract it into the instance directory before Lima's `Prepare`, so creating the app's VM downloads nothing. `make build-rdd` stages the image with `download-guest-deps`. rdd decodes the embedded image itself on every host and leaves any other image to Lima, so a downloaded `.xz` image needs `xz` on the host.

The other commits decode the image into a sparse file, name the embedded guest agent after its architecture, and boot the Lima bats VMs from the embedded distro.


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260914-225700-pr-763.html

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -394,75 +392,69 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	return ctrl.Result{}, nil
 }
 
-// ensureImageDecompressed pre-produces the decompressed distro image with the
```
Nino-K, line 397: Just want to confirm that nothing (a custom template or the dev go build workflow without `with_distro`) still relies on downloading a `.xz` on a host without `xz`. Might also be worth trimming the `xz` package doc comment, which still says its job is to provision "without a system xz binary on the host".

rdd/pkg/guestagent/guestagent_amd64.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
```
Nino-K, line 1: I looks like `guestAgent` now only compiles for `amd64` and `arm64`, whereas before it was arch agnostic. Since this is quite narrow, do you think a `//go:build` guard with a clear message or a fallback file would fail more loudly?

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
```diff
@@ -394,75 +392,69 @@ func (r *LimaVMReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-		return xz.DecompressFile(ctx, cached, imagePath)
-	}
-	// No image matched this arch; let Prepare surface its usual error.
+	}
+	return false
+}
+
+// decompressEmbeddedImage decompresses the distro image embedded in this
+// binary to imagePath.
+func decompressEmbeddedImage(ctx context.Context, imagePath string) error {
```
Nino-K, line 433: `usesEmbeddedImage` keys off the embedded: prefix of the location, but `decompressEmbeddedImage` never looks at the rest of it. It always writes `embedded.Distro`. So the `distro.raw.xz` in `embedded:///distro.raw.xz` is decorative; any `embedded://<anything>` would resolve to the same single baked-in image. That's fine while there's exactly one embedded artifact, but the template reads like it selects a named file when it doesn't. It might be worth either adding a short comment that the path is ignored, or passing `img.File.Location` through and matching on it, so a second embedded image later doesn't silently extract the wrong one.

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: No, a build without `with_distro` will fail to create the App VM with "this binary has no embedded distro image; build it with make".

I've updated the doc comments.

If you create a LimaVM with a template that references a `.xz` compressed image URL, then the Lima code will still attempt to decompress with an external binary. We will address this later; right now we are only concerned about the App VM.

rdd/pkg/guestagent/guestagent_amd64.go
jandubois: It was not really arch agnostic; it would just pick whatever arch was last compiled locally because they would both compile to the same binary name. I think having the arch in the filename is better.

I'm not concerned about any other archs right now.

rdd/pkg/controllers/lima/limavm/controllers/limavm_controller.go
jandubois: This is correct (and I have updated the comment). I had an elaborate proposal in #233, but once I realized that we are not going to embed all the bundled utilities (docker, helm, etc) in the binary, I decided it was overengineered and closed it. 

For now we just want to bundle the distro; we can refactor it when we bundle more (e.g. reference the guestagent via URL, or want to bundle the distro in the App controller and not the LimaVM controller).

Nino-K (approved):

PR #766, opened 2026-09-14
Name release artifacts by platform and architecture

### LINKED ISSUE #669: Unify release artifact names: one prefix, with platform, architecture and extension the only differences
A single release publishes four different name shapes today, and the generated filenames differ again from the released ones, because the release names are applied by hand.

| | generated by the build | published in the release |
| --- | --- | --- |
| macOS dmg | `Rancher Desktop 2-<version>.dmg`, `…-arm64.dmg` | `Rancher.Desktop-<version>.x86_64.dmg`, `….aarch64.dmg` |
| macOS zip | `Rancher Desktop 2-<version>-mac.zip`, `…-arm64-mac.zip` | `Rancher.Desktop-<version>-mac.x86_64.zip`, `….aarch64.zip` |
| Windows msi | `Rancher.Desktop.Setup.<version>.msi` | `Rancher.Desktop.Setup.<version>.msi` |
| Windows zip | `Rancher Desktop 2-<version>-win.zip` | not published |
| Linux zip | `rancher-desktop-2-<version>-linux.zip` | `rancher-desktop-linux-v<version>.zip` |
| rdd | `rdd`, `rdd.exe` | `rdd.<version>.darwin.aarch64`, `rdd.<version>.exe`, `rdd.<version>.linux` |

Four prefixes (`Rancher.Desktop-`, `Rancher.Desktop.Setup.`, `rancher-desktop-linux-v`, `rdd.`), the version with and without a leading `v`, spaces in some names and dots in others, and the architecture spelled `x86_64`/`aarch64` in one place and `arm64` or nothing in another. Only macOS marks the architecture at all, and only there because it has shipped two of them for years.

## Proposal

Every artifact shares one prefix of product name and version, and differs only in what it has to: the platform, the architecture, and the extension.

- **No spaces anywhere.** Windows tooling is happier without them, and one scheme that holds everywhere is worth more than per-platform prettiness.
- **Drop `Setup` from the Windows installer.** The `.msi` extension already says what it is.
- **Every artifact carries its architecture, including x86_64.** Today the Intel macOS build is identifiable only because the other one says `arm64`; unzip both into one directory and nothing marks the Intel one.
- **The build emits the final names.** The release names are produced manually today, which is why the two halves of the table disagree. If the build emits what gets published, that step disappears rather than being automated.

One shape to argue about, borrowed from what `rdd` already does:

```
rancher-desktop-2.<version>.<platform>.<arch>.<extension>
rancher-desktop-2.2.0.0.darwin.aarch64.dmg
rancher-desktop-2.2.0.0.windows.x86_64.msi
rancher-desktop-2.2.0.0.linux.aarch64.zip
```

## Open questions

**The platform component looks unavoidable.** Architecture and extension alone cannot separate the macOS zip from the Linux zip, since both are `.zip`, and the macOS zip cannot simply go away because the updater consumes it.

**Does `rdd` share the prefix?** It is a separate binary rather than a packaged app, and it already uses `<name>.<version>.<platform>.<arch>`. Either it adopts the product prefix or it stays as it is and we accept two families.

**Dots or dashes as the separator**, and whether the version keeps its `v` prefix anywhere.

## Why now

RD2 is still in alpha, so we can change these names freely. That stops being true at 2.0, and every release until then makes the change slightly more disruptive.

arm64 for Windows and Linux (#434) doubles the artifact count, and the current scheme has nowhere to put a second architecture for either: `Rancher.Desktop.Setup.<version>.msi` and `rancher-desktop-linux-v<version>.zip` carry no architecture at all.

#665 waits on this. It teaches the release smoke test to pick the artifact matching the leg it runs on, which cannot be written cleanly until the names settle.

jandubois:
Closes #669.

The build and `yarn sign` now write every published name, so a release needs no renaming by hand.

```
rancher-desktop-<version>.<platform>.<arch>.<ext>
rdd.<version>.<platform>.<arch>[.exe]
```

The Package and upgrade-generate workflows name their CI artifacts the same way without the version, e.g. `rancher-desktop.darwin.aarch64.dmg` and `rdd.windows.x86_64.exe`. The second commit makes the updater pick the Windows installer by architecture.

Upload the next release's files under the names the build gives them, because `linux-release.yaml` now fetches `rancher-desktop-<version>.linux.x86_64.zip` by that name.

Merge before #662, which rebases onto these names and drops its own naming commit.


jandubois:
Old AI review: https://jandubois.github.io/rancher-desktop-2/20260920-210308-pr-766.html

I don't plan on fixing I1; the script is specific to RD1 and doesn't work currently.

S1 to S3 have been addressed, but I don't think another `/deep-review` is required.

.github/workflows/package.yaml
```diff
@@ -77,45 +75,43 @@ jobs:
...
     - name: Upload mac zip
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-      if: matrix.platform == 'mac'
+      if: matrix.platform == 'darwin'
       with:
-        name: Rancher Desktop-mac.${{ matrix.arch }}.zip
-        path: dist/Rancher Desktop*.zip
+        name: rancher-desktop.darwin.${{ matrix.arch }}.zip
+        path: dist/rancher-desktop-*.darwin.*.zip
         if-no-files-found: error
```
mook-as: We should merge the various zip versions together, now that they have a consistent name:
```yaml
with:
  name: rancher-desktop.${{ matrix.platform }}.${{ matrix.arch }}.zip
  path: dist/rancher-desktop-*.${{ matrix.platform }}.*.zip
```
The `dmg` and `msi` ones still need `matrix.platform` checks, of course.

docs/development/release-checklist.md
```diff
@@ -4,11 +4,10 @@
 - [ ] Tag release branch. Wait for the CI to build artifacts.
 - [ ] Sign windows installer.
 
-### Sign mac installer (As there's an issue with the zip produced by the build script, we need to manually build and zip, rename the file to replace space with dot etc )
+### Sign mac installer (As there's an issue with the zip produced by the build script, we need to manually build and zip)
```
mook-as: Hmm, I no longer recall what the issue is. In fact, I don't think we build manually anymore either.

docs/development/release-checklist.md
```diff
@@ -18,8 +17,10 @@
 - [ ] Update metrics, roadmap on Confluence page
 
 ### Release
+- [ ] Upload the release artifacts to the GitHub draft Release page under the names the build gave them.
+  - From `yarn sign`'s `dist/`: the macOS dmg, zip, and `rdd` for both architectures, the Windows msi and `rdd.exe`, and the `.sha512sum` of each. The Package run's copies under the same names are unsigned.
```
mook-as: This may be better as per-platform items as checkboxes, so they can be checked as each platform is handled.

docs/development/release-checklist.md
mook-as: I wonder if this should be done as a GitHub issue template instead, so we just file an issue per release using that template…

docs/development/signing.md
```diff
@@ -19,7 +19,14 @@ In general, the process involves:
     yarn sign path/to/archive.zip
     ```
 
-4. Look in `dist/` for the signed files (`Rancher.Desktop.Setup.msi`, etc.).
+   To sign an archive built for the other architecture, export `GOARCH`
+   (`amd64` or `arm64`) first. `yarn sign` accepts only an archive built for
+   `GOARCH`, or for the architecture of Node.js itself when `GOARCH` is unset.
```
mook-as: I guess that's a funny way to say _for the host architecture_…

docs/development/signing.md
```diff
@@ -19,7 +19,14 @@ In general, the process involves:
...
+
+4. Look in `dist/` for the signed files, such as
+   `rancher-desktop-<version>.windows.x86_64.msi`.  They already have their
```
mook-as: Oh, right, we don't produce a windows zip file with signed executables, do we? I would if that would be useful? (Though that may lead to half-signed things if people mix-and-match…)

pkg/rancher-desktop/main/update/__tests__/LonghornProvider.spec.ts
```diff
@@ -452,10 +458,10 @@ describe('LonghornProvider.checkForUpdates', () => {
         name:         'Rancher Desktop 9.9.9',
         body:         'Simulated release.',
         published_at: '2038-01-01T00:00:00Z',
-        assets:       [
-          { name: assetName, browser_download_url: `${ serverURL }/msi`, size: 1 },
-          { name: `${ assetName }.sha512sum`, browser_download_url: `${ serverURL }/sha512sum`, size: 1 },
-        ],
+        assets:       assetNames.flatMap(name => [
+          { name, browser_download_url: `${ serverURL }/${ name }`, size: 1 },
+          { name: `${ name }.sha512sum`, browser_download_url: `${ serverURL }/${ name }.sha512sum`, size: 1 },
```
mook-as: For the URL, it would be good to make sure it can be arbitrary (not end with the correct name); that is, keep it ending with `/msi` and `/sha512sum`, or even call it `/installer` and `/checksum`.

pkg/rancher-desktop/utils/releaseArtifacts.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Names of release artifacts: `rancher-desktop-<version>.<platform>.<arch>.<ext>`
+ * for the application, `rdd.<version>.<platform>.<arch>[.exe]` for the daemon.
+ * The build writes these names and the updater looks them up.
+ */
+
+export type ArtifactPlatform = 'darwin' | 'linux' | 'windows';
+export type ArtifactArch = 'aarch64' | 'x86_64';
+
+/** electron-builder's `${ext}` macro, for an `artifactName` shared by several targets. */
+export const EXT_MACRO = `\${ext}`;
```
mook-as:
```suggestion
export const EXT_MACRO = '${ext}';
```
Seems silly to need escaping here?

pkg/rancher-desktop/utils/releaseArtifacts.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export function artifactPlatform(platform: string): ArtifactPlatform {
```
mook-as:
```suggestion
export function artifactPlatform(platform: NodeJS.Platform): ArtifactPlatform {
```
That's not as limited as we need, but it's closer? Alternatively, we can do a literal `"darwin" | "linux" | "win32"`.

pkg/rancher-desktop/utils/releaseArtifacts.ts
```diff
@@ -0,0 +1,51 @@
...
+    return 'windows';
+  }
+  throw new Error(`No release artifact name for platform ${ platform }`);
+}
+
+/**
+ * No artifact name may contain `arm64`, because electron-updater's MacUpdater
+ * treats any update file whose URL contains it as an arm64 build.
+ */
+export function artifactArch(arch: string): ArtifactArch {
```
mook-as: Similarly, `arch` can be `NodeJS.Architecture` or similar.

scripts/lib/build-utils.ts
```diff
@@ -451,20 +458,31 @@ export default {
...
     return process.arch === 'arm64' ? 'arm64' : 'x64';
   },
 
+  /**
+   * Throw unless a packaged application was built for `arch`, so that signing
+   * never names its files after another architecture.
+   * @param artifactName The `artifactName` in the application's electron-builder.yml.
+   */
+  checkArchiveArch(artifactName: string | null | undefined, version: string, platform: string): void {
+    const goarch = _.findKey(archByGoarch, arch => artifactName === appArtifactName(version, platform, arch, EXT_MACRO)) as keyof typeof archByGoarch | undefined;
```
mook-as:
```suggestion
    const goArch = _.findKey(archByGoarch, arch => artifactName === appArtifactName(version, platform, arch, EXT_MACRO)) as keyof typeof archByGoarch | undefined;
```

mook-as (changes requested):
Some trivial stuff, some stuff for follow-ups.

docs/development/release-checklist.md
jandubois: Outside the scope of the PR; let's discuss separately.

docs/development/signing.md
jandubois: We don't because the updater also uses the MSI. Could the updater do the same for DMG files on macOS? In that case I would drop the ZIP files too, although maybe that is controversial, idk.

pkg/rancher-desktop/utils/releaseArtifacts.ts
jandubois: ESLint's `no-template-curly-in-string` rejects `${ext}` in a plain string.

docs/development/signing.md
jandubois: On macOS ARM with Rosetta the architecture of Node might not match the host architecture.

docs/development/signing.md
mook-as: I think even if we could, that's enough work that it wouldn't be in scope in this PR.

mook-as (approved):

PR #779, opened 2026-09-17
Overlay nerdctl, buildkit and mkcert into the guest distro

jandubois:
Stacked on #447, which adds the `distro-overlay` tool this uses; merge that first. It also needs rancher-sandbox/rancher-desktop-opensuse#72 released, because that is the image that drops nerdctl, buildkit and mkcert.

`make build-rdd` now overlays nerdctl, buildctl, buildkitd and mkcert into the embedded distro. mkcert joins `rdd/dependencies.yaml` as a new GitHub dependency, and `rdd/overlay/manifest.yaml` says where each one goes in the guest. Each asset in `rdd/dependencies.yaml` also names the file the build stages it under, so `download-guest-deps` keeps no list of its own.

`XZ_OPTIONS` defaults to `-1` and the packaging workflow sets `-9`. On an 8-core M-series Mac the arm64 image takes 12 s at `-1` for 284 MB, against 116 s at `-9` for 217 MB, the size of the released v0.2.8 artifact.

Verified against a local build of rancher-sandbox/rancher-desktop-opensuse#72 (arm64): the raw image passes `e2fsck -fn`, every overlaid file is one extent at 0755 root:root, and all four binaries come back byte-identical by sha256. The WSL tarball extracts with the same modes and bytes. It also builds green against the pinned v0.2.8, which still has the old binaries, where the overlay grows them in place.


rdd/Makefile
```diff
@@ -96,14 +121,49 @@ $(GUESTAGENT_GZ): $(GUESTAGENT_BINARY)
...
+$(DISTRO_IMAGE): $(PRISTINE_DISTRO) $(OVERLAY_SOURCES) $(OVERLAY_TOOL_SOURCES)
+	@command -v xz >/dev/null || { \
+		echo 'xz is missing; install it to build the guest distro image.' >&2; \
+		exit 1; }
+	xz --decompress --stdout $(PRISTINE_DISTRO) > $(WORKING_DISTRO)
+	WSLENV=${WSLENV}:GOOS:GOARCH GOOS=$(HOST_OS) GOARCH=$(HOST_ARCH) \
+	go$(EXE) run ./cmd/distro-overlay --manifest overlay/manifest.yaml --source overlay \
+		--mtime $(OVERLAY_MTIME) --in-place $(WORKING_DISTRO)
+	xz $(XZ_OPTIONS) --stdout $(WORKING_DISTRO) > $@
+	rm -f $(WORKING_DISTRO)
```
Nino-K: If `distro-overlay` or `xz` error out, this line is never reached, and it would leave a large file (> GB) behind, no?

rdd/Makefile
```diff
@@ -96,14 +121,49 @@ $(GUESTAGENT_GZ): $(GUESTAGENT_BINARY)
...
+	ls -lh $@
+
+# A failed xz leaves a truncated image that make would take for a good one.
+.DELETE_ON_ERROR:
```
Nino-K: So this now applies to every target in the Makefile, not just the image.

rdd/pkg/guestdeps/manifest.go
```diff
@@ -123,6 +126,11 @@ func (a Asset) validate() error {
 	if !checksumPattern.MatchString(a.Checksum) {
 		return fmt.Errorf("asset %s has checksum %q, want sha256:<64 lowercase hex digits>", a, a.Checksum)
 	}
+	// Stage joins the filename onto the destination directory, so a name with a
+	// separator in it would write outside the build tree.
+	if !isPathElement(a.Filename) {
```
Nino-K, line 167: It looks like this checks each filename's but nothing stops two dependencies from picking the same one.

rdd/Makefile
jandubois: Fixed

rdd/Makefile
jandubois: Yes, it does. Moved it to the top of the file.

rdd/pkg/guestdeps/manifest.go
jandubois: Fixed (I think)

Nino-K (approved):
LGTM

PR #780, opened 2026-09-17
CI: give the Go cache a prefix fallback and write it only on main

jandubois:
Replaces `actions/setup-go`'s built-in cache with a `setup-rdd-go` composite action. It keys on `rdd/go.sum` with a `go-<os>-<arch>-<goversion>-` prefix fallback, and only a push to `main` writes it.

Measured on a fork. Times are the whole BATS step, so they include test time, not just the build:

| | ubuntu bats-docker | ubuntu bats-others | macos bats-docker | windows bats-others |
|---|---|---|---|---|
| cold | 343s | 601s | 903s | 1296s |
| warm, exact key | 121s | 367s | 363s | 682s |

A `go.sum` change invalidating 18 of 3005 packages missed its primary key, restored the previous entry through the prefix, and recovered 85-97% of that saving: 128s, 393s and 444s on the first three columns.

Only a small bump has been measured this way. Across the 51 `go.sum` changes that reached main in the last 92 days, the median invalidates 4.7% of the package graph and nine invalidate more than half, where recovery will be lower.

The prefix fallback and the main-only write follow [Dan Peterson's "Better GitHub Actions caching for Go"](https://danp.net/posts/github-actions-go-cache/). Upstream recommended the same shape in actions/setup-go#404.


.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
+# The prefix fallback and the main-only write follow Dan Peterson, "Better
+# GitHub Actions caching for Go": https://danp.net/posts/github-actions-go-cache/
```
mook-as: I don't think this comment is necessary in the file (you already have it in the commit message, where it's more relevant).

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+name: Setup Go with a cache
+description: >-
+  Installs the Go version rdd/go.mod names and restores rdd's module and build
+  caches. When rdd/go.sum has changed the restore falls back to the newest cache
+  for the same runner and Go version, so a dependency bump recompiles only the
+  packages it affects.
```
mook-as: Ugh, this has the weird grammar that AI likes again. Try something like

> Installs the go toolchain as specified in rdd/go.mod, and restores caches for it.  On rdd/go.sum mismatch, restore based on runner and go version.

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+inputs:
+  save:
+    description: >-
+      Whether this job may write the cache. Turn it off for a job that does not
+      build rdd, because the first job to finish claims the key and every later
+      job restores what it stored.
+    default: "true"
```
mook-as: Should the default be false instead?  We probably only want it on a specific job, to ensure we're saving the cache consistently.

I'd suggest the `Lint RDD` job; that should build RDD + golangci-lint (which is sort of big by itself), though I guess `go test` might download more things too…

.github/actions/setup-go-cached/action.yaml
mook-as, line 1: Since this is using RDD for configuration, maybe call it `setup-rdd-go`?

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+  using: composite
+  steps:
+  - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+    with:
+      cache: false
+      go-version-file: rdd/go.mod
+
+  # GOCACHE is under %LocalAppData% on Windows and ~/.cache elsewhere, so ask
+  # the toolchain for both paths. The key includes the architecture because
+  # macos-latest and macos-15-intel both report runner.os == 'macOS'.
```
mook-as:
```suggestion
  # GOCACHE / GOMODCACHE varies per platform, so ask the toolchain for both
  # paths.
```
The architecture comment is unnecessary, because it should be obvious that the mod cache (which contains compiled things) is different per-platform; this will also be the case once we support linux/arm64 (and possibly one day, windows/arm64).

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+  - id: go-env
+    shell: bash
+    run: |
+      set -o errexit -o pipefail
```
mook-as: Isn't this set by default?

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+      {
+        echo "build-path=$(go env GOCACHE)"
+        echo "module-path=$(go env GOMODCACHE)"
+        echo "prefix=go-${{ runner.os }}-${{ runner.arch }}-$(go env GOVERSION)-"
+        echo "save=${{ inputs.save == 'true' && github.event_name == 'push' && github.ref == 'refs/heads/main' }}"
```
mook-as: The workflow expansion is better done as an environment variable; doing it that way makes it easier to see what it evaluated to, ant it also allows line wrapping (via the use of `>-`).

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+      } >>"$GITHUB_OUTPUT"
```
mook-as:
```suggestion
      } | tee -a "$GITHUB_OUTPUT"
```
Maybe do this so we can see the output too? (`--append` isn't documented to be supported on macOS/BSD `tee`)

.github/actions/setup-go-cached/action.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  # A cache written on a pull request is visible to that pull request alone and
+  # dies with it, so only main writes and every other ref reads what main left.
```
mook-as: I think a better justification is:

> We only save on pushes to main to avoid churn from pull requests and dependabot pushes; as the caches are large, saving too often can push out caches from other platforms / architectures.

It's not just a pull request thing: pushes to non-main branches count too.

mook-as (changes requested):
Please review your AI's English harder, thanks!

.github/actions/setup-go-cached/action.yaml
jandubois: Yes, only saving on the `lint` job is the right thing. Need to use the same tags and CGO settings as the regular build though, but that is not a bad idea anyways.

jandubois: Also need to use same tags and options on `make test`.

And for macOS Intel we need to save the cache on bats because we run the linter on M1.

mook-as (approved):
I have no comments.

PR #772, opened 2026-09-15
RDD: engine: Fix containerd namespace mirroring

mook-as:
This matches the new API introduced in #651 where we hash invalid names.

PR #786, opened 2026-09-17
RDD: engine: containerd namespace: fix failure to get labels

mook-as:
Per AI review for #772: if we failed to get a label, we cannot just use `nil` in the apply, as that will remove the labels because it's owned by the same field owner.  Fix this by using a separate field owner.


jandubois (approved):
Thanks, LGTM

AI review claims you should have included a test: https://jandubois.github.io/rancher-desktop-2/20260917-155123-pr-786.html if you want to add one

jandubois (approved):
Thanks, LGTM

AI review claims you should have included a test: https://jandubois.github.io/rancher-desktop-2/20260917-155123-pr-786.html if you want to add one

mook-as:
I1: Doesn't matter; we don't expect the labels to be useful soon.
S1: That's mostly about existing data, which doesn't matter: this is alpha, and it's all mirrored from containerd anyway.
S2: Meh, that was just to trap AI anyway.
S3: It's only used (and intentionally only used) in one place.
S4: The logger is only used in one place; it's just an inlined variable.

PR #787, opened 2026-09-17
RDD: namespace mirror names: address review comments

mook-as:
Fixed the name encoding scheme to ensure we hash input names that look like they could be already hashed, to avoid collisions.  Also update the docs, and one usage (volumes) where a very similar algorithm is used.

Part of #772 feedback (S1).


jandubois:
AI review: https://jandubois.github.io/rancher-desktop-2/20260918-113246-pr-787.html

mook-as:
I1: irrelevant; we are still in alpha, and do not support objects from the previous commit, much less the previous release.  We may want to consider dealing with it once we have versioned APIs or something.

Most others are fixed.

As usual with AI reviews, it pulled in more things -_-

jandubois (approved):
Updated AI review: https://jandubois.github.io/rancher-desktop-2/20260922-142801-pr-787.html

I think getting the encoding to work for both engines is getting too complicated in this PR, and we should move the `containerd` work to a followup (of this followup). I can take that on if you want.

AI suggest to still do the following though:

- **I2, the separator test that can't fail.** The fix changes two fixture strings. Without it, the test gives false coverage, and the PR deleted the only test that actually guarded the separator. This is the one I'd insist on.
   
- **I3, the `MirrorName` doc comment.** The fix replaces nine lines of comment with six. It's the exported contract, so it's better fixed now, but deferring it wouldn't break anything.
   
- **I1, the Image doc bullet.** If containerd moves to a follow-up, a two-word qualifier ("Under moby,") keeps the doc from stating the wrong rule for containerd in the meantime. That's optional.

mook-as:
I'll do a follow-up, including the various AI suggestions; this is mostly to make rebasing easier.

PR #791, opened 2026-09-18
RDA: Port rdd-client generation script to bash

mook-as:
This ports the script we use to generate rdd-client from TypeScript to bash, so that we can run it easier; in particular, this also makes it part of `make verify-generate`.

Note that `make generate` does _not_ do this, because it relies on running `rdd` to introspect out the OpenAPI specs.  `make generate` may be needed to make `rdd` buildable again.

This currently doesn't work on Windows (where it's expected to use msys), because of an unrelated issue with docker mounts; once I track that down I'll file an issue.


jandubois (changes requested):
AI review: https://jandubois.github.io/rancher-desktop-2/20260918-104403-pr-791.html

Only I1 (breaks on macOS) is important; the rest look cosmetic.

jandubois (approved):
Thanks, LGTM

PR #794, opened 2026-09-19
Raise the BATS target timeout to 75 minutes

jandubois:
Raises `BATS_TIMEOUT` from 60 to 75 minutes.

A stopgap: bats-app on macos-15-intel has breached the 60-minute cap on 9 of 52 runs since #763 merged. Splitting the suite and cutting its VM restarts is the real fix.


mook-as (approved):
Yeah, we should split that up soon: we still block on this to iterate / merge PRs.

PR #795, opened 2026-09-19
EXPERIMENT: decode the distro without punching holes

jandubois:
**Not for merge.** A CI measurement, stacked on #794.

Decodes the embedded distro without punching holes, so `BATS (macos-15-intel, bats-app)` runs against a fully allocated image. The gap between this run's step duration and #794's says whether the holes account for the time #763 added beyond the in-process decode.


jandubois:
Experiment showed that not punching holes is even slower (but really within noise margins).

PR #796, opened 2026-09-19
RDD: Decode the embedded distro's xz blocks in parallel

jandubois:
Decodes the blocks of an `xz --threads` stream concurrently, each into its own region of the output file.

Measured on `macos-15-intel` bats-app, four runs per branch normalized by a per-run speed probe: this branch is indistinguishable from pre-#763 (p=0.714), while main differs from both (p=0.0286, the floor for four against four). The per-run decode total goes from 494s on main to 265s here.

Verified on a live VM, where the decode of the embedded image logs 5.76s during a create. `tests/32-app-controllers/engine-docker.bats` passes 43/43.


rdd/pkg/xz/parallel.go
```diff
@@ -0,0 +1,229 @@
...
+	for off := int64(streamHeaderSize); ; {
+		n, err := r.ReadAt(buf, off)
+		if n == 0 && err != nil {
+			return nil, errNotSplittable
+		}
+		b := buf[:n]
+		if b[0] == indexIndicator {
+			// The index ends the stream. Anything after the stream footer is
+			// a second stream or stream padding, which this decoder skips.
+			break
```
Nino-K, line 138: Doesn't skipping a second stream here means silently dropping its data? `parseBlocks`  returns just the first stream's blocks with no error, so the decode comes out truncated.

jandubois: Good catch! Fixed now. Also needed a bunch of other error checks that the single-threaded version does.

jandubois:
AI review for reference: https://jandubois.github.io/rancher-desktop-2/20260921-231632-pr-796.html

Nino-K (approved):

PR #797, opened 2026-09-19
RDD: Record where the bats suites spend their time

jandubois:
Records each suite's elapsed time, cap and outcome from the wrapper, times every file's `local_setup_file` and `teardown_file`, puts per-test durations in the TAP stream with `bats --timing`, and samples the runner's disk, memory and load counters. The daemon times Lima's `Prepare` and records the size and allocated bytes of both files a create writes. `build-rdd` moves into a step of its own so the timed step runs bats alone.

`bats-others` runs six suites inside one step, so five of them have never had a duration at all.

It does not touch `BATS_TIMEOUT`, which #794 already raised to 4500.

Verified on a live VM: `tests/32-app-controllers/engine-docker.bats` 43/43, with `local_setup_file` recorded at 56.2s.

Overlaps #442, which adds a macOS-only host sampler and an in-guest one behind `RDD_VM_TELEMETRY`. Both hook `rdd-bats.yaml`, so the two samplers want reconciling.


.github/actions/unix-machine-info/action.yaml
mook-as: Not this PR, but perhaps we should merge the Windows one too, and just have a generic action that checks `runner.os` and does the right thing.

.github/workflows/rdd-bats.yaml
```diff
@@ -257,6 +264,22 @@ jobs:
             ;;
         esac
 
+    # Build ahead of the test step so the timed step runs bats alone. A build
+    # inside it varies with the Go cache and distorts the duration.
+    - name: Build rdd
+      if: needs.check-paths.outputs.should-run
+      run: make -C bats build-rdd
+      working-directory: rdd
```
mook-as: Kinda weird to make `build-rdd` in the `bats` directory, which immediately runs `make` in its parent…

rdd/scripts/bats-with-timeout.sh
```diff
@@ -30,9 +30,21 @@ shift
 
 instance="${RDD_INSTANCE:-2}"
 
+# The last argument is the directory bats runs, which names the suite. The
+# instance cannot: bats-helpers sets no RDD_INSTANCE and falls back above, and
+# bats-docker and bats-nerdctl both run tests/40-docker.
```
mook-as: I fail to parse the second sentence: _The instance cannot: …_ — I think a verb may be missing; I don't understand what _The instance cannot name the suite…_ might be, if that was implied by the previous sentence.

As before, please rephrase anything with _[verb] no [noun]_; here, it would probably be _bats-helper does not set RDD_INSTANCE and…_

The _very_ specific mention of `tests/40-docker` is probably misplaced?

rdd/scripts/bats-with-timeout.sh
```diff
@@ -30,9 +30,21 @@ shift
...
+suite=${*: -1}
+suite=${suite%/}
+suite=${suite##*/}
```
mook-as: This triple-munging probably needs a comment, showing sample text before and after.

rdd/scripts/bats-with-timeout.sh
```diff
@@ -491,6 +507,21 @@ done
 exit_code=0
 wait "${cmd_pid}" || exit_code=$?
 
+# Record the suite before the post-run bundle, whose capture takes seconds.
+# A killed run's step duration is the cap, not the suite's length, so the
+# outcome column marks which rows are comparable.
```
mook-as: The comment could probably be reworded.

> Record the suite before post-run executes, as that can take multiple seconds.

I don't understand the second sentence enough to rephrase it.

rdd/scripts/metrics.bash
```diff
@@ -0,0 +1,71 @@
...
+
+# metrics_init <repo-root>
+metrics_init() {
+    : "${RDD_METRICS_DIR:=${1}/rdd-logs/_diag/metrics}"
+    export RDD_METRICS_DIR
+    mkdir -p "${RDD_METRICS_DIR}"
+}
+
+# Milliseconds since the epoch. EPOCHREALTIME (bash 5) has microsecond
+# resolution; BSD date has no %N, so older bash falls back to whole seconds.
```
mook-as:
```suggestion
# resolution; BSD date does not have %N, so older bash falls back to whole seconds.
```

The implementation detail should go _inside_ the function, not the doc comment.

rdd/scripts/metrics.bash
```diff
@@ -0,0 +1,71 @@
...
+metrics_phase() {
+    local phase=$1 start_ms=$2 end_ms
+    end_ms=$(metrics_now_ms)
+    metrics_header phase file phase start_ms end_ms elapsed_ms
+    metrics_record phase \
+        "$(basename "${BATS_TEST_FILENAME:-unknown}")" "${phase}" \
+        "${start_ms}" "${end_ms}" "$((end_ms - start_ms))"
+}
+
+# metrics_timeout <seconds> <command>... GNU timeout is absent from a stock
```
mook-as: Again, implementation detail (whether `timeout` works) goes inside the function.

mook-as (changes requested):

.github/workflows/rdd-bats.yaml
jandubois: Fixed; the step now runs `make build-rdd` from `rdd/`.

rdd/scripts/bats-with-timeout.sh
jandubois: The Makefile now passes the make target as the suite name, so the wrapper no longer derives it

mook-as (approved):

PR #800, opened 2026-09-20
Stage a locally built distro instead of downloading it

jandubois:
Adds `--stage NAME=PATH` to `download-guest-deps`, which copies a local file into the staging directory in place of that dependency's download, and `STAGE_DISTRO` to `rdd/Makefile`, which forwards a path to it as `--stage distro=PATH`.

The staged bytes go unchecked by design. The log line for the copy ends in `unverified`, and both the README and the `STAGE_DISTRO` comment say only a test build should set it.


rdd/Makefile
```diff
@@ -138,7 +143,7 @@ $(GUESTAGENT_GZ): $(GUESTAGENT_BINARY)
 $(PRISTINE_DISTRO): FORCE
 	WSLENV=${WSLENV}:GOOS:GOARCH GOOS=$(HOST_OS) GOARCH=$(HOST_ARCH) \
 	go$(EXE) run ./cmd/download-guest-deps --os $(TARGET_OS) --arch $(TARGET_ARCH) \
-		--dest $(OVERLAY_BUILD)
+		--dest $(OVERLAY_BUILD) $(if $(STAGE_DISTRO),--stage distro=$(STAGE_DISTRO))
```
Nino-K: Are we ever going to come across a path with spaces in it? Probably not.

Nino-K (approved):

rdd/Makefile
jandubois: Rare, but a Windows path also lost its backslashes without quotes (`C:\Users\x` became `C:Usersx`), so I quoted it in 4aead395d.

Nino-K (approved):
LGTM

PR #811, opened 2026-09-20
Give the macOS BATS guest the host's TSC frequency

jandubois:
The guest on the macOS BATS legs runs on the HPET, because the runner is itself a VM and the guest cannot calibrate its TSC, so every clock read traps out to the hypervisor. The macOS legs now read the host's frequency from `sysctl` and append `tsc_early_khz=` to the guest image's boot entries, through distro-overlay's new `--kernel-params` flag.

Nothing ships with the parameter, and no other platform is touched. RD1 established that bare-metal Intel Macs calibrate normally, so users are unaffected.

## Measured effect

Three interleaved control/treatment pairs on `macos-15-intel`, dispatched from my fork, all six green. "Suite" is the TAP plan line to the last `ok`, so it excludes the build.

| bats-app suite | control | treatment | ratio |
|---|---|---|---|
| raw | 59.2 min | 40.0 min | 1.48 |
| minus the xz decode | 46.3 min | 28.2 min | 1.64 |

Normalizing each run by a probe of the tests that never create an instance puts those at 1.40x and 1.55x. All four separate perfectly, so p is 0.05 one-sided, the floor at three against three.

The slowest control run spent 4034s of the 4500s per-target cap that `rdd/bats/Makefile:13` calls a stopgap; the treatment runs used 2340s to 2554s.


rdd/pkg/overlay/boot.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// appendToLinuxLines appends params to every `linux` line and reports how many
+// it changed. grub2-mkconfig writes one per menu entry, holding the kernel path
+// and the whole command line.
+func appendToLinuxLines(data []byte, params string) (patched []byte, count int) {
+	lines := strings.Split(string(data), "\n")
+	for n, line := range lines {
+		fields := strings.Fields(line)
+		if len(fields) == 0 || fields[0] != "linux" {
```
Nino-K: I know our guest is `openSUSE` but we might want to consider that this is the exact match that would skip  `linuxefi / linux16` line and return `count == 0`, just thinking in terms of a little more flexibility.

.github/workflows/rdd-bats.yaml
```diff
@@ -239,6 +239,18 @@ jobs:
...
+    - name: "macOS: Give the guest the host's TSC frequency"
+      if: needs.check-paths.outputs.should-run && runner.os == 'macOS'
+      shell: bash
+      # The runner is itself a VM, so the guest cannot calibrate its TSC and
+      # falls back to the HPET, where every clock read traps out to the
+      # hypervisor. tsc_early_khz= hands the kernel the frequency instead.
+      # Only this runner knows its own, so no image can carry the value.
+      run: |
+        set -o errexit -o nounset -o pipefail
+        echo "KERNEL_PARAMS=tsc_early_khz=$(( $(sysctl -n machdep.tsc.frequency) / 1000 ))" \
```
Nino-K: I guess we have to update this later if we ever decide to run on Apple Silicon since `machdep.tsc.frequency` is Intel-only and Apple Silicon runner would fail the sysctl.

Nino-K (approved):

